PR #5, opened 2024-10-22
Propertytree redux

engbergandreas:
This pr adds the property tree middleware and reducer, it also implements the `events` middleware but only handles the cases for the property tree in `handleData`. I did not want to get ahead of myself before we start implementing other parts that rely on the event topic.


src/redux/propertytree/propertyTreeSlice.ts
```diff
@@ -0,0 +1,194 @@
...
+export const propertyOwners = createSlice({
+  name: 'propertyOwners',
+  initialState: initialStatePropertyOwners,
+  reducers: {
+    addPropertyOwners: (
+      state,
+      action: PayloadAction<{ propertyOwners: PropertyOwner[] }>
+    ) => {
+      const inputOwners = action.payload.propertyOwners;
+      // const newState = { ...state };
```
WeirdRubberDuck, line 130: :)

src/redux/propertytree/propertyTreeSlice.ts
```diff
@@ -0,0 +1,194 @@
...
+      return state;
+    },
+    clearPropertyTree: () => {
+      return {
+        isInitialized: false,
+        propertyOwners: {}
+      };
+    },
+    removePropertyOwners: (state, action: PayloadAction<{ uris: string[] }>) => {
+      // const newState = { ...state };
```
WeirdRubberDuck, line 165: :)

src/redux/propertytree/propertyTreeSlice.ts
```diff
@@ -0,0 +1,194 @@
...
+      });
+      return state;
+    }
+  }
+});
+
+export const { addPropertyOwners, removePropertyOwners } = propertyOwners.actions;
+export const propertyTreeReducer = combineSlices({
+  props: properties.reducer,
+  owners: propertyOwners.reducer
```
WeirdRubberDuck, line 193: In particular, I do not think we should use the `props` shortening here (since this can be misinterpreted with the redux props). 

But I also think we should avoid renaming these states, in general. It's better to be too specific rather than vague in the naming, and not renaming them will simplify for us when implementing the existing components



```suggestion
  properties: properties.reducer,
  propertyOwners: propertyOwners.reducer
```

engbergandreas: Yeah I agree, what I did not like about having their original names was that when we access the state `properties` or `propertyOwners` it will be written twice for example here: `(state) => state.propertyTree.props.properties[NavigationAimKey]` vs ` (state) => state.propertyTree.properties.properties[NavigationAimKey]`

The combineSlices or combineReducers does not seem to work as they did before where it was all smudged into one reducer. So one alternative could be that we write them as completely separate reducers, then you would access them as `state.properties[key]` and `state.propertyOwner[key]`, though it is nice that they are in the `propertyTree` subobject so I'm not too sure about this either..

src/redux/propertytree/propertyTreeSlice.ts
WeirdRubberDuck: Here's a comment on this behavior in a video, for future self
https://youtu.be/5yEG6GhoJBs?t=980

More notes to self: 
In summary, `createSlice` allows us to write "mutating code", but there is a bunch of magic happening behind the scene to create copy of the state and really only edit the copy. 

My first intuition is that we want to write the code more explicitly, by creating a copy like Ylva had started doing, to avoid confusion and reduce the need to know about the "magic"... But yeah let's discuss

PR #6, opened 2024-10-28
Add scrollbars to windows if the content takes up more than 100% height

WeirdRubberDuck:
Not sure if this is the way to go, since I'm not that familiar with the window code setup.  Hence the PR :)

This works for the Scene and Session Recording panels, but would be good to have someone else try it in more contexts


engbergandreas:
Unfortunately, this breaks deeper `ScrollArea` like `FilterList`. I'll continue investigating other solutions

WeirdRubberDuck:
Should this PR be closed?

PR #19, opened 2024-11-27
add screenspace renderable panel

engbergandreas:
this adds the screenspace panel, I thought it might be good to have the `stringToIdentifier` function so that the user does not have to set the slide name according to all our rules for an identifier. I don't know if we have more rules that we need to include in the regex. Also, this function is an exact copy of the geolocation panel to deal with addresses so we should probably use the same for that panel if this one is ok


src/panels/ScreenSpaceRenderablePanel/ScreenSpaceRenderablePanel.tsx
```diff
@@ -0,0 +1,109 @@
...
+}
+
+export function ScreenSpaceRenderablePanel() {
+  const [slideName, setSlideName] = useState('');
+  const [slideURL, setSlideURL] = useState('');
+  const luaApi = useOpenSpaceApi();
+  const screenSpacePropertyOwner = useGetPropertyOwner(ScreenSpaceKey);
+
+  const renderables = screenSpacePropertyOwner?.subowners ?? [];
+  const isButtonDisable = !slideName || !slideURL;
```
ylvaselling, line 34:
```suggestion
  const isButtonDisabled = !slideName || !slideURL;
```

src/panels/ScreenSpaceRenderablePanel/ScreenSpaceRenderablePanel.tsx
```diff
@@ -0,0 +1,109 @@
...
+      Type: 'ScreenSpaceImageLocal'
+    };
+
+    if (slideURL.indexOf('http') !== 0) {
+      renderable.Type = 'ScreenSpaceImageLocal';
+      renderable.TexturePath = slideURL;
+    } else {
+      renderable.Type = 'ScreenSpaceImageOnline';
+      renderable.URL = slideURL;
+    }
```
ylvaselling, line 49: Switch these two to make it easier to read?

ylvaselling (approved):

src/panels/ScreenSpaceRenderablePanel/ScreenSpaceRenderablePanel.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    luaApi?.addScreenSpaceRenderable(renderable);
+  }
+
+  function removeSlide(uri: string) {
+    // We need to remove the 'ScreenSpaceKey.' part from the URI which has the format:
+    // `ScreenSpaceKey.{identifier}`
```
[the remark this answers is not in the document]
WeirdRubberDuck: Maybe we should even make a helper function for it. I can add one to the Scene Tree branch

PR #20, opened 2024-12-02
Feature/async thunks

ylvaselling:
Changed all async logic in redux to thunks
Moved lua api to provider

Testing would be helpful 🙏


src/redux/connection/connectionMiddleware.ts
```diff
@@ -1,40 +1,48 @@
...
+      thunkAPI.dispatch(onOpenConnection());
+    }
+
+    function onDisconnect() {
+      thunkAPI.dispatch(onCloseConnection());
+
+      let reconnectionInterval = 1000;
+      setTimeout(() => {
+        api.connect();
+        reconnectionInterval += 1000;
```
WeirdRubberDuck, line 23: Interesting, didn't know that we increased the interval with a second at every attempt. Wonder if this is still meaningful?

WeirdRubberDuck (commented):
Looks good to me. COmmented on some things I wondered about

Will test with the updated scene tree as well

src/redux/propertytree/propertyTreeMiddleware.ts
```diff
@@ -1,29 +1,65 @@
...
-export const addUriToPropertyTree = createAction<{ uri: string }>('addUriToPropertyTree');
+export const reloadPropertyTree = createAction<void>('propertyTree/reload');
 export const removeUriFromPropertyTree = createAction<{ uri: string }>(
-  'removeUriFromPropertyTree'
+  'propertyTree/removeUri'
+);
+
+export const addUriToPropertyTree = createAsyncThunk(
+  'propertyTree/addUri',
+  async (uri: string) => {
```
engbergandreas, line 30: Imo we should explicitly type the return type here

engbergandreas: Perhaps on all of these async thunk functions

src/api/LuaApiProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+      }
+    };
+    if (isConnected) {
+      fetchApi();
+    }
+  }, [isConnected]);
+
+  // The groups ordering is not part of any topic so we request it
+  // once we have the api
+  // TODO 2024-12-02 (ylvse): make this a part of a topic?
```
WeirdRubberDuck, line 41: Maybe it would be better to dispatch an action that says essentially "lua Api ready" and then listen to that in the groups? Or, it could be handled the same way as the exoplanets (the case is similar)

There is also an event that should trigger this action, but I think that still has to be implemented. 

Butyeah, let's discuss what to do with this, because I find it pretty weird to have this code in the provider for the API

src/main.tsx
```diff
@@ -2,13 +2,16 @@ import { StrictMode } from 'react';
...
 
 createRoot(document.getElementById('root')!).render(
-  <StrictMode>
-    <Provider store={store}>
-      <App />
-    </Provider>
-  </StrictMode>
+  <Provider store={store}>
+    <LuaApiProvider>
+      <StrictMode>
```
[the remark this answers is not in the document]
ylvaselling: Yes, the strict mode runs everything twice, so effects are run & finished twice, which happens way too quickly for a websocket to establish a connection. That has the effect that the websocket isn't properly opened before it should be closed, and then we cant close it properly. So it should not live inside of the strict mode. For the future it would be better to move all websockets into the RTK Query which will handle all that for us but for now we can have them outside of strict mode

src/redux/connection/connectionMiddleware.ts
ylvaselling: I think this is a way to basically not create an interval. It will just make retries with 1 second interval

src/redux/store.ts
```diff
@@ -16,20 +15,15 @@ export const store = configureStore({
...
   },
   middleware: (getDefaultMiddleware) =>
-    getDefaultMiddleware({
-      serializableCheck: {
-        ignoredActions: ['luaApi/initializeLuaApi'],
-        ignoredPaths: ['luaApi']
-      }
-    }).prepend([listenerMiddleware.middleware])
+    getDefaultMiddleware().concat([listenerMiddleware.middleware]),
+  devTools: true
```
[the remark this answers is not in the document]
ylvaselling: Ah no, good catch

src/redux/actions/actionsMiddleware.ts
```diff
@@ -1,46 +1,37 @@
...
-  const topic = api.startTopic('shortcuts', {
-    event: 'get_shortcut',
-    identifier: uri
-  });
-  const { value } = await topic.iterator().next();
-  topic.cancel();
-  return value.shortcuts;
-}
-const getAction = createAction<string>('getAction');
+// TODO 2024-11-27 (ylvse): This action should be added to the events handling
```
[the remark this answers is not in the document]
ylvaselling: Yeah, I agree that our todo comments are a bit all over the place. Making a ticket for it in our notion

src/redux/actions/actionsMiddleware.ts
```diff
@@ -1,46 +1,37 @@
...
-      listenerApi.dispatch(initializeShortcuts(result));
-    }
-  });
-
-  startListening({
-    actionCreator: getAction,
-    effect: async (action, listenerApi) => {
-      const result = await getShortcut(action.payload);
-      listenerApi.dispatch(addActions(result));
+      listenerApi.dispatch(getAllShortcuts());
```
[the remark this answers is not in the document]
ylvaselling: It should be used in the events middleware so leaving it for now. The thunk creates an action that you can dispatch

src/redux/propertytree/propertyTreeMiddleware.ts
ylvaselling: Tried this. It was not straightforward... putting it in notion as to be done later

src/api/LuaApiProvider.tsx
ylvaselling: We discussed - leaving for now

PR #21, opened 2024-12-04
Feature/user panels panel

ylvaselling:
Note that the scroll doesn't work properly for the iframes. I think we really need the height of the windows from the provider... Leaving it for now.

To test: local panels: add folders with an index.html file in the user/webpanels folder
url: add url (webpage). it has to be a webpage that allows being an iframe. example: ui.openspaceproject.com

Also added an optional title to the url pages


src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+      // eslint-disable-next-line no-template-curly-in-string
+      const root = await luaApi.absPath('${USER}/webpanels');
+      const folders = await luaApi.walkDirectoryFolders(root);
+
+      if (!folders) return;
+
+      const slash = navigator.platform.indexOf('Win') > -1 ? '\\' : '/';
+      const folderNames = Object.values(folders).map((panel) =>
+        // Get the folder name from the path
+        panel.substr(panel.lastIndexOf(slash) + 1)
```
WeirdRubberDuck, line 48: Will this really work when the slash is the `'\\'` version?  (thinking of the +1)

Maybe:
```suggestion
        panel.split(slash).pop()
```
?

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const luaApi = useOpenSpaceApi();
+  const isDataInitialized = useAppSelector((state) => state.userPages.isInitialized);
+
+  const localPanels = useAppSelector((state) => state.userPages.panels || []);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    // Collect all folder paths in the USER folder
+    const getUserPanels = async () => {
+      if (!luaApi) return;
```
WeirdRubberDuck, line 37: Lots of one line returns in this PR. Up until now we have avoided those and always used braces, i.e.
```suggestion
      if (!luaApi) {
        return;
      }
```
We should decide which version we prefer and stick to one. 

One thing to consider is that this would not be allowed according to our C++ coding guidelines, so if we want to be a similar to those as possible, we should used the braced version

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+      // eslint-disable-next-line no-template-curly-in-string
+      const root = await luaApi.absPath('${USER}/webpanels');
+      const folders = await luaApi.walkDirectoryFolders(root);
+
+      if (!folders) return;
```
WeirdRubberDuck, line 43: See previous comment

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+      dispatch(intializeUserPages({ panels: folderNames, isInitialized: true }));
+    };
+
+    if (luaApi && !isDataInitialized) {
+      getUserPanels();
+    }
+  }, [dispatch, isDataInitialized, luaApi]);
+
+  function addLocalPanel() {
+    if (!selectedPanel) return;
```
WeirdRubberDuck, line 59: See previous comment

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+    addWindow(<UserPage src={src} title={selectedPanel} />, {
+      title: selectedPanel,
+      position: 'right',
+      id: selectedPanel
+    });
+    setSelectedPanel(null);
+  }
+
+  function addWebPanel() {
+    if (!panelURL) return;
```
WeirdRubberDuck, line 74: See previous comment

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const localPanels = useAppSelector((state) => state.userPages.panels || []);
+  const dispatch = useAppDispatch();
+
+  useEffect(() => {
+    // Collect all folder paths in the USER folder
+    const getUserPanels = async () => {
+      if (!luaApi) return;
+
+      // eslint-disable-next-line no-template-curly-in-string
+      const root = await luaApi.absPath('${USER}/webpanels');
```
WeirdRubberDuck, line 40: How about moving the  `'${USER}/webpanels'` to keys.ts?

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+      id: selectedPanel
+    });
+    setSelectedPanel(null);
+  }
+
+  function addWebPanel() {
+    if (!panelURL) return;
+    const startsWithHttp = panelURL.indexOf('http') === 0;
+    const src = startsWithHttp ? panelURL : `http://${panelURL}`;
+    // As of now, we don't have titles for the URLs so we just use the URL
```
WeirdRubberDuck, line 77: We do now though? :) 
```suggestion
```

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+      id: title
+    });
+
+    setPanelURL('');
+    setUrlPanelTitle('');
+  }
+
+  return (
+    <Container>
+      <Title my={'xs'} order={4}>
```
WeirdRubberDuck, line 92:
```suggestion
      <Title my={'xs'} order={2}>
```

We should avoid using the order to code the title to a specific size, and use the "correct" order for the hierarchy. 

The other panels start with a `h2` title, so this should also be order 2

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+        User Panels
+      </Title>
+      <Title order={5}>Add Local Panel</Title>
```
WeirdRubberDuck, line 95: See previous comment
```suggestion
      <Title order={3}>Add Local Panel</Title>
```

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+          flex={3}
+          onKeyDown={(e) => e.key === 'Enter' && addLocalPanel()}
+        />
+        <Button onClick={addLocalPanel} disabled={!selectedPanel} p={'xs'}>
+          <UserPageIcon />
+          <Text mx={'xs'}>Add</Text>
+        </Button>
+      </Group>
+      <Divider my={'md'} />
+      <Title order={5}>Add from URL</Title>
```
WeirdRubberDuck, line 111:
```suggestion
      <Title order={3}>Add from URL</Title>
```

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+        User Panels
+      </Title>
+      <Title order={5}>Add Local Panel</Title>
+      <Group align={'flex-end'}>
+        <Select
+          placeholder={'Select Panel'}
+          data={localPanels}
+          onChange={setSelectedPanel}
+          value={selectedPanel}
+          flex={3}
```
WeirdRubberDuck, line 102:
```suggestion
          flex={1}
```

Flex 1 here gives the same result as 3, as the button has no set value (and hence will not grow). So set it to 1 to reduce confusion

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+  return (
+    <Container>
+      <Title my={'xs'} order={4}>
+        User Panels
+      </Title>
+      <Title order={5}>Add Local Panel</Title>
+      <Group align={'flex-end'}>
+        <Select
+          placeholder={'Select Panel'}
```
WeirdRubberDuck, line 98: In the other places, we use sentence casing for the placeholders, not title case. So let's do that there too
```suggestion
          placeholder={'Select panel'}
```

(Personally I also prefer that, since that makes it clearer what is the title and what is the "content" :) )

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+          <UserPageIcon />
+          <Text mx={'xs'}>Add</Text>
+        </Button>
+      </Group>
+      <Divider my={'md'} />
+      <Title order={5}>Add from URL</Title>
+      <TextInput
+        value={urlPanelTitle}
+        label={'Title (optional)'}
+        placeholder={'Input Title (optional)'}
```
WeirdRubberDuck, line 115:
```suggestion
        placeholder={'Input title (optional)'}
```
see previous comment

src/redux/userpages/userPagesSlice.ts
```diff
@@ -0,0 +1,28 @@
...
+const initialState: UserPagesState = {
+  isInitialized: false,
+  panels: []
+};
+
+export const userPagesSlice = createSlice({
+  name: 'userPages',
+  initialState,
+  reducers: {
+    // Use `PayloadAction` to declare the contents of `action.payload`
```
WeirdRubberDuck, line 17:
```suggestion
```

Don't think this comment is needed

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+        onChange={(evt) => setUrlPanelTitle(evt.target.value)}
+      />
+      <Group align={'flex-end'} justify={'space-between'}>
+        <TextInput
+          value={panelURL}
+          label={'URL'}
+          placeholder={'Input URL'}
+          onChange={(evt) => setPanelURL(evt.target.value)}
+          onKeyDown={(e) => e.key === 'Enter' && addWebPanel()}
+          flex={3}
```
WeirdRubberDuck, line 125:
```suggestion
          flex={1}
```

WeirdRubberDuck (approved):
Works great! Nice with the addition of adding a title. 

Another suggested addition would be that the panel remembers which sites have been recently opened and shows them in a clickable list, to make it easier to open them again later. This is currently a bit annoying, especially for web URLs, since once the window is closed, there is no way to open it again.  It could be considered a possible future improvement, but felt like a relatively easy addition that greatly enhances the usefulness. 

 The scrolling seems to work for me, but the scrollbar looks weird. Is that what you mean? Either way, it doesn't seem super important :) 
 
image

src/panels/UserPanelsPanel/UserPanel.tsx
```diff
@@ -0,0 +1,15 @@
+import { Flex } from '@mantine/core';
+
+interface Props {
+  src: string;
+  title: string;
+}
+
+export function UserPage({ src, title }: Props) {
```
WeirdRubberDuck, line 8: I would rename this file to `UserPage.tsx`, to match the component name

And also consider renaming the `UserPanelsPanel` to just `UserPanels`, to get rid of the double "Panels" in the naming. 

OBS! We need to talk about this naming scheme! Sometimes we use "Panel" in the name, and sometimes we do not (e.g. Scene, SessionRecording)

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+    const getUserPanels = async () => {
+      if (!luaApi) return;
+
+      // eslint-disable-next-line no-template-curly-in-string
+      const root = await luaApi.absPath('${USER}/webpanels');
+      const folders = await luaApi.walkDirectoryFolders(root);
+
+      if (!folders) return;
+
+      const slash = navigator.platform.indexOf('Win') > -1 ? '\\' : '/';
```
WeirdRubberDuck, line 45: My VS Code warns that `platform` is deprecated. 
image

So this should probably be implemented some other way.

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
WeirdRubberDuck: Here I also get a deprecation warning for `substr`, so using that version of the function should probably also be avoided. If you want to keep using that version of the code, replace it with `substring`

image

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
WeirdRubberDuck: Use the path module and `path.dirname` instead?

https://stackoverflow.com/questions/818576/get-directory-of-a-file-name-in-javascript

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,134 @@
...
+          value={selectedPanel}
+          flex={3}
+          onKeyDown={(e) => e.key === 'Enter' && addLocalPanel()}
+        />
+        <Button onClick={addLocalPanel} disabled={!selectedPanel} p={'xs'}>
+          <UserPageIcon />
+          <Text mx={'xs'}>Add</Text>
+        </Button>
+      </Group>
+      <Divider my={'md'} />
```
WeirdRubberDuck, line 110:
```suggestion
      <Divider my={'xs'} />
```

All other dividers currently is `xs` sized, so maybe stick with that for consistency :)

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
engbergandreas: I'd prefer that we stick with braced version

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
ylvaselling: It will work - clarified the code a bit

WeirdRubberDuck:
Nice with the addition of the recently opened URLs!  Might I also request that we sort them so that the most recently opened one is at the top? ;)

engbergandreas:
Really nice with the recently opened. Should we also add the option to remove URLs as well?

src/panels/UserPanelsPanel/UserPanelsPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+        placeholder={'Input title (optional)'}
+        onChange={(evt) => setUrlPanelTitle(evt.target.value)}
+      />
+      <Group align={'flex-end'} justify={'space-between'}>
+        <TextInput
+          value={panelURL}
+          label={'URL'}
+          placeholder={'Input URL'}
+          onChange={(evt) => setPanelURL(evt.target.value)}
+          onKeyDown={(e) => e.key === 'Enter' && addWebPanel()}
```
engbergandreas, line 137: I think we should try to be consistent with the event parameter, should we do `e`, `event`, or some other variant? This goes for all places in code

PR #24, opened 2024-12-11
Feature/missions panel

engbergandreas:
Adds the mission panel, there are some TODOs that I would like to discuss on how to solve


src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -0,0 +1,228 @@
...
+            <>
+              <Text>{title}</Text>
+              <Text c={'dimmed'}>{timeString}</Text>
+              <Text my={'xs'}>{displayedPhase.data.description}</Text>
+              {displayedPhase.data.link && (
+                <Button onClick={() => console.log(displayedPhase.data.link)}>
+                  TODO: fix Read more
+                </Button>
+              )}
+              {displayedPhase.data.image && (
```
WeirdRubberDuck, line 202: THere seems to be something wrong with this check, as there are cases when it is rendered even if there is no image. 

E.g. in Apollo profile: 
image

src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -0,0 +1,228 @@
...
+            <Button
+              onClick={() => setDisplayCurrentPhase((prevState) => !prevState)}
+              variant={displayCurrentPhase ? 'filled' : 'light'}
+            >
+              Current Phase
+            </Button>
+          </Group>
+          {displayedPhase.data ? (
+            <>
+              <Text>{title}</Text>
```
WeirdRubberDuck, line 194: This should probably be a `Title` of some sort (h2?) ? Instead of a pure text. 

For information hierarchy reasons (useful or screen readers), at least. We can style it to make it smaller later

src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -0,0 +1,228 @@
...
+              <Text c={'dimmed'}>{timeString}</Text>
+              <Text my={'xs'}>{displayedPhase.data.description}</Text>
+              {displayedPhase.data.link && (
+                <Button onClick={() => console.log(displayedPhase.data.link)}>
+                  TODO: fix Read more
```
WeirdRubberDuck, line 199: Don't forget to do this ;)   Should be an `Anchor` component, I guess?

src/panels/MissionsPanel/MissionTimeButtons.tsx
```diff
@@ -0,0 +1,67 @@
...
+export function MissionTimeButtons({
+  isMissionOverview,
+  currentPhase
+}: MissionTimeButtonsProps) {
+  const luaApi = useOpenSpaceApi();
+  // Since this is being used by the MissionPanel which already subscribes to time, we
+  // and this component re-renders every second due to parent being re-rendered,
+  // we don't need to do any subscription here- TODO: do we want to anyways? obs this will be
+  // one subscription for every button we have...
+  const now = useAppSelector((state) => state.time.timeCapped);
```
WeirdRubberDuck, line 24: I looked at the function for where this is used and it contains a lot of logic. It seems a bit "magic" that we interpolate if the time difference is less than a day, or jump otherwise. I wonder if there was some specific motivation behind that, and if the interpolation case is really necessary. 

Either way, maybe we could make a lua function that does this logic instead, including the fading?  That is, make a `openspace.time.jumpToTime` function, that works similarly to `openspace.pathnavigation.jumpTo`, `openspace.globebrowsing.jumpToGeo` etc.. That could be useful in more situations, and it would simplify the logic as well as get rid of the dependency of the current time.

WeirdRubberDuck: It would also give the user control over how long the fading takes, and make it consistent with the other transitions :) 

So I'd say, let's do it!

src/panels/MissionsPanel/TimeLine/Circle.tsx
```diff
@@ -0,0 +1,36 @@
...
+  marginLeft: number;
+  scale: number;
+}
+export function Circle({ capture, yScale, marginLeft, scale }: CircleProps) {
+  const now = useAppSelector((state) => state.time.timeCapped);
+  const luaApi = useOpenSpaceApi();
+  const circleRadius = 3;
+
+  return (
+    <Tooltip label={'Insturment Activity'}>
```
WeirdRubberDuck, line 21:
```suggestion
    <Tooltip label={'Instrument Activity'}>
```

WeirdRubberDuck: Do we have data about what type of activity? This information seems pretty limited

src/panels/MissionsPanel/TimeLine/Circle.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <ellipse
+        cx={marginLeft - 5}
+        cy={yScale(new Date(capture))}
+        rx={circleRadius}
+        ry={circleRadius / scale}
+        fill={'yellow'}
+        onClick={(event) => {
+          if (event.shiftKey) {
+            jumpToTime(now, capture, luaApi);
```
WeirdRubberDuck, line 30: See previous comment about making a function in the Lua API for this

src/panels/MissionsPanel/TimeLine/Polygon.tsx
```diff
@@ -0,0 +1,63 @@
...
+      label={
+        <>
+          <Text fw={'bold'}>Milestone</Text>
+          {milestone.name}
+        </>
+      }
+    >
+      <polygon
+        points={`0, ${w * 0.5} ${w * 0.5}, ${w} ${w}, ${w * 0.5} ${w * 0.5}, 0`}
+        transform={`translate(${x + centerOffsetX}, ${yPos})scale(1, ${1 / scale})`}
```
WeirdRubberDuck, line 50: Wondering if we should stick to the same code style policy as in C++ here, when we know that we want non-integer values for the numbers. Thoughts?

E.g:
```suggestion
        transform={`translate(${x + centerOffsetX}, ${yPos})scale(1.0, ${1.0 / scale})`}
```

src/panels/MissionsPanel/TimeLine/Rectangle.tsx
```diff
@@ -0,0 +1,80 @@
...
+  yScale: ScaleTime<number, number, never>;
+  nestedLevels: number;
+  setDisplayedPhase: (phase: DisplayedPhase) => void;
+  phase: Phase;
+  nestedLevel: number;
+  padding?: number;
+  color?: React.CSSProperties['color'];
+}
+
+export function Rectangle({
```
WeirdRubberDuck, line 24: Should this be called `PhaseRectangle` ? Since it is specifically used for phases?

src/panels/MissionsPanel/TimeLine/Rectangle.tsx
```diff
@@ -0,0 +1,80 @@
...
+  xScale,
+  yScale,
+  nestedLevels,
+  setDisplayedPhase,
+  phase,
+  nestedLevel,
+  padding = 0,
+  color
+}: RectangleProps) {
+  const now = useAppSelector((state) => state.time.timeCapped);
```
WeirdRubberDuck, line 35: See the same comment as before. Would be creat to not have to depend on `now` in all the components. 

Here it seems like it might still be needed, however? Although, to avoid constant rerendering you could return `isCurrent` instead: 
```suggestion
  const isCurrent = useAppSelector((state) => {
    const now = state.time.timeCapped;
    const isBeforeEndTime = now < endTime.valueOf();
    const isAfterBeginning = now > startTime.valueOf();
    return isBeforeEndTime && isAfterBeginning;
  });
```
Of that would mean moving the variables for start and end time to before this selector

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,327 @@
...
+  };
+
+  // Depth of nesting for phases
+  const nestedLevels = allPhasesNested.length;
+  // Minimum width of a phase
+  const minLevelWidth = 20;
+  // Ensure graph is large enough to show all phases
+  const minWidth = minLevelWidth * nestedLevels + margin.left + margin.right;
+  // Height of graph
+  //   const height = fullHeight - zoomButtonHeight;
```
WeirdRubberDuck, line 75: Remove? Or add a TODO comment that we want to revive the full height

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,327 @@
...
+  ];
+  const xScale = scaleLinear()
+    .range([margin.left, width - margin.right])
+    .domain([0, nestedLevels]);
+  const yScale = scaleUtc([height - margin.bottom, margin.top]).domain(timeRange);
+
+  // Calculate axes
+  const yAxis = axisLeft(yScale);
+
+  // TODO: we dont need the xAxis since we're not plotting anything on it anyways
```
WeirdRubberDuck, line 106: Remove, then?

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,327 @@
...
+                key={`${phase.name}-${phase.timerange.start}-${phase.timerange.end}`}
+                scale={scale}
+                xScale={xScale}
+                yScale={yScale}
+                nestedLevels={nestedLevels}
+                setDisplayedPhase={setDisplayedPhase}
+                phase={phase}
+                nestedLevel={index}
+              />
+            ); //createRectangle(phase, index);
```
WeirdRubberDuck, line 218: ? 
```suggestion
            );
```

src/panels/MissionsPanel/util.ts
```diff
@@ -0,0 +1,42 @@
...
+  fadeTime: number = 1
+): void {
+  if (currentTime === undefined) {
+    return;
+  }
+
+  const utcDate = new Date(time);
+
+  if (!isDateValid(utcDate)) {
+    return;
```
WeirdRubberDuck, line 16: Just a thought: We should maybe start distinguishing between which of our early outs are "valid" ones, and which are not. For those that are not, we should throw an error with a meaningful message instead.  

Something to consider

src/panels/MissionsPanel/util.ts
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  const timeDiffInSeconds = Math.abs(currentTime - utcDate.valueOf()) / 1000;
```
WeirdRubberDuck, line 19:
```suggestion
  const timeDiffInSeconds = Math.abs(currentTime - utcDate.valueOf()) / 1000.0;
```
? 

See previous comment? Would like to discuss if we want to use the same coding style for numbers here as we do in C++

src/redux/actions/actionsSlice.ts
```diff
@@ -73,23 +73,19 @@ export const actionsSlice = createSlice({
...
+      const indexPath = state.actions.findIndex(
         (element) => element.guiPath === state.navigationPath
       );
-      const newNavigationPath = indexPath < 0 ? '/' : state.navigationPath;
-      return {
-        ...state,
-        data: [...newData],
-        navigationPath: newNavigationPath
-      };
+      state.navigationPath = indexPath < 0 ? '/' : state.navigationPath;
```
WeirdRubberDuck, line 88:
```suggestion
      if (indexPath === -1) {
        state.navigationPath = '/';
      }
```
 Checking for -1 explicitly is clearer here, since that's the expected return value. And we don't need to set the value if that was not the case

WeirdRubberDuck (commented):
Felt like I skimmed through the code a bit, so consider this a partial review.  May want to look at it again after some of the discussions

WeirdRubberDuck:
Not a fan of the fading that goes on on the timeline, or at least not when the timeline is entirely zoomed out. It looks like there might be more, even when there isn't

image

WeirdRubberDuck:
Things to think about:   The timeline can currently only be interacted with using keyboard.

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,327 @@
...
+    if (!zoomRef.current) {
+      return;
+    }
+    select(svgRef.current)
+      .transition()
+      .duration(transitionDuration)
+      .call(zoomRef.current.scaleBy, zoomValue);
+  }
+
+  function centerTime() {
```
WeirdRubberDuck, line 175: This function has a weird "zooming out" motion that I find very confusing. I know that that's how it worked in the old UI, but the transition happenes in a way so that its hard to see what is going on with the time (with zooming and translation happening simultaneously)

I think that expected behavior for me would be that it just "scrolls" (translated) to where the time indicator is, without zooming. Or first zooms out, then moves to show the indicator, then zooms in

src/panels/MissionsPanel/MissionContent.tsx
engbergandreas: In this case the link is dead and gives 404: https://www.nasa.gov/sites/default/files/styles/full_width_feature/public/thumbnails/image/s69-31739orig.jpg

src/panels/MissionsPanel/TimeLine/Rectangle.tsx
engbergandreas: I'm on board with this, however, the whole timeline is being rerendered every second since the `MissionContent` is also subscribed to time. We should perhaps go back to this component later and see if we can avoid the constant rerenders?

ylvaselling:
Did some changes... feel free to test it out

src/panels/MissionsPanel/MissionContent.tsx
WeirdRubberDuck: Hm, in the old UI, it wasn't shown, though 🤔 
image

We should decide what the expected behavior is in this case. I there is an image but the link is broken (as for 3/4 of the phases in the Apollo profile), maybe we do want to show it but in a way that highlights the error. So that we know that we need to fix it

src/panels/MissionsPanel/MissionPhase.tsx
```diff
@@ -0,0 +1,77 @@
...
+        {!isMissionOverview &&
+          displayedPhase.data?.actions?.map((uri) => (
+            <ActionsButton key={uri} uri={uri} />
+          ))}
+        {/* We always want to show the actions for the whole mission */}
+        {missionOverview?.actions?.map((uri) => <ActionsButton key={uri} uri={uri} />)}
+      </Flex>
+    </>
+  ) : (
+    <Text> No data for this time range </Text>
```
WeirdRubberDuck, line 80: Seems like this should be moved to a higher level, i.e. to `MissionContent`, as the mission phase itself does not know about the current time range.

WeirdRubberDuck (changes requested):
Went through the code and tested it out and it mostly looks great, but some of the changes are quite confusing to me.  I tried to comment in the code where relevant, but there are another couple of things that I thought should be addressed:  

1. The TimeIndicator changes to a pointing hand on hover, which makes it look interactive even if isn't. Maybe it could show a tooltip that says that it shows the current time (and maybe show the timestamp?

    * Another thing related to the indicator is that it is pretty difficult to see, IMO.  Would it maybe be better if it spanned the entire width of the timeline? E.g. across the axis that shows the month etc. . A benefit with that would be that it would also be visible even during an ongoing phase 🙂 


2. The buttons for the actions seem to have somehow disappeared from the Actions panel!  This has to be fixed :)

tsconfig.app.json
```diff
@@ -4,7 +4,7 @@
     "composite": true,
     "target": "ES2020",
     "useDefineForClassFields": true,
-    "lib": ["ES2021", "DOM", "DOM.Iterable"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
WeirdRubberDuck, line 7: What's the motivation for this change?  And do we maybe want to use `EsNext` instead of a fixed value so we don't have to update it manually as soon as we want to use a new version? 

I'm not very familiar with this setting, however, so this could use some discussion

tsconfig.app.json
```diff
@@ -4,7 +4,7 @@
     "composite": true,
     "target": "ES2020",
```
WeirdRubberDuck, line 5: And should we also change the target?

WeirdRubberDuck (approved):
OMG I love the draggable time indicator

One small comment is that I would personally prefer if the text in these buttons would need to wrap when it does not fit. I.e. the button would then be twice as high
image

src/panels/MissionsPanel/MissionTimeButtons.tsx
```diff
@@ -0,0 +1,64 @@
...
+  currentPhase: DisplayedPhase;
+}
+
+export function MissionTimeButtons({ currentPhase }: MissionTimeButtonsProps) {
+  const jumpToTime = useJumpToTime();
+  const isMissionOverview = currentPhase.type === DisplayType.Overview;
+
+  function jumpToEndOfPhase() {
+    if (currentPhase.type === DisplayType.Phase || isMissionOverview) {
+      jumpToTime(currentPhase.data.timerange.end);
```
WeirdRubberDuck, line 19: This should set the time to _just before_ the phase ends, in my opinion. Now there is some weirdness going on if if one has the "current phase" toggle selected and click this button. 

It will either just be empty, or even result in some weird flickering behavior

https://github.com/user-attachments/assets/685ff590-aeae-4ecc-80e7-7823e2da3714

ylvaselling: I can't reproduce this 🤔

engbergandreas:
Discovered a minor bug If we are outside the time range and click the arrow it will show the given time range, however since the zoom and translation are fixed to the mission time range we are sent back. This might cause confusion so we should restrict the arrow snapping to only be within the timer ange as well?

https://github.com/user-attachments/assets/00f84056-36fb-48aa-93a6-3bf851ad5a1f

tsconfig.app.json
engbergandreas: We should avoid using ESNext since it could mean we find ourselves with deprecated unintended function changes. I don't remember if it was me or Ylva who changed it but it has to do with what JavaScript functions we have available, ES2023 adds some utility for Arrays for example: https://www.typescriptlang.org/tsconfig/#lib which is most likely the case here.

engbergandreas: Could also be that the `tsconfig.node.json` has the ´ES2023` as lib which I've changed for the same reason as above, so these should at least match each others

src/panels/MissionsPanel/TimeLine/TimeIndicator.tsx
```diff
@@ -0,0 +1,110 @@
+import { forwardRef, useEffect, useRef, useState } from 'react';
+import { Tooltip } from '@mantine/core';
+import { useThrottledCallback } from '@mantine/hooks';
+import { drag, ScaleTime, select } from 'd3';
+
+import { useOpenSpaceApi } from '@/api/hooks';
+
+import { TimeIndicatorConfig } from './config';
+
+interface TimeIndicatorProps {
```
engbergandreas, line 10:
```suggestion
interface Props {
```

src/panels/MissionsPanel/TimeLine/TimeIndicator.tsx
```diff
@@ -0,0 +1,110 @@
...
+  yScale: ScaleTime<number, number, never>;
+  timelineWidth: number;
+  scale: number;
+  now: number;
+}
+
+export const TimeIndicator = forwardRef<SVGRectElement, TimeIndicatorProps>(
```
engbergandreas, line 17:
```suggestion
export const TimeIndicator = forwardRef<SVGRectElement, Props>(
```

src/panels/MissionsPanel/TimeLine/TimeIndicator.tsx
```diff
@@ -0,0 +1,110 @@
...
+  function TimeIndicator({ yScale, timelineWidth, scale, now }: TimeIndicatorProps, ref) {
```
engbergandreas, line 18:
```suggestion
  ({ yScale, timelineWidth, scale, now }: Props, ref) => {
```

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,281 @@
...
+import { ActivityCircle } from './ActivityCircle';
+import { TimeLineConfig } from './config';
+import { MileStonePolygon } from './MilestonePolygon';
+import { Phases } from './Phases';
+import { TimeArrow } from './TimeArrow';
+import { TimeIndicator } from './TimeIndicator';
+
+import './TimeLine.css';
+
+interface TimeLineProps {
```
engbergandreas, line 29:
```suggestion
interface Props {
```

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,281 @@
...
+  missionOverview: Phase;
+  setDisplayedPhase: (phase: DisplayedPhase) => void;
+}
+
+export function TimeLine({
+  allPhasesNested,
+  displayedPhase,
+  missionOverview,
+  setDisplayedPhase
+}: TimeLineProps) {
```
engbergandreas, line 41:
```suggestion
}: Props) {
```

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,281 @@
...
+  const nestedLevels = allPhasesNested.length;
+
+  // Ensure graph is large enough to show all phases
+  const minWidth = minLevelWidth * nestedLevels + margin.left + margin.right;
+
+  // Height of graph
+  // TODO anden88: right now we don't have a way to get the size of the panel we're in so we can't
+  // do any updates related to a panel resize and therefore use a fixed height.
+  // When height changes of window, rescale y axis
+  // const height = fullHeight - zoomButtonHeight;
```
engbergandreas, line 71: Now that we can get the height, do we want to do this straight away or in the future?

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,281 @@
...
+      new Date(missionOverview.timerange.start),
+      new Date(missionOverview.timerange.end)
+    ];
+    const yScale = scaleUtc([height - margin.bottom, margin.top]).domain(timeRange);
+
+    const yAxis = axisLeft(yScale);
+    return [yAxis, yScale];
+  }, [missionOverview, margin.bottom, margin.top, height]);
+
+  // TODO: When we get the height from the panel we can update the timeline size as well.
```
engbergandreas, line 92: Related

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,281 @@
...
+  useEffect(() => {
+    if (yAxisRef.current) {
+      select(yAxisRef.current).call(yAxis);
+      // TODO set font family? Previously was 'Segoe UI' font on ticks
+      select(yAxisRef.current).selectAll('.tick text').style('font-size', '1.3em');
+    }
+  }, [yAxis]);
+
+  // Add zoom and update it every the the y scale changes
+  // (TODO: panel resize)
```
engbergandreas, line 113: related?

src/panels/MissionsPanel/TimeLine/TimeLine.tsx
```diff
@@ -0,0 +1,269 @@
...
+    // Min and max scale
+    const scaleExtent: [number, number] = [1, maxScale];
+    // Min and max translation
+    const translateExtent: [[number, number], [number, number]] = [
+      [0, 0],
+      [width, height - margin.bottom]
+    ];
+    zoomRef.current = zoom<SVGElement, unknown>()
+      .on('zoom', (event) => {
+        // TODO: event is any here which is super annoying, try to find a way to get this
```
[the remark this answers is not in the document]
engbergandreas: it is _objectively_ super annoying 😄 

but yes I agree, its nice with the timestamps

src/panels/MissionsPanel/MissionCaptureButtons.tsx
```diff
@@ -0,0 +1,65 @@
+import { Button, Group } from '@mantine/core';
+
+import { useSubscribeToTime } from '@/api/hooks';
+import { Phase } from '@/types/mission-types';
+
+import { useJumpToTime } from './hooks';
+
+interface MissionCaptureButtonsProps {
```
engbergandreas, line 8:
```suggestion
interface Props {
```

src/panels/MissionsPanel/MissionCaptureButtons.tsx
```diff
@@ -0,0 +1,65 @@
...
+  mission: Phase;
+}
+
+export function MissionCaptureButtons({ mission }: MissionCaptureButtonsProps) {
```
engbergandreas, line 12:
```suggestion
export function MissionCaptureButtons({ mission }: Props) {
```

src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+// TODO anden88: for some reason if this was exported from @types file intelisense showed
+// displayedPhase variable as any, aka no type completion :/
+export type DisplayedPhase =
+  | { type: DisplayType.Phase; data: Phase }
+  | { type: DisplayType.Overview; data: Phase }
+  | { type: DisplayType.Milestone; data: Milestone }
+  | { type: undefined; data: undefined };
+
+interface MissionContentProps {
```
engbergandreas, line 20:
```suggestion
interface Props {
```

src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -0,0 +1,156 @@
...
+  missionOverview: Phase;
+}
+
+export function MissionContent({ missionOverview }: MissionContentProps) {
```
engbergandreas, line 24:
```suggestion
export function MissionContent({ missionOverview }: Props) {
```

src/panels/MissionsPanel/MissionPhase.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { OpenWindowIcon } from '@/icons/icons';
+import { ActionsButton } from '@/panels/ActionsPanel/ActionsButton';
+import { DisplayType } from '@/types/enums';
+import { Phase } from '@/types/mission-types';
+
+import { MissionCaptureButtons } from './MissionCaptureButtons';
+import { DisplayedPhase } from './MissionContent';
+import { MissionTimeButtons } from './MissionTimeButtons';
+
+interface MissionPhaseProps {
```
engbergandreas, line 12:
```suggestion
interface Props {
```

src/panels/MissionsPanel/MissionPhase.tsx
```diff
@@ -0,0 +1,90 @@
...
+  displayedPhase: DisplayedPhase;
+  missionOverview: Phase;
+}
+
+export function MissionPhase({ displayedPhase, missionOverview }: MissionPhaseProps) {
```
engbergandreas, line 17:
```suggestion
export function MissionPhase({ displayedPhase, missionOverview }: Props) {
```

src/panels/MissionsPanel/MissionTimeButtons.tsx
```diff
@@ -0,0 +1,56 @@
+import { Button, Group } from '@mantine/core';
+
+import { GoToEnd, GoToStart, TimerIcon } from '@/icons/icons';
+import { DisplayType } from '@/types/enums';
+
+import { useJumpToTime } from './hooks';
+import { DisplayedPhase } from './MissionContent';
+
+interface MissionTimeButtonsProps {
```
engbergandreas, line 9:
```suggestion
interface Props {
```

src/panels/MissionsPanel/MissionTimeButtons.tsx
```diff
@@ -0,0 +1,56 @@
...
+  currentPhase: DisplayedPhase;
+}
+
+export function MissionTimeButtons({ currentPhase }: MissionTimeButtonsProps) {
```
engbergandreas, line 13:
```suggestion
export function MissionTimeButtons({ currentPhase }: Props) {
```

PR #25, opened 2024-12-17
add links and actions to help menu

engbergandreas:
The only thing left to do is for this menu now is the `Getting Started Tour` and update the fallback port to 4680 if and when we change this for the project


src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+  const luaApi = useOpenSpaceApi();
+  const openSpaceVersion = useAppSelector((state) => state.version.openSpaceVersion);
+
+  async function toggleLuaConsole() {
+    if (!luaApi) {
+      return;
+    }
+    const visible = (await luaApi.propertyValue('LuaConsole.IsVisible')) as
+      | boolean
+      | undefined;
```
ylvaselling, line 27: use `useGetBoolPropertyValue` instead (above the function) ? Why get it from the engine when we already have it in redux?

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+    luaApi.setPropertyValueSingle('LuaConsole.IsVisible', !visible);
```
ylvaselling, line 28: use invertproperty function instead?

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+    if (!luaApi) {
+      return;
+    }
+    const portProperty = (await luaApi.propertyValue('Modules.WebGui.Port')) as
+      | number
+      | undefined;
+    const port = portProperty ?? 4670; // TODO: Change this to correct port when hosted
+    const addressProperty = (await luaApi.propertyValue('Modules.WebGui.Address')) as
+      | string
+      | undefined;
```
ylvaselling, line 41: use useGetXPropertyValue here as well

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+
+    const link = `http://${address}:${port}`;
+    window.open(link, '_blank');
+  }
+
+  function toggleShutdown() {
+    luaApi?.toggleShutdown();
+  }
+
+  function osVersionNumber(): React.JSX.Element {
```
ylvaselling, line 52: If this function only returns the string we don't need to return a function with JSX

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+      major: number;
+      minor: number;
+      patch: number;
+    }): string {
+      return version.major !== 255 && version.minor !== 255 && version.patch !== 255
+        ? `${version.major}.${version.minor}.${version.patch}`
+        : 'Custom';
+    }
+
+    return <Text>OpenSpave version: {formatVersion(openSpaceVersion)}</Text>;
```
ylvaselling, line 67:
```suggestion
    return <Text>OpenSpace version: {formatVersion(openSpaceVersion)}</Text>;
```

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
-        <Menu.Item>Toggle Console</Menu.Item>
-        <Menu.Item>Open GUI in Browser</Menu.Item>
-        <Menu.Divider />
-        <Menu.Item>Quit OpenSpace</Menu.Item>
-        <Menu.Divider />
-        <Menu.Item>About</Menu.Item>
-      </Menu.Dropdown>
-    </Menu>
+    <>
+      <Modal opened={opened} onClose={close} title={'About'} size={'40%'}>
```
ylvaselling, line 72: Make the title About OpenSpace instead? The title looks very small compared to the header

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,144 @@
...
+        <Grid>
```
ylvaselling, line 73: Couldnt this grid just be a group with two containers? I think itll make it easier to read

src/windowmanagement/HelpMenu.tsx
ylvaselling, line 72: Should we make this Modal its own component?

src/windowmanagement/HelpMenu.tsx
engbergandreas: I can't get the image and text to play nice with a group, leaving it for now

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
+import { Anchor, Container, Grid, Group, Image, Modal, Text, Title } from '@mantine/core';
```
WeirdRubberDuck, line 1: Cleanup unused components here (`Container`, `Group`)

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+import { useAppSelector } from '@/redux/hooks';
+
+interface Props {
+  opened: boolean;
+  close: () => void;
+}
+export function About({ opened, close }: Props) {
```
WeirdRubberDuck, line 9:
```suggestion

export function About({ opened, close }: Props) {
```

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const openSpaceVersion = useAppSelector((state) => state.version.openSpaceVersion);
+
+  function osVersionNumber(): string {
+    if (!openSpaceVersion) {
+      return 'Fetching OpenSpace version...';
+    }
+
+    function formatVersion(version: {
```
WeirdRubberDuck, line 17: Create a type for the version type in types.d.ts? That would make this a a lot easier to read. E.g.

```typescript
export interface OpenSpaceVersionNumber { 
  major: number;
  minor: number;
  patch: number;
}
```

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
...
+      patch: number;
+    }): string {
+      return version.major !== 255 && version.minor !== 255 && version.patch !== 255
+        ? `${version.major}.${version.minor}.${version.patch}`
+        : 'Custom';
+    }
+
+    return `OpenSpace version: ${formatVersion(openSpaceVersion)}`;
+  }
+  return (
```
WeirdRubberDuck, line 29:
```suggestion

  return (
```

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <Modal opened={opened} onClose={close} title={'About OpenSpace'} size={'40%'}>
+      <Grid>
+        <Grid.Col span={4}>
+          <Image src={'openspace-logo.png'} alt={'OpenSpace logo'} w={'100%'} />
+        </Grid.Col>
+        <Grid.Col span={8}>
+          <Title order={1}>OpenSpace</Title>
```
WeirdRubberDuck, line 36: Hoa about using a `Stack` around this column content?  To give some natural spacing :)

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,97 @@
...
+          </Menu.Item>
+          <Menu.Divider />
+          <Menu.Item onClick={toggleLuaConsole} leftSection={<ConsoleIcon />}>
+            Toggle Console
+          </Menu.Item>
+          <Menu.Item onClick={openGuiInBrowser} leftSection={<OpenInBrowserIcon />}>
+            Open GUI in Browser
+          </Menu.Item>
+          <Menu.Divider />
+          <Menu.Item onClick={toggleShutdown} leftSection={<ExitAppIcon />}>
```
WeirdRubberDuck, line 86: Clicking this menu item should open a modal that says something like "Do you really want to exit the application? Confirm / Cancel"

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,97 @@
...
+          <Menu.Item
+            component={'a'}
+            href={'http://data.openspaceproject.com/feedback'}
+            target={'_blank'}
+            leftSection={<FeedbackIcon />}
+          >
+            Send Feedback
+          </Menu.Item>
+          <Menu.Divider />
+          <Menu.Item onClick={toggleLuaConsole} leftSection={<ConsoleIcon />}>
```
WeirdRubberDuck, line 79: The old button had the shortcut for doing this on the menu item, This one should too :) 
You can use this component: https://mantine.dev/core/kbd/

src/types/types.d.ts
```diff
@@ -19,6 +19,12 @@ export interface Keybind {
 
 export type ActionOrKeybind = Action | Keybind;
 
+export interface OpenSpaceVersion {
+  latestOpenSpaceVersion?: { major: number; minor: number; patch: number };
```
WeirdRubberDuck, line 23: See previous comment about making a type for the version number

src/types/openspace-api-js.d.ts
```diff
@@ -449,7 +449,7 @@ export interface openspace {
   /**
    * Returns the value the property, identified by the provided URI. Deprecated in favor of the 'propertyValue' function
    */
-  propertyValue: () => Promise<void>
+  propertyValue: (uri: string) => Promise<unknown>
```
WeirdRubberDuck, line 452: Use the new type :) 
```suggestion
  propertyValue: (uri: Uri) => Promise<unknown>
```

src/components/About/About.tsx
```diff
@@ -0,0 +1,51 @@
...
+          <Text>
+            OpenSpace is open source interactive data visualization software designed to
```
WeirdRubberDuck, line 38:
```suggestion
            OpenSpace is an open-source interactive data visualization software designed to
```

src/redux/version/versionSlice.ts
```diff
@@ -0,0 +1,21 @@
+import { createSlice } from '@reduxjs/toolkit';
+
+import { OpenSpaceVersion } from '@/types/types';
+
+import { getVersion } from './versionMiddleware';
+
+const initialState: OpenSpaceVersion = {};
```
WeirdRubberDuck, line 7: Hm, for the other states we have the type for the state in the slice. Feel like maybe we should do that here too? 

It makes it much easier to see what information is part of the state

src/types/types.d.ts
WeirdRubberDuck: Also, this type (with the three different version states) should be moved to the versionSlice.ts file?

src/types/types.d.ts
```diff
@@ -19,6 +19,12 @@ export interface Keybind {
...
+  openSpaceVersion?: { major: number; minor: number; patch: number };
```
WeirdRubberDuck, line 24: The naming of the type is a bit confusing, since this is the actual openspace version, but the type itself is named "OpenSpaceVersion".  Consider renaming the type

src/windowmanagement/HelpMenu.tsx
```diff
@@ -1,26 +1,97 @@
...
+    useGetBoolPropertyValue('LuaConsole.IsVisible');
+  const [portProperty] = useGetIntPropertyValue('Modules.WebGui.Port');
+  const [addressProperty] = useGetStringPropertyValue('Modules.WebGui.Address');
+
+  function toggleLuaConsole() {
+    setIsConsoleVisible(!isConsoleVisible);
+  }
+
+  function openGuiInBrowser() {
+    const port = portProperty ?? 4670; // TODO: Change this to correct port when hosted
```
WeirdRubberDuck, line 34: This could already be set to the correct port, since we get it form the property anyways right now. Gives us one less thing to care about later on

```suggestion
    const port = portProperty ?? 4680;
```

WeirdRubberDuck (changes requested):

src/windowmanagement/HelpMenu.tsx
engbergandreas: I was thinking about that, is it hardcoded that esc always quit or is that something that users can change? Thinking if I need to access user-specific keybindings?

src/types/types.d.ts
engbergandreas: Hehe so Ylva said the complete opposite. But we just had a new discussion about it and concluded that we should always have an interface with the name `xyzState` to describe the `initialState` of the redux store and any other interfaces or types should be in the @types file. So in this case I'll make a `export type OpenSpaceVersionState = OpenSpaceVersion` in the versionSlice.ts file and keep the OpenSpaceVersion in the types file

src/types/openspace-api-js.d.ts
engbergandreas: I want to move away from manually editing this file as much as possible since any change we make will be overwritten when we generate a new file. So I'll leave it out for now

src/types/types.d.ts
WeirdRubberDuck: Sounds good enough to me :)

src/types/types.d.ts
WeirdRubberDuck: What I meant is NOT that the `{ major: number; minor: number; patch: number };` interface should be moved to the slice, but this one. 

```
{
  latestOpenSpaceVersion?: { major: number; minor: number; patch: number };
  openSpaceVersion?: { major: number; minor: number; patch: number };
  socketApiVersion?: { major: number; minor: number; patch: number };
 }
 
 I don't see a reason why this should be anywhere else than in the slice where it is used

Comment (WeirdRubberDuck) [reply src/types/types.d.ts]: Also, how about. 
```
interface SemanticVersion {
  major: number;
  minor: number;
  patch: number
}
```
?

File: src/windowmanagement/HelpMenu.tsx (hunk 14 above)
Comment (WeirdRubberDuck) [reply src/windowmanagement/HelpMenu.tsx]: I think this specific one is hardcoded.

File: src/types/openspace-api-js.d.ts (hunk 16 above)
Comment (WeirdRubberDuck) [reply src/types/openspace-api-js.d.ts:452]: Ah, sorry!  Didn't see that it was that file :) 

Ignore this comment then

File: src/redux/version/versionSlice.ts (hunk 18 above)
Comment (WeirdRubberDuck) [reply src/redux/version/versionSlice.ts]: To clarify, based on the comment below. I would prefer if here we had it like this: 

```
interface OpenSpaceVersionState {
  latestOpenSpaceVersion?: SematicVersion;
  openSpaceVersion?: SematicVersion;
  socketApiVersion?: SematicVersion;
}
 
const initialState: OpenSpaceVersionState = {};
```

To make it more consistent with the other states we have.  And these three version numbers are not really used anywhere else

WeirdRubberDuck (approved):

PR #29, opened 2025-01-14
Feature/keybinds panel

ylvaselling:
Add keybinds panel.

Tried to minimize the custom css but some is needed.


src/panels/KeybindsPanel/KeybindsPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+        setSelectedKey={setSelectedKey}
+        selectedKey={selectedKey}
+        activeModifiers={activeModifiers}
+      />{' '}
+      <Group my={'md'}>
+        <Title order={3}>Selected keybind:</Title>
+        <KeybindButtons selectedKey={selectedKey} modifiers={activeModifiers} />
+      </Group>
+      <Divider />
+      <Title order={3} my={'md'}>
```
engbergandreas, line 40: I had comments from Emma on this, we should decide how we want to deal properly with lower-order titles when there aren't any higher ones. Something to do with information flow for screen readers? @WeirdRubberDuck

src/panels/KeybindsPanel/KeybindsPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+        Mapped actions
+      </Title>
+      {selectedActions.length > 0 ? (
+        <Grid mx={'xs'}>
+          {selectedActions.map((selectedAction) => (
+            <Paper key={selectedAction.identifier} p={'md'}>
+              <Title order={4} mb={'md'}>
```
engbergandreas, line 47: Same here

engbergandreas (approved):

src/panels/KeybindsPanel/FullKeyboard/FullKeyboard.css
```diff
@@ -0,0 +1,86 @@
...
+.simple-keyboard {
+  background: none;
+}
+
+.hg-theme-default .hg-button {
+  background: var(--mantine-color-dark-9);
+  color: var(--mantine-color-white);
+  border-bottom: 1px solid var(--mantine-color-dark-2);
+}
+/* Button highlights */
```
WeirdRubberDuck, line 20:
```suggestion
}

/* Button highlights */
```

src/redux/actions/actionsSlice.ts
```diff
@@ -80,8 +80,19 @@ export const actionsSlice = createSlice({
   extraReducers: (builder) => {
     builder.addCase(getAllActions.fulfilled, (state, action) => {
       const [actions, keybinds] = splitActionsAndKeybinds(action.payload);
+      const modifiedKeybinds = keybinds.map((keybind) => {
+        const modifiers = Object.values(keybind.modifiers)
+          .map((value, i) => (value ? Object.keys(keybind.modifiers)[i] : null))
+          .filter((value) => value !== null);
+        return {
```
WeirdRubberDuck, line 87:
```suggestion
          .filter((value) => value !== null);
          
        return {
```

src/redux/actions/actionsSlice.ts
```diff
@@ -80,8 +80,19 @@ export const actionsSlice = createSlice({
   extraReducers: (builder) => {
     builder.addCase(getAllActions.fulfilled, (state, action) => {
       const [actions, keybinds] = splitActionsAndKeybinds(action.payload);
+      const modifiedKeybinds = keybinds.map((keybind) => {
```
WeirdRubberDuck, line 83:
```suggestion
      const modifiedKeybinds: Keybind[] = keybinds.map((keybind) => {
```

src/types/types.d.ts
```diff
@@ -12,12 +12,7 @@ export interface Action {
 export interface Keybind {
   action: string;
   key: string;
-  modifiers: {
-    alt: boolean;
-    control: boolean;
-    shift: boolean;
-    super: boolean;
-  };
+  modifiers: string[];
```
WeirdRubberDuck, line 15: Make a type for the allowed modifiers?

src/panels/KeybindsPanel/KeybindsPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+                  </Table.Tr>
+                </Table.Tbody>
+              </Table>
+            </Paper>
+          ))}
+        </Grid>
+      ) : (
+        <Text>
+          {hasSelectedKeys
+            ? 'No action is associated with this keybind.'
```
WeirdRubberDuck, line 80: This check does not seem to work. When I select a key without a keybind I get "No key selected"

image

src/panels/KeybindsPanel/KeybindsPanel.tsx
WeirdRubberDuck: Yes, I think we should stick with a title hierarchy that reflects the information hierarchy of the page, and be consistent in all out panels. The hierarchy of the titles conveys a lot of information to screen reader users. 

https://atendesigngroup.com/articles/guide-accessible-heading-hierarchy

An important note is that `h3` titles like this should not exist without a h2 title, for context (see link above).  So this code here is an issue. 

One alternative could be that we consider the tab name the "h2" title for sighted users, and add visible hidden titles of order 2 for screen reader users.   

Let's discuss IRL perhaps and come up with a scheme that we think will work well for us?

src/panels/KeybindsPanel/KeybindsPanel.tsx
```diff
@@ -0,0 +1,86 @@
...
+    <Container maw={'none'}>
+      <FullKeyboard
+        setSelectedActions={setSelectedActions}
+        setActiveModifiers={setActiveModifiers}
+        setSelectedKey={setSelectedKey}
+        selectedKey={selectedKey}
+        activeModifiers={activeModifiers}
+      />{' '}
+      <Group my={'md'}>
+        <Title order={3}>Selected keybind:</Title>
```
WeirdRubberDuck, line 36: SHould this really be a `Title`? Seems more like a label, IMO

Also, see next comment on title orders

src/panels/KeybindsPanel/KeybindsPanel.tsx
WeirdRubberDuck: Either way, importantly, the Title order should not be used to set a desired size of the title - it should reflect the hierarchy of the site

src/panels/KeybindsPanel/FullKeyboard/FullKeyboard.css
```diff
@@ -0,0 +1,86 @@
+/* Background */
+.simple-keyboard-background {
+  color: black;
+  font-size: 0.8em;
+  background-color: var(--mantine-color-dark-6);
+  border-bottom: 1px solid var(--mantine-color-dark-2);
+  border-radius: 5px;
```
WeirdRubberDuck, line 7:
```suggestion
  border-radius: 5px;
  width: fit-content;
```

To prevent the div to stretch out indefinitely, like it does now: 
image

WeirdRubberDuck (changes requested):
Looks great!  Just some small requests for changes :)

WeirdRubberDuck (approved):

PR #30, opened 2025-01-17
Some quality of life additions to Window and FilterList

ylvaselling:
This pr adds
* Window component which exposes a hook for all panels; width and height. This will make it possible for panels to use calculations for percentage of heights and widths
* Error boundary to the window component and the app; this will prevent white screens and display errors in the UI locally
* Loading blocks component
* FilterList prop: heightPercentage: it makes it possible to make the height relative to the window
* FilterList prop isLoading; will display the loading blocks when loading


src/util/fallbackRenderer.tsx
```diff
@@ -0,0 +1,17 @@
+import { FallbackProps } from 'react-error-boundary';
+import { Alert, Button, Code, Text } from '@mantine/core';
+
+// Call resetErrorBoundary() to reset the error boundary and retry the render.
```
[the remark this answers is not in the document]
WeirdRubberDuck: Also, maybe make a separate folder in util for "errorhandling" and put the file there? I.e. so the path is: ...../util/errorhandling/fallbackRender.tsx

 That would make it clearer that this component/function is used for handling errors, and should not be used for other things?

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -208,7 +208,7 @@ export function OriginPanel() {
           <TelescopeIcon size={'70%'} />
         </ActionIcon>
       </Group>
-      <FilterList height={300}>
+      <FilterList heightPercent={40}>
```
WeirdRubberDuck, line 211: Please make this height bigger 😄 And maybe consider giving it a fixed height based on the size of the window, instead of a relative one?

src/util/fallbackRenderer.tsx
```diff
@@ -0,0 +1,17 @@
...
+export function fallbackRender({ error, resetErrorBoundary }: FallbackProps) {
+  return (
+    <Alert variant={'light'} color={'red'} title={'🚨 Houston, we have a problem...🚨'}>
+      <Text>An error with the following message was thrown:</Text>
+      <Code block my={'md'}>
+        {error.message}
+      </Code>
+      <Button onClick={resetErrorBoundary} mt={'md'}>
+        Try reloading property tree
```
WeirdRubberDuck, line 13: OBS! This button will not always reset the property tree, it depends on the provided reset function.  So this text should be changed, either to something more general or something that could be configured from the properties. 

Since we don't expect users to see this, maybe just "Reset" is fine?
```suggestion
        Reset
```

src/panels/ExoplanetsPanel/ExoplanetsPanel.tsx
```diff
@@ -87,29 +87,24 @@ export function ExoplanetsPanel() {
...
+              <ExoplanetEntry
+                key={`entry${name}`}
+                name={name}
+                isAdded={isAdded(name)}
+                onClick={() => handleClick(name)}
+              />
+            );
+          }}
+          matcherFunc={wordBeginningSubString}
+        ></FilterList.Data>
```
WeirdRubberDuck, line 105:
```suggestion
        />
```

WeirdRubberDuck (changes requested):
Great updates!  

Have some small suggestions for changes, however

src/panels/OriginPanel/OriginPanel.tsx
engbergandreas: Played around with these changes for a bit. I tried something like this as a way to compute it, thought I'm not sure what I think of it since the sibling height is chosen kind of arbitrarily by looking at the current elements, which means anytime it changes through HTML or CSS this will still be fixed. However, I can't think of any other general solution to use instead since we need to determine the height based on the remaining space and siblings which can vary between components. Thoughts? 

```
  function computeFilterListSize(): number {
    const siblingHeight = 115;
    const bottomMargin = 20;
    const filterListHeight = useWindowSize().height - siblingHeight - bottomMargin;
    const minSize = 300
    return Math.max(filterListHeight, minSize);
  }
  ...
  <FilterList height={computeFilterListSize()}>

  ```

It behaves exactly how I would expect it to though so that's something 🤔

File: src/util/fallbackRenderer.tsx (hunk 3 above)
Comment (engbergandreas) [reply src/util/fallbackRenderer.tsx:13]: Perhaps Reload ?

File: src/util/fallbackRenderer.tsx (hunk 1 above)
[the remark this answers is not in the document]
Comment (engbergandreas) [reply src/util/fallbackRenderer.tsx:4]: If anything this should go in the components folder

File: src/windowmanagement/Window/Window.tsx [hunk 5]
```diff
@@ -0,0 +1,33 @@
...
+    <WindowSizeContext.Provider
+      value={{
+        width: width,
+        height: height
+      }}
+    >
+      <ScrollArea h={'100%'} ref={ref}>
+        <ErrorBoundary
+          fallbackRender={fallbackRender}
+          onReset={() => dispatch(reloadPropertyTree())}
```
engbergandreas, line 25: I don't understand the choice of reset function here, it should be general in this case, no? If so, something like a window.reload would be sufficient or if we can come up with some smarter solution to try and reload a specific component

src/windowmanagement/WindowLayout/WindowLayoutProvider.tsx
```diff
@@ -1,7 +1,8 @@
 import React, { useRef } from 'react';
-import { ScrollArea } from '@mantine/core';
 import DockLayout, { BoxData, PanelData, TabData } from 'rc-dock';
 
+import { Window } from '../Window/Window';
```
engbergandreas, line 4:
```suggestion
import { Window } from '@/windowmanagement/Window/Window';
```

src/windowmanagement/Window/Window.tsx
ylvaselling: I just put something there was something, not a lot of thought put into it:) I'll just make a window reload and a todo comment

src/panels/OriginPanel/OriginPanel.tsx
ylvaselling: Made some updates, added a prop option to pass a function to calculate the height. lmk what you think

src/util/fallbackRenderer.tsx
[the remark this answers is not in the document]
ylvaselling: I moved it to components now, to a new folder called ErrorFallback

src/windowmanagement/Window/Window.tsx
```diff
@@ -0,0 +1,34 @@
+import { PropsWithChildren } from 'react';
+import { ErrorBoundary } from 'react-error-boundary';
+import { ScrollArea } from '@mantine/core';
+import { useElementSize } from '@mantine/hooks';
+
+import { fallbackRender } from '@/components/ErrorFallback/fallbackRenderer';
```
WeirdRubberDuck, line 6:
```suggestion
import { fallbackRenderer } from '@/components/ErrorFallback/fallbackRenderer';
```
Rename this to match the file name?

src/components/ErrorFallback/fallbackRenderer.tsx
```diff
@@ -0,0 +1,21 @@
+import { FallbackProps } from 'react-error-boundary';
+import { Alert, Button, Code, Text } from '@mantine/core';
+
+// This function is called by the npm package react-error-boundary when an
```
WeirdRubberDuck, line 4: Nice 👍 Very clear

src/components/LoadingBlocks/LoadingBlocks.tsx
```diff
@@ -0,0 +1,21 @@
...
+      [...Array(n)].map(() => {
+        const min = 50;
+        const max = 100;
+        return `${Math.floor(Math.random() * (max - min + 1)) + min}%`;
+      }),
+    [n]
+  );
+  return widths.map((width, i) => (
+    <Skeleton key={i} height={16} width={width} radius={'xs'} />
+  ));
```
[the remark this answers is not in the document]
WeirdRubberDuck: No no, go ahead and merge this first!  That's completely fine :)  I prefer your version of the component so I would just replace the one I have in that PR with this one. 

But please add the stack, according to my suggestion ;)

WeirdRubberDuck (approved):

src/components/FilterList/FilterList.tsx
```diff
@@ -1,19 +1,56 @@
...
+  heightPercent,
+  isLoading,
+  children
+}: Props) {
+  const { height: windowHeight } = useWindowSize();
+
+  const calculatedHeight =
+    height ??
+    (heightPercent && windowHeight * (heightPercent / 100.0)) ??
+    (heightFunc && heightFunc(windowHeight));
```
engbergandreas, line 49: This is a minor one but I think the linter will complain about this syntax, I believe the correct one is to do `heightFunc?.(windowHeight)`, in any case I like what you did with the props here 👍

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -169,46 +170,53 @@ export function OriginPanel() {
...
-          aria-label={'Select aim'}
-          size={'lg'}
-          style={actionIconStyle(NavigationActionState.Aim)}
-        >
-          <TelescopeIcon size={'70%'} />
-        </ActionIcon>
-      </Group>
-      <FilterList height={300}>
+    <Container>
+      <Stack ref={ref}>
```
engbergandreas, line 187: Do we need the stack here at all, the previous style on the container was from a much older version and did not do anything. I think we can wrap the Groups in a simple div instead.

engbergandreas:
Looks good, don't forget to do a last lint-fix if necessary

engbergandreas (approved):

PR #32, opened 2025-01-24
Adds task bar icons

engbergandreas:
Also makes the task bar horizontally scrollable when menu items don't fit

### DIFF
--- a/src/windowmanagement/TaskBar.tsx
+++ b/src/windowmanagement/TaskBar.tsx
@@ -15,11 +15,14 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
   });
 
   return (
-    <Group
+    <div
       style={{
         backgroundColor: '#00000080',
         height: 60,
-        paddingLeft: 'var(--mantine-spacing-md)'
+        overflowX: 'scroll',
+        display: 'flex',
+        flexWrap: 'wrap',
+        flexDirection: 'column'
       }}
     >
       {checkedMenuItems.map((item) => {
@@ -31,13 +34,15 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
           });
 
         return item.renderMenuButton ? (
-          item.renderMenuButton(item.componentID, handleClick)
+          <div style={{ margin: '0 2px' }}>
+            {item.renderMenuButton(item.componentID, handleClick)}
+          </div>
         ) : (
-          <Button key={item.componentID} size={'xl'} onClick={handleClick}>
+          <Button key={item.componentID} size={'xl'} onClick={handleClick} mx={2}>
             {item.title}
           </Button>
         );
       })}
-    </Group>
+    </div>
   );
 }
--- a/src/windowmanagement/TaskBar.tsx
+++ b/src/windowmanagement/TaskBar.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
-import { Button, Group } from '@mantine/core';
+import { ActionIcon, Button, Flex } from '@mantine/core';
 import { WindowLayoutOptions } from 'src/windowmanagement/WindowLayout/WindowLayout';
+
+import { IconSize } from '@/types/enums';
 
 import { menuItemsDB } from './data/MenuItems';
 
@@ -15,34 +17,48 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
   });
 
   return (
-    <div
+    <Flex
+      gap={2}
       style={{
         backgroundColor: '#00000080',
         height: 60,
-        overflowX: 'scroll',
-        display: 'flex',
-        flexWrap: 'wrap',
-        flexDirection: 'column'
+        whiteSpace: 'nowrap',
+        overflowX: 'scroll'
       }}
     >
       {checkedMenuItems.map((item) => {
-        const handleClick = () =>
+        function handleClick() {
           addWindow(item.content, {
             title: item.title,
             position: item.preferredPosition,
             id: item.componentID
           });
+        }
 
-        return item.renderMenuButton ? (
-          <div style={{ margin: '0 2px' }}>
-            {item.renderMenuButton(item.componentID, handleClick)}
+        return (
+          <div style={{ flex: '0 0 auto' }}>
+            {item.renderMenuButton ? (
+              item.renderMenuButton(item.componentID, handleClick)
+            ) : (
+              <>
+                {item.icon ? (
+                  <ActionIcon
+                    key={item.componentID}
+                    onClick={handleClick}
+                    size={'input-xl'}
+                  >
+                    {item.icon?.(IconSize.lg)}
+                  </ActionIcon>
+                ) : (
+                  <Button key={item.componentID} onClick={handleClick} size={'xl'}>
+                    {item.title}
+                  </Button>
+                )}
+              </>
+            )}
           </div>
-        ) : (
-          <Button key={item.componentID} size={'xl'} onClick={handleClick} mx={2}>
-            {item.title}
-          </Button>
         );
       })}
-    </div>
+    </Flex>
   );
 }
--- a/src/windowmanagement/TaskBarMenuChoices.css
+++ b/src/windowmanagement/TaskBarMenuChoices.css
@@ -1,0 +1,22 @@
+.ToggleTaskBarItemParent {
+  /* Same padding as Mantine Button */
+  padding: 0 var(--mantine-spacing-sm);
+  &:hover {
+    background-color: var(--mantine-color-dark-4);
+  }
+}
+
+.ToggleTaskBarItem {
+  flex-grow: 1;
+
+  .mantine-Checkbox-body {
+    justify-content: space-between;
+    align-items: center;
+  }
+
+  .mantine-Checkbox-labelWrapper {
+    width: 100%;
+    /* Same padding as Mantine Button */
+    padding: calc(var(--mantine-spacing-xs) / 1.5) 0;
+  }
+}
--- a/src/windowmanagement/TaskBarMenuChoices.tsx
+++ b/src/windowmanagement/TaskBarMenuChoices.tsx
@@ -1,7 +1,12 @@
 import { Dispatch, SetStateAction } from 'react';
-import { Checkbox, Menu, Stack } from '@mantine/core';
+import { Checkbox, Group, Menu, Stack } from '@mantine/core';
+
+import { ChevronRightIcon } from '@/icons/icons';
+import { IconSize } from '@/types/enums';
 
 import { menuItemsDB } from './data/MenuItems';
+
+import './TaskBarMenuChoices.css';
 
 interface TaskBarChoicesProps {
   visibleMenuItems: string[];
@@ -27,37 +32,48 @@ export function TaskBarMenuChoices({
       arrowPosition={'center'}
     >
       <Menu.Target>
-        <Menu.Item>Task Bar</Menu.Item>
+        <Menu.Item rightSection={<ChevronRightIcon size={IconSize.sm} />}>
+          Task Bar
+        </Menu.Item>
       </Menu.Target>
       <Menu.Dropdown>
         <Menu.Label>Toggle Task Bar Items</Menu.Label>
-        <Stack gap={'xs'}>
-          {menuItemsDB.map((item) => (
-            <Checkbox
-              key={item.componentID}
-              label={item.title}
-              defaultChecked={checkedMenuItems[item.componentID]}
-              onChange={(event) => {
-                const { checked } = event.currentTarget;
-                setVisibleMenuItems((prevstate) => {
-                  if (checked) {
-                    // This check is necessary because the callback is sometimes called
-                    // twice, meaning we would otherwise add the item multiple times
-                    if (prevstate.includes(item.componentID)) {
-                      return prevstate;
-                    }
-                    prevstate.push(item.componentID);
-                  } else {
-                    const index = prevstate.findIndex((id) => id === item.componentID);
-                    if (index >= 0) {
-                      prevstate.splice(index, 1);
-                    }
-                  }
-                  return [...prevstate];
-                });
-              }}
-            />
-          ))}
+        <Stack gap={0}>
+          {menuItemsDB.map((item) => {
+            return (
+              <Group className={"ToggleTaskBarItemParent"}>
+                {item.icon?.(IconSize.xs)}
+                <Checkbox
+                  className={"ToggleTaskBarItem"}
+                  key={item.componentID}
+                  label={item.title}
+                  labelPosition={'left'}
+                  defaultChecked={checkedMenuItems[item.componentID]}
+                  onChange={(event) => {
+                    const { checked } = event.currentTarget;
+                    setVisibleMenuItems((prevstate) => {
+                      if (checked) {
+                        // This check is necessary because the callback is sometimes called
+                        // twice, meaning we would otherwise add the item multiple times
+                        if (prevstate.includes(item.componentID)) {
+                          return prevstate;
+                        }
+                        prevstate.push(item.componentID);
+                      } else {
+                        const index = prevstate.findIndex(
+                          (id) => id === item.componentID
+                        );
+                        if (index >= 0) {
+                          prevstate.splice(index, 1);
+                        }
+                      }
+                      return [...prevstate];
+                    });
+                  }}
+                />
+              </Group>
+            );
+          })}
         </Stack>
       </Menu.Dropdown>
     </Menu>
--- a/src/windowmanagement/TaskBarMenuChoices.tsx
+++ b/src/windowmanagement/TaskBarMenuChoices.tsx
@@ -1,7 +1,7 @@
 import { Dispatch, SetStateAction } from 'react';
 import { Checkbox, Group, Menu, Stack } from '@mantine/core';
 
-import { ChevronRightIcon } from '@/icons/icons';
+import { ChevronRightIcon, TaskBarIcon } from '@/icons/icons';
 import { IconSize } from '@/types/enums';
 
 import { menuItemsDB } from './data/MenuItems';
@@ -32,7 +32,10 @@ export function TaskBarMenuChoices({
       arrowPosition={'center'}
     >
       <Menu.Target>
-        <Menu.Item rightSection={<ChevronRightIcon size={IconSize.sm} />}>
+        <Menu.Item
+          leftSection={<TaskBarIcon />}
+          rightSection={<ChevronRightIcon size={IconSize.sm} />}
+        >
           Task Bar
         </Menu.Item>
       </Menu.Target>
@@ -41,10 +44,10 @@ export function TaskBarMenuChoices({
         <Stack gap={0}>
           {menuItemsDB.map((item) => {
             return (
-              <Group className={"ToggleTaskBarItemParent"}>
+              <Group className={'ToggleTaskBarItemParent'}>
                 {item.icon?.(IconSize.xs)}
                 <Checkbox
-                  className={"ToggleTaskBarItem"}
+                  className={'ToggleTaskBarItem'}
                   key={item.componentID}
                   label={item.title}
                   labelPosition={'left'}
--- a/src/icons/icons.tsx
+++ b/src/icons/icons.tsx
@@ -49,8 +49,9 @@ export {
   MdFiberManualRecord as RecordIcon,
   MdSettings as SettingsIcon,
   MdStop as StopIcon,
+  MdOutlineCategory as TreeViewIcon,
   MdVideocam as VideocamIcon
 } from 'react-icons/md';
-export { PiBookOpenText as BookIcon, PiTreeView as TreeViewIcon } from 'react-icons/pi';
+export { PiBookOpenText as BookIcon } from 'react-icons/pi';
 export { RiFocus3Line as FocusIcon, RiRouteLine as RouteIcon } from 'react-icons/ri';
 export { VscFeedback as FeedbackIcon } from 'react-icons/vsc';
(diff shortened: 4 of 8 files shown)


WeirdRubberDuck (changes requested):
Please make it work on Chrome as well ;) 
image


Also, side-scroll is a bit uncomfortable. Since there is not any other action bound to the regular scrolling with the scrollwheel here, how about scrolling sideways using the regular scroll?

src/windowmanagement/TaskBar.tsx
```diff
@@ -15,29 +16,51 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
...
+          <div style={{ flex: '0 0 auto' }}>
+            {item.renderMenuButton ? (
+              item.renderMenuButton(item.componentID, handleClick)
+            ) : (
+              <>
+                {item.icon ? (
+                  <ActionIcon
+                    key={item.componentID}
+                    onClick={handleClick}
+                    size={'input-xl'}
```
WeirdRubberDuck, line 50: What's the motivation for `input-xl` here?  Just curious - is just `xl` not enough here?  :)

src/windowmanagement/TaskBar.tsx
```diff
@@ -15,29 +16,51 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
   });
 
   return (
-    <Group
+    <Flex
+      gap={2}
       style={{
         backgroundColor: '#00000080',
```
ylvaselling, line 22: Use "var(--mantine- ... )" ?

src/windowmanagement/TaskBar.tsx
```diff
@@ -15,29 +16,51 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
...
     >
       {checkedMenuItems.map((item) => {
-        const handleClick = () =>
+        function handleClick() {
           addWindow(item.content, {
             title: item.title,
             position: item.preferredPosition,
             id: item.componentID
           });
+        }
```
ylvaselling, line 38: Place this function in the component body instead? And pass item as an argument? That way we don't have to redefine the function each render

src/windowmanagement/TaskBar.tsx
```diff
@@ -15,29 +16,51 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
...
 
-        return item.renderMenuButton ? (
-          item.renderMenuButton(item.componentID, handleClick)
-        ) : (
-          <Button key={item.componentID} size={'xl'} onClick={handleClick}>
-            {item.title}
-          </Button>
+        return (
+          <div style={{ flex: '0 0 auto' }}>
```
ylvaselling, line 41: You need a key here (the outermost element needs the key in a map)

src/windowmanagement/TaskBar.tsx
```diff
@@ -15,29 +16,51 @@ export function TaskBar({ addWindow, visibleMenuItems }: TaskBarProps) {
...
+              item.renderMenuButton(item.componentID, handleClick)
+            ) : (
+              <>
+                {item.icon ? (
+                  <ActionIcon
+                    key={item.componentID}
+                    onClick={handleClick}
+                    size={'input-xl'}
+                  >
+                    {item.icon?.(IconSize.lg)}
```
WeirdRubberDuck, line 52: The `?.` should not be needed here, since we check for `item.icon` above.  Also, it makes this line quite hard to read

And, I know that @ylvaselling and I sometimes disagree about this, but here I would prefer to have the rendered output in a variable to make the code (IMO) more readable. Something like this, where comments explaining each case could really be useful in this scenario

```typescript
const button; 
if (item.renderMenuButton) {
  // If there is a function for rendering a custom button, always use that one
  button = item.renderMenuButton(item.componentID, handleClick);
}
else if (item.icon) {
  // Else, use the provided icon for this task bar button, if there is one
  button = (
    <ActionIcon
      key={item.componentID}
      onClick={handleClick}
      size={'input-xl'}
    >
      {item.icon?.(IconSize.lg)}
    </ActionIcon>
  );
}
else {
  // If no icon was provided, simply use the title as the button text
  button = (
    <Button key={item.componentID} onClick={handleClick} size={'xl'}>
      {item.title}
    </Button>
  );
}
``` 
Or maybe a function that `return`s the button instead of assigning it to a variable. But I hope you get my point. 

However, this might just be personal preference

ylvaselling, line 52: How about calling the icon function something more function sounding? Like `renderIcon`?

WeirdRubberDuck (approved):

src/windowmanagement/TaskBar.tsx
ylvaselling: Also I was thinking whether we want to use `<div>` or `<Box>` from mantine. Box could be nice as it will be uniform and we can use mantine styling variables

src/windowmanagement/TaskBarMenuChoices.css
```diff
@@ -0,0 +1,22 @@
...
+    justify-content: space-between;
+    align-items: center;
+  }
+
+  .mantine-Checkbox-labelWrapper {
+    width: 100%;
+    /* Same padding as Mantine Button */
+    padding: calc(var(--mantine-spacing-xs) / 1.5) 0;
+  }
+}
```
ylvaselling, line 22: Maybe if we move the windows to its own top bar menu we dont need the nested popups and therefore this is unnecessary?:-) or we could have the styling in the tsx file (i assume that everything except for hover can be in the tsx file?)

src/windowmanagement/TaskBar.tsx
engbergandreas: The `Input-xl` of `ActionIcon` matches the size of `xl` in `Buttons` 👍

src/windowmanagement/TaskBarMenuChoices.tsx
```diff
@@ -27,37 +32,51 @@ export function TaskBarMenuChoices({
...
+                  onChange={(event) => {
+                    const { checked } = event.currentTarget;
+                    setVisibleMenuItems((prevstate) => {
+                      if (checked) {
+                        // This check is necessary because the callback is sometimes called
+                        // twice, meaning we would otherwise add the item multiple times
+                        if (prevstate.includes(item.componentID)) {
+                          return prevstate;
+                        }
+                        prevstate.push(item.componentID);
```
ylvaselling, line 64: I don't think it is recommended to modify function arguments like this. Consider spreading the prevstate first before modifying it

src/windowmanagement/TaskBarMenuChoices.tsx
```diff
@@ -27,37 +32,51 @@ export function TaskBarMenuChoices({
...
+            return (
+              <Group className={'ToggleTaskBarItemParent'}>
+                {item.icon?.(IconSize.xs)}
+                <Checkbox
+                  className={'ToggleTaskBarItem'}
+                  key={item.componentID}
+                  label={item.title}
+                  labelPosition={'left'}
+                  defaultChecked={checkedMenuItems[item.componentID]}
+                  onChange={(event) => {
```
ylvaselling, line 55: Make this a function in the component body? Right now it will create a new function each render

ylvaselling (commented):
Nice! 🎉  It's starting to feel like home again 😄 Wrote some suggestions for changes. We might also want to look at how we are using css... was thinking about this https://mantine.dev/styles/styles-api/

But with the changes to the top bar we talked about, we might not need the nested menus, and then the css might not be needed 🤔

ylvaselling:
image
image

src/windowmanagement/TaskBar.tsx
engbergandreas: Yeah the `?.` was from some previous tests I did woops. I agree that its more readable with a function, especially when there are more than 2 conditionals to test, will update 👍

ylvaselling: Or make the icon a new component?

src/windowmanagement/TaskBarMenuChoices.tsx
WeirdRubberDuck: This would also make the component a bit easier to read, IMO :)

src/windowmanagement/TaskBarMenuChoices.tsx
```diff
@@ -27,37 +32,51 @@ export function TaskBarMenuChoices({
...
-                    if (index >= 0) {
-                      prevstate.splice(index, 1);
-                    }
-                  }
-                  return [...prevstate];
-                });
-              }}
-            />
-          ))}
+        <Stack gap={0}>
```
WeirdRubberDuck, line 44: Why are we not using `Menu.Item` here?  Using a stack of groups, we lose the navigation behavior that there are in other menus (arrow up/down)

engbergandreas: I'm not sure, but it might be because the `Menu.Item` acts as a button and so that would mess with the `Checkbox`. But I'll have to go back and check again, this was done a loooong time ago

src/windowmanagement/TaskBar.tsx
WeirdRubberDuck: Even better ;)

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+export function TaskBar({ addWindow, visibleMenuItems }: Props) {
+  const checkedMenuItems = menuItemsData.filter((item) => {
+    return visibleMenuItems.includes(item.componentID);
+  });
+
+  return (
+    <Flex
+      gap={2}
+      style={{
+        backgroundColor: '#00000080',
```
WeirdRubberDuck, line 22: Consider removing this background color.  It doesn't look great with the OS rendering in the background
image


Also, we should avoid using HEX colors and instead use Mantine's variables

WeirdRubberDuck: Actually, please just remove it :)

src/panels/Menu/TopMenuBar/TopMenu.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+interface Props extends MenuProps, PropsWithChildren {
+  menuItemTabIndex?: -1 | 0;
+  position?: FloatingPosition;
+  offset?: number | FloatingAxesOffsets;
+  withArrow?: boolean;
+  arrowPosition?: ArrowPosition;
+}
+
+export function TopMenu({
```
WeirdRubberDuck, line 18: I know we discussed this, but looking at it now this name seems too generic IMO. It sounds like the entire top menu, and not the individual menus... 

Don't have a good suggestion for another name though... Maybe `MenuWrapper` , `TopStyledMenu` ?   Or similar...?   Naming is so hard...

src/panels/Menu/TaskBar/TaskBar.tsx
engbergandreas: One thing to consider is that it will be unintuitive why you cannot put a panel at the very bottom, for example on the right hand side. And I think it makes the panel look weird that they don't go all the way to the bottom. That's the only reason I added the tint at all

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  return (
+    <Group
+      style={{
+        height: 30,
+        width: '100%',
+        backgroundColor: 'var(--mantine-color-gray-filled)'
+      }}
+    >
```
WeirdRubberDuck, line 57:
```suggestion
    <Group  h={30} w={'100%'} bg={'var(--mantine-color-gray-filled)'}>
```

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,115 @@
...
+          <TaskBarMenuChoices
+            visibleMenuItems={visibleMenuItems}
+            setVisibleMenuItems={setVisibleMenuItems}
+          />
+          <Menu.Item leftSection={<SaveIcon />}>Load/Save Layout</Menu.Item>
+          <Menu.Divider />
+          <Menu.Item leftSection={<VisibilityIcon />}>User Visibility</Menu.Item>
+        </Menu.Dropdown>
+      </TopMenu>
+
```
WeirdRubberDuck, line 111:
```suggestion
```

src/panels/Menu/TopMenuBar/TaskBarMenuChoices.tsx
```diff
@@ -0,0 +1,66 @@
...
+  // Mapping menu item ids to boolean flag to determine their checked status
+  const checkedMenuItems: { [key: string]: boolean } = {};
+  visibleMenuItems.forEach((menuItemID) => {
+    checkedMenuItems[menuItemID] = true;
+  });
+
+  function toggleMenuItem(id: string): void {
+    setVisibleMenuItems((prevstate) => {
+      const index = prevstate.indexOf(id);
+      const isChecked = index >= 0;
```
WeirdRubberDuck, line 28:
```suggestion
      const isChecked = index !== -1;
```
? Alternatively, 
```suggestion
      const isChecked = index > -1;
```

Either way, I think it's much clearer to compare with -1

src/panels/Menu/TaskBar/TaskBar.tsx
engbergandreas: see eg
image

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+  visibleMenuItems: string[];
+}
+
+export function TaskBar({ addWindow, visibleMenuItems }: Props) {
+  const checkedMenuItems = menuItemsData.filter((item) => {
+    return visibleMenuItems.includes(item.componentID);
+  });
+
+  return (
+    <Flex
```
ylvaselling, line 19: To get the styling of the scroll you could use a scroll area instead

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+import { TaskBarMenuButton } from './TaskBarMenuButton';
+
+interface Props {
+  addWindow: (component: React.JSX.Element, options: WindowLayoutOptions) => void;
+  visibleMenuItems: string[];
+}
+
+export function TaskBar({ addWindow, visibleMenuItems }: Props) {
+  const checkedMenuItems = menuItemsData.filter((item) => {
```
ylvaselling, line 14: rename checkedMenuItems to visibleTaskBarButtons or something? Took me a while to understad what checked menu items was.

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -0,0 +1,40 @@
...
+      }}
+      onWheel={(event) => {
+        event.currentTarget.scrollLeft += event.deltaY;
+      }}
+    >
+      {checkedMenuItems.map((item) => {
+        return (
+          <Box key={item.componentID} style={{ flex: '0 0 auto' }}>
+            <TaskBarMenuButton addWindow={addWindow} item={item} />
+          </Box>
```
ylvaselling, line 35: Why do we need a box here? cant we set the style on the taskbarmenubutton?

src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
```diff
@@ -0,0 +1,43 @@
...
+      position: item.preferredPosition,
+      id: item.componentID
+    });
+  }
+
+  if (item.renderMenuButton) {
+    // If there is a function for rendering a custom button, always use that
+    return item.renderMenuButton(item.componentID, () => handleClick(item));
+  } else if (item.renderIcon) {
+    // Otherwise, use the provided icon for this menu item, if the is one
```
ylvaselling, line 25:
```suggestion
    // Otherwise, use the provided icon for this menu item, if there is one
```

src/panels/Menu/TopMenuBar/TaskBarMenuChoices.tsx
```diff
@@ -0,0 +1,66 @@
...
+export function TaskBarMenuChoices({
+  visibleMenuItems,
+  setVisibleMenuItems
+}: TaskBarChoicesProps) {
+  // Mapping menu item ids to boolean flag to determine their checked status
+  const checkedMenuItems: { [key: string]: boolean } = {};
+  visibleMenuItems.forEach((menuItemID) => {
+    checkedMenuItems[menuItemID] = true;
+  });
+
```
ylvaselling, line 24: For some reason I still find the way this is written very confusing... i think the perfomance loss is negligible if we just have a `includes` in later in the `checked={...}` prop. i guess it would be `checked={visibleMenuItems.includes(item.componentID)}`

src/panels/Menu/TopMenuBar/TaskBarMenuChoices.tsx
```diff
@@ -0,0 +1,66 @@
...
+      const index = prevstate.indexOf(id);
+      const isChecked = index !== -1;
+
+      if (isChecked) {
+        prevstate.splice(index, 1);
+        return [...prevstate];
+      } else {
+        return [...prevstate, id];
+      }
+    });
```
ylvaselling, line 36: Consider using mantines hook for state objects, then you dont have to do a deep copy https://mantine.dev/hooks/use-set-state/. Then you could use an object instead of an array with the `visible[id] = true` for example. That would also make the check above easier since you have a O(1) map rather than the array

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface TopMenuBarProps {
+  visibleMenuItems: string[];
+  setVisibleMenuItems: Dispatch<SetStateAction<string[]>>;
+  addWindow: (component: React.JSX.Element, options: WindowLayoutOptions) => void;
+}
+
+export function TopMenuBar({
+  visibleMenuItems,
+  setVisibleMenuItems,
+  addWindow
```
ylvaselling, line 29: This seems to only be passed as a prop - use the provider instead where the addWindow function is needed?

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,108 @@
...
+}: TopMenuBarProps) {
+  const luaApi = useOpenSpaceApi();
+
+  const [isConsoleVisible, setIsConsoleVisible] =
+    useGetBoolPropertyValue('LuaConsole.IsVisible');
+
+  function toggleLuaConsole() {
+    setIsConsoleVisible(!isConsoleVisible);
```
ylvaselling, line 37: usually for toggles i think it is better to use the (old => !old) although here i guess it doesnt really matter since you only manipulate state once

src/windowmanagement/WindowLayout/WindowLayout.tsx
```diff
@@ -69,7 +69,7 @@ export function WindowLayout() {
 
   // Populate default visible items for taskbar
   useEffect(() => {
-    const defaultVisibleMenuItems = menuItemsDB
+    const defaultVisibleMenuItems = menuItemsData
       .map((menuItem) => {
         if (menuItem.defaultVisible) {
           return menuItem.componentID;
```
ylvaselling, line 75: No need to use map here, you can just filter directly

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -0,0 +1,40 @@
+import { Box, Flex } from '@mantine/core';
+import { WindowLayoutOptions } from 'src/windowmanagement/WindowLayout/WindowLayout';
+
+import { menuItemsData } from '@/windowmanagement/data/MenuItems';
+
+import { TaskBarMenuButton } from './TaskBarMenuButton';
+
+interface Props {
+  addWindow: (component: React.JSX.Element, options: WindowLayoutOptions) => void;
```
ylvaselling, line 9: use addwindow from the provider instead? visiblemenuitems could also be part of the provider

src/panels/Menu/TaskBar/TaskBar.tsx
ylvaselling: I think it is better to not have the color as well. I agree that it looks funny with the gap but it is better than with the color imo

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,108 @@
...
+      title: 'Confirm action',
+      children: <Text>Are you sure you want to quit OpenSpace?</Text>,
+      labels: { confirm: 'Quit', cancel: 'Cancel' },
+      confirmProps: { color: 'red' },
+      onConfirm: () => luaApi?.toggleShutdown()
+    });
+  }
+
+  return (
+    <Group h={30} w={'100%'} bg={'var(--mantine-color-gray-filled)'}>
```
ylvaselling, line 51: Entering design territory but i would prefer the color to be darker hehe

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -0,0 +1,108 @@
...
+      <MenuWrapper>
+        <Menu.Target>
+          <Button size={'xs'} color={'gray'}>
```
ylvaselling, line 54: Consider using the transparent variant here. That way we only have to keep tabs on one color, the background. Goes for the rest of them in this file too.

src/windowmanagement/WindowLayout/WindowLayout.tsx
engbergandreas: lol 100% smarter than whatever this was

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
engbergandreas: I'll postpone it till we do the theme task

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
engbergandreas: If it's transparent we lose the mouse-over effect so I'd rather not do that, and then we still have to change the text color from blue anyway so it's a lose-win situation

PR #36, opened 2025-01-30
Fix text wrapping and truncating

### LINKED ISSUE #16: Make focus entry names overflow with `...`  for long names
If the origin panel is smaller than node names it breaks into two rows in a weird way, we should add `...` overflow to cut the long name instead of breaking it into a new row
image

ylvaselling:
Closes #16

### DIFF
--- a/src/panels/OriginPanel/FocusEntry.tsx
+++ b/src/panels/OriginPanel/FocusEntry.tsx
@@ -36,7 +36,7 @@ export function FocusEntry({
       <Button
         onClick={onSelectEntry}
         justify={'left'}
-        flex={'1 1'}
+        flex={1}
         variant={buttonVariant}
       >
         <Text truncate>{entry.name}</Text>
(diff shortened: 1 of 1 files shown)


src/panels/OriginPanel/FocusEntry.tsx
```diff
@@ -32,14 +32,14 @@ export function FocusEntry({
   }
 
   return (
-    <Group justify={'space-between'} gap={'xs'}>
+    <Group justify={'space-between'} gap={'xs'} w={'100%'}>
       <Button
         onClick={onSelectEntry}
         justify={'left'}
-        style={{ flexGrow: 1 }}
+        flex={'1 1'}
```
engbergandreas, line 39:
```suggestion
        flex={1}
```

engbergandreas:
insane how easy it is with Mantine, I tried sooooo many solutions that did not work 🤦‍♂️

WeirdRubberDuck (approved):

PR #38, opened 2025-01-31
Feature/skybrowser panel

ylvaselling:
Adds the skybrowser panel. 

There are many hooks file on different levels. My thought is that these are only as high up in the folder hierarchy as they are used. Therefore there's quite a few of them.

The skybrowser topic has a pretty "dumb" way of sending data. Therefore some comparison functions to mitigate too many updates. 

Using the vite-plugin-svgr plugin for loading svg did not add the animation. Tried some different ways but nothing really worked well... Therefore i changed the svg to just a functional component with the help of ✨ AI ✨ 

Adds:
* Virtual Grid Component
* FilterList Grid Component
* DragDropList with local cache Component
* Exposes size for FilterList
* Makes ResizeableContent styleable

*Future work* 
* The settings panel is not implemented yet. Leaving that for later.
* I removed the updating of the size of the gui window when the ratio changes from OpenSpace. This seemed to me a bit weird to implement in this new gui
* Fixing the white background of the iframe when the border radius of wwt is larger than 0


src/redux/skybrowser/skybrowserMiddleware.ts
```diff
@@ -0,0 +1,67 @@
...
+
+export const setupSubscription = createAsyncThunk(
+  'skyBrowser/setupSubscription',
+  async (_, thunkAPI) => {
+    skybrowserTopic = api.startTopic('skybrowser', {
+      event: 'start_subscription'
+    });
+
+    (async () => {
+      // TODO: add type to the data
```
engbergandreas, line 25: reminder or maybe this will be done later?

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+  roll: number;
+  scale: number;
+  selectedImages: number[];
+  targetId: string;
+}
+
+interface SkyBrowserUpdate
+  extends Pick<SkyBrowserState, 'selectedBrowserId' | 'cameraInSolarSystem'> {
+  browsers: { [id: string]: SkyBrowserBrowser };
+}
```
engbergandreas, line 42: Move to types file?

src/redux/store.ts
```diff
@@ -31,7 +32,8 @@ export const store = configureStore({
     sessionRecording: sessionRecordingReducer,
     time: timeReducer,
     userPanels: userPanelsReducer,
-    version: versionReducer
+    version: versionReducer,
+    skybrowser: skyBrowserReducer
```
engbergandreas, line 36: Put in alphabetical order

engbergandreas:
Getting an error after I opened a skybrowser and closed all windows and then trying to open it up again: 
image

engbergandreas:
A nice feature would be to rename the sky browser instead of the automatic now. Can put this on future improvements though

engbergandreas:
The icons in the wwt does not seem to be interactable?
image

engbergandreas:
I keep getting crashes or extreme slowdown if I've added a browser, and at least 1 image, close both windows and when trying to open the panel again everything freezes. Also looking at the memory firefox consumed 5 GB of RAM after like 5 minutes of usage (adding, removing images, adding additional skybrowsers, trying the search)

src/panels/SkyBrowserPanel/WorldWideTelescope/WwtProvider/types.d.ts
```diff
@@ -0,0 +1,60 @@
...
+  event: 'set_background_color';
+  data: number[];
+}
+
+interface AimMessage {
+  event: 'center_on_coordinates';
+  ra: number;
+  dec: number;
+  fov: number;
+  roll: number;
```
WeirdRubberDuck, line 28: Do we want tot add acomment for each of these ,mentioning what types of values are expected?  I.e. whether the unit is in degrees/radians for example. I think that's good practice at least when dealing with angles

src/panels/SkyBrowserPanel/WorldWideTelescope/WwtProvider/types.d.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+interface BorderRadiusMessage {
+  event: 'set_border_radius';
+  data: number;
+}
+
+interface ColorMessage {
+  event: 'set_background_color';
+  data: number[];
```
WeirdRubberDuck, line 20: Add a comment specifying how this list of numbers is expected to look like? I.e. how many values and also what "unit" (e.g. rgb 255, 0-1, etc)

Also, if we know exactly how many values it should be, maybe do e.g. 
```
[number, number, number]
```
?

src/panels/SkyBrowserPanel/WorldWideTelescope/WwtProvider/types.d.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface LoadCollectionMessage {
+  event: 'load_image_collection';
+  url: string;
+  loadChildFolders: true;
+}
+
+interface BorderRadiusMessage {
+  event: 'set_border_radius';
+  data: number;
```
WeirdRubberDuck, line 15: Add a comment explainng that this is the radius value and what type of value is expected

src/panels/SkyBrowserPanel/WorldWideTelescope/WwtProvider/types.d.ts
```diff
@@ -0,0 +1,60 @@
...
+  roll: number;
+  instant: true;
+}
+
+interface ImageMessage {
+  event: 'image_layer_create';
+  id: string;
+  url: string;
+  mode: 'preloaded';
+  goto: false;
```
WeirdRubberDuck, line 37: what does this mean?   Add a comment about it, perhaps?

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+  useSelectedBrowserCoords,
+  useSelectedBrowserFov,
+  useSelectedBrowserRadius,
+  useSkyBrowserSelectedImages,
+  useSkyBrowserSelectedOpacities
+} from '../hooks';
+
+import { useWwtProvider } from './WwtProvider/hooks';
+
+// These are the hooks that will keep tabs of the redux state and
```
WeirdRubberDuck, line 17: keep tabs?  Do you mean "keep track"?

Since we use the word "tab" for so many things, here especially, I think we should change the phrasing here so it's a bit more clear :)

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+// when it changes, it will pass along these messages to WWT
+
+// The hooks are: images, opacities, aim, border color and border radius
```
WeirdRubberDuck, line 20: Remove this comment or make it part of the comment above and adda a new line between this and the function

Now it looks like it's a comment describing the `updateSelectedImages` function, which is confusing.  Also, the comment would potentially be outdated pretty quickly, so I'd opt for removing it

WeirdRubberDuck: Instead, add a comment above each hook function explaining what that function is responsible for?  That would be useful in this case, I think

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+  const { imageCollectionLoaded, loadImage, removeImage } = useWwtProvider();
+  const imageList = useAppSelector((state) => state.skybrowser.imageList);
+  const selectedImages = useSkyBrowserSelectedImages();
+
+  // Update images in WWT when the selected images changes
+  useEffect(() => {
+    if (
+      imageList.length === 0 ||
+      !imageCollectionLoaded ||
+      selectedImages === undefined
```
WeirdRubberDuck, line 31: I usually prefer making a varialbe instead of such a complex if statement. The if gets hard to read and adds a lot of otherwise unnecessary whitespace when written like this, and a suitable named variable makes the check a lot easier to understand. 

But might be personal preference here

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+    ) {
+      return;
+    }
+    // Brute force this as the performance loss is negligible and there are many complicated cases
```
WeirdRubberDuck, line 35: What is "this"?  Please epxlain :)  It's not obvious to me from the code below and could benefit from a more thorough explanantion

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+  }, [imageList, selectedImages, imageCollectionLoaded, loadImage, removeImage]);
+}
+
+export function useUpdateOpacities() {
+  const { imageCollectionLoaded, setOpacity } = useWwtProvider();
+  const imageList = useAppSelector((state) => state.skybrowser.imageList);
+  const selectedImages = useSkyBrowserSelectedImages();
+  const opacities = useSkyBrowserSelectedOpacities();
+
+  // Update opacities in WWT when the opacities changes
```
WeirdRubberDuck, line 49: same here, this could be moved to above the function and be the documentation for this hoow. 

although, please clarify what "the opacities" are.  The opacities of what?

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+  useEffect(() => {
+    if (imageList.length === 0 || !imageCollectionLoaded || opacities === undefined) {
```
WeirdRubberDuck, line 51:
```suggestion
    if (imageList.length === 0 || !imageCollectionLoaded || !opacities) {
```

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+      return;
+    }
+    // Brute force this as the performance loss is negligible and there are many complicated cases
+    opacities.map((opacity, i) => {
+      if (!selectedImages) {
```
WeirdRubberDuck, line 56: Seems like this check should be moved to outside the .map call.  Part of the if statement above, maybe?

Also, seems like we expect `selectedImages` and `opacities` to have the same size.  Should these two things really be one combined data structure?  To avoid the direct dependency between the two lists.... I assume that `opacities` is the opacity setting for each of the selected images?

src/panels/SkyBrowserPanel/WorldWideTelescope/hooks.ts
```diff
@@ -0,0 +1,103 @@
...
+  const { ra, dec, roll } = useSelectedBrowserCoords();
+  const fov = useSelectedBrowserFov();
+  const { setAim } = useWwtProvider();
+
+  useEffect(() => {
+    if (
+      ra !== undefined &&
+      dec !== undefined &&
+      fov !== undefined &&
+      roll !== undefined
```
WeirdRubberDuck, line 75: Why is this check needed... 🥲  Could we somehow write the code so that it isn't?  It makes me sad

WeirdRubberDuck: Maybe it could simply be handled in the `setAim` function, if we can't prevent checking for undefined

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+} from '@/redux/skybrowser/skybrowserMiddleware';
+import {
+  setActiveImage,
+  setImageCollectionData
+} from '@/redux/skybrowser/skybrowserSlice';
+
+export function useGetWwtImageCollection() {
+  const luaApi = useOpenSpaceApi();
+  const dispatch = useAppDispatch();
+  useEffect(() => {
```
WeirdRubberDuck, line 17: this useeffect is quite hard to read due to the length of the function.  can we maybe move function out, or otherwise simplify it somehow?

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+        if (images.length !== 0) {
+          const imgDataWithKey = images.map((image) => ({
+            ...image,
+            key: image.identifier
+          }));
+          dispatch(setImageCollectionData(imgDataWithKey));
+        }
+      } else {
+        throw new Error('No AAS WorldWide Telescope images!');
+      }
```
WeirdRubberDuck, line 34: This already simplifies it a little
```suggestion
      if (!imgData) {
        throw new Error('No AAS WorldWide Telescope images!');
      }
      const images = Object.values(imgData);
      if (images.length > 0) {
        const imgDataWithKey = images.map((image) => ({
          ...image,
          key: image.identifier
        }));
        dispatch(setImageCollectionData(imgDataWithKey));
      }
      
```

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+    }
+    try {
+      getWwtListOfImages();
+    } catch (e) {
+      throw Error(`Could not load image collection from OpenSpace: ${e}`);
```
WeirdRubberDuck, line 39:
```suggestion
      throw Error(`Could not load image collection from OpenSpace. Error: ${e}`);
```

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+            ...image,
+            key: image.identifier
+          }));
+          dispatch(setImageCollectionData(imgDataWithKey));
+        }
+      } else {
+        throw new Error('No AAS WorldWide Telescope images!');
+      }
+    }
+    try {
```
WeirdRubberDuck, line 36:
```suggestion

    try {
```

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+  const dispatch = useAppDispatch();
+  useEffect(() => {
+    dispatch(subscribeToSkyBrowser());
+    return () => {
+      dispatch(unsubscribeToSkyBrowser());
+    };
+  }, [dispatch]);
+}
+
+export function useActiveImage(): [string, (url: string) => void] {
```
WeirdRubberDuck, line 55: What's the difference between the active and selected image?

WeirdRubberDuck: Oh, maybe the selected are the ones that I think of as "added" ?

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+// the value itself actually updates. TODO (ylvse 2025-01-29): rewrite
+// the topic so that most of the data are properties instead, and only
+// pass the data for the currently selected browser
+
+export function useSelectedBrowserColor(): number[] | undefined {
+  const color = useAppSelector(
+    (state) => state.skybrowser.selectedBrowser?.color,
+    lowPrecisionEqualArray
+  );
+  return color;
```
WeirdRubberDuck, line 77:
```suggestion
  return useAppSelector(
    (state) => state.skybrowser.selectedBrowser?.color,
    lowPrecisionEqualArray
  );
```
?

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+    lowPrecisionEqual()
+  );
+  return radius;
+}
+
+export function lowPrecisionEqualMatrix(
+  lhs: number[][] | undefined,
+  rhs: number[][] | undefined
+) {
+  if (lhs === undefined && rhs === undefined) {
```
WeirdRubberDuck, line 97: All these three functions use the same check, which is a bit nasty/fragile to me.  Could the functions be generalized somehow?

WeirdRubberDuck: Maybe add a function that does this very check?

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+
+export function useSelectedBrowserRadius(): number | undefined {
+  const radius = useAppSelector(
+    (state) => state.skybrowser.selectedBrowser?.borderRadius,
+    lowPrecisionEqual()
+  );
+  return radius;
+}
+
+export function lowPrecisionEqualMatrix(
```
WeirdRubberDuck, line 93: these functions are not hooks, so they should be moved out of this file.  

Seems like they should be placed in some type of general utility file for equal checks?

Also, the "low precision" versions need documentation!  What does "low" precision mean here?

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+export function lowPrecisionEqual(epsilon: number = 1e-3) {
+  return (lhs: number | undefined, rhs: number | undefined) => {
+    if (lhs === undefined && rhs === undefined) {
+      return true;
+    }
+    if (lhs === undefined || rhs === undefined) {
+      return false;
+    }
```
WeirdRubberDuck, line 146: Same here. Move this check to one general function 

With some type flexibility and checks, it should be able to use the same function as the equality functions above

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+    return Math.abs(lhs - rhs) < epsilon;
+  };
+}
+
+export function useSelectedBrowserCoords() {
+  const ra = useAppSelector(
+    (state) => state.skybrowser.selectedBrowser?.ra,
+    lowPrecisionEqual(1e-6)
```
WeirdRubberDuck, line 154: Why is this specific precision needed for ra dec and roll?  Would be a lot nicer if the `lowPrecisionEqual` function didn't have to work differently than the other equality functions

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+  return { ra, dec, roll };
+}
+
+export function useSelectedBrowserFov() {
+  const fov = useAppSelector(
+    (state) => state.skybrowser.selectedBrowser?.fov,
+    lowPrecisionEqual(1e-6)
+  );
+
+  return fov;
```
WeirdRubberDuck, line 174:
```suggestion
  return useAppSelector(
    (state) => state.skybrowser.selectedBrowser?.fov,
    lowPrecisionEqual(1e-6)
  );
```

Also, same comment as above on the precision

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+export function useSkyBrowserCartesianDirection() {
+  const cartesianDirection = useAppSelector(
+    (state) => state.skybrowser.selectedBrowser?.cartesianDirection,
+    lowPrecisionEqualArray
+  );
+  return cartesianDirection;
```
WeirdRubberDuck, line 182:
```suggestion
  return useAppSelector(
    (state) => state.skybrowser.selectedBrowser?.cartesianDirection,
    lowPrecisionEqualArray
  );
```

src/panels/SkyBrowserPanel/hooks.ts
```diff
@@ -0,0 +1,223 @@
...
+}
+
+export function useSkyBrowserColors() {
+  const browserColors = useAppSelector(
+    (state) => state.skybrowser.browserColors,
+    lowPrecisionEqualMatrix
+  );
+  return browserColors;
```
WeirdRubberDuck, line 190:
```suggestion
  return useAppSelector(
    (state) => state.skybrowser.browserColors,
    lowPrecisionEqualMatrix
  );
```

and for the final hooks in this file.. you get it by now :)

src/redux/skybrowser/skybrowserMiddleware.ts
WeirdRubberDuck, line 25: Please do ! :)

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
+import { createSlice, PayloadAction } from '@reduxjs/toolkit';
+
+export interface SkyBrowserImage {
```
WeirdRubberDuck, line 3: These data types (`SkybrowserImage`, `SkyBrowserBrowser` and `SkybrowserState`) would benefit from descriptions of what the different parameters are.  Now one has to know how the skybrowser works to understand that 

Also specify if angles are radians/degrees

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+  opacities: number[];
+  ra: number;
+  ratio: number;
+  roll: number;
+  scale: number;
+  selectedImages: number[];
+  targetId: string;
+}
+
+interface SkyBrowserUpdate
```
WeirdRubberDuck, line 39: What is this type for?  could use some explanantion

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+  extends Pick<SkyBrowserState, 'selectedBrowserId' | 'cameraInSolarSystem'> {
+  browsers: { [id: string]: SkyBrowserBrowser };
+}
+
+export interface SkyBrowserState {
```
WeirdRubberDuck, line 44: I think @engbergandreas and I previously discusse that the redux slices should only have the type for the actual state, and that other types/interfaces should be moved to a types.d.ts. file.  So here I would only keep this one, and move the other types above to another file. 

the motivation is to make the code for the redux slices easier to overview, and make it clearer what information is part of each state in the sotre

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+  isInitialized: boolean;
+  cameraInSolarSystem: boolean;
+  selectedBrowserId: string;
+  selectedBrowser: SkyBrowserBrowser | null;
+  browserIds: string[];
+  browserNames: string[];
+  browserColors: number[][];
```
WeirdRubberDuck, line 51: Smae issue here as before. A color as a list of arbitrary number of numbers is too unspecific. 

Do we need a named type for colors?

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+        state.selectedBrowser.selectedImages = state.selectedBrowser.selectedImages.map(
+          (id) => (typeof id === 'string' ? parseInt(id) : id)
+        );
+      }
+      state.browserIds = Object.keys(action.payload.browsers) ?? [];
+      // Get all colors
+      state.browserColors = state.browserIds.map(
+        (id) => action.payload.browsers[id].color
+      );
+      state.browserNames = state.browserIds.map((id) => action.payload.browsers[id].name);
```
WeirdRubberDuck, line 93: Do we want to combine these three values into one datastructure? so we don't have to keep track of three lists that should have the same length?  Seems less fragile

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      return state;
+    },
+    setImageCollectionData: (state, action: PayloadAction<SkyBrowserImage[]>) => {
+      state.imageList = action.payload;
+      return state;
+    },
+    onCloseConnection: (state) => {
```
WeirdRubberDuck, line 101: Woot, why is this needed?  Should there be a todo comment or something inside of here?

WeirdRubberDuck (changes requested):
Here's phase 1 of my comment - on the code itself. What a beast PR :D 

There will be some comments on the behavior as well, since I had some questions and opinions

src/panels/SkyBrowserPanel/Tabs/SelectedImageCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+            </TabButton>
+            <TabButton
+              onClick={() =>
+                luaApi?.skybrowser.removeSelectedImageInBrowser(
+                  selectedBrowserId,
+                  image.url
+                )
+              }
+              text={'Remove image'}
+              variant={'filled'}
```
WeirdRubberDuck, line 65: How about making this a red outlined button and moving it to the farthest right of the three buttons?

Among the actions (focus image, read about image, and delete) this seems like the one that comes least in the "hiearachy"

WeirdRubberDuck:
And here are the content-related comments: 

1. The "world Wide Telescope" window should not be opened until a browser has been added, IMO
1. The "Add browser" buttons needs some love/styling.  Maybe make it full-width / add some padding? Or at least center it and make it bigger?
1. do we really want to keep using zero indexing for the skybrowsers?  At least in the UI, we should probably start with "Sky browser 1", not "Sky browser 0"
1. We never want to simply rely on color for identifying something. this is currently a problem in the World Wide telescope Window. The name of the browser could maybe be added to the tab?  If I didn't see color, it would behard to know which sky browser was showing in the window here:
image
1. I also found it very confusing that the same window was used to show all added browsers. What if I want to look at two parts of the sky at the same time??   We should open a new window for each browser, when clicking the "Open Telescope View", and maybe only ad the first one per default?  (or the subsequent ones as tabs in the same window or something). I do see how this may become complicated though, so it should be done with care
1. We should add tooltips to all truncated texts, like these ones in the image list, that show the full name on hover.  Now it's hard to know how to find out what the name of the image actually is
image
1. The image part of the items in the list in the image above is a bit small., especially compared to the parts with the buttons. can we maybe discuss and consider an alternative design, that makes the image more visibly dominant?  
    - I have som other related thoughts on this card, such as the icons being too "cluttered" and the add and info button having the same visual focus, while the "add" should be considered the "main" action.  Let's maybe include that as part of the same discussion
1. It look like the "read more" link here would expand the text, not open another browser window. maybe change the style and add the "open in browser" icon to make it clearer what happens on click?  I think a "real" button would make more sense than a link here, maybe
image   image
1. See comment on the `TabButton` component aside, but I think ALL buttons in this UI should have tooltips. Now its confusing that most do, but a few (like the info, add skybrowser, and add/select image buttons) do not
1. For the "Add image to current browser" button (maybe this is the txt it should have for the tooltip?), just a + icon would probably be enough. The current one with an image is a bit cluttered.  he same goes for the info icon - maybe we could just use an icon with an i, and not the circle?  Like the one we use for the inforboxes? 
image
1. The "Delete" button in setting does not seem to work.  I also miss the little "x" for closing a skybrowser, but perhaps that's just me. (We do expect our users to do this quite often though, since the target renderables stay visible when closing the skybrowser otherwise, so maybe the delete action should be more easily accessible?)
1. I find it a bit weird that the cards in the grid filter list are not interactive aside from the buttons.  I don't know if it's habit, but I expected some type of action on the entire card. Maybe showing an enlarged version of the image and the info on hover, if we do not want them to be clickable? 

I think that was all. Sorry for the wall of text 😇

src/panels/SkyBrowserPanel/WorldWideTelescope/WwtProvider/types.d.ts
engbergandreas: Agree, we have the `Vec3` type from OpenSpace API but I think we can create a new one in this repo instead

src/components/ResizeableContent/ResizeableContent.tsx
```diff
@@ -5,18 +5,23 @@ import {
...
   minHeight?: number;
   defaultHeight: number;
 }
 
-export function ResizeableContent({ minHeight = 40, defaultHeight, children }: Props) {
+export function ResizeableContent({
+  minHeight = 40,
+  defaultHeight,
+  children,
+  ...props
```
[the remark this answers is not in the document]
ylvaselling: Closing as resolved as we couldn't come up with a better solution

src/panels/SkyBrowserPanel/hooks.ts
ylvaselling: Yes - the active one is the one with the border. Selected / added are the ones in the list

src/panels/SkyBrowserPanel/Tabs/TabContent.tsx
```diff
@@ -0,0 +1,97 @@
...
+            <TrashIcon />
+          </TabButton>
+          <TabButton text={'Open Telescope View'} onClick={openWorldWideTelescope}>
+            <OpenInNewIcon />
+          </TabButton>
+        </ActionIcon.Group>
+        <TabButton
+          text={'Settings'}
+          onClick={() => setShowSettings((old) => !old)}
+          variant={showSettings ? 'filled' : 'default'}
```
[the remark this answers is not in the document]
ylvaselling: I agree, but I couldn't find one.. tried with "chip" and "radio" but neither looked particularly good.

src/redux/skybrowser/skybrowserSlice.ts
ylvaselling: They are currently separate arrays as that makes it easy to compare if there has been any changes in redux. If it is an object that is a bit trickier. I also plan to rewrite the topic so that all these comparisons aren't necessary

src/panels/SkyBrowserPanel/Tabs/BrowserTabs.tsx
```diff
@@ -0,0 +1,69 @@
...
+    return state.skybrowser.selectedBrowserId;
+  });
+  const luaApi = useOpenSpaceApi();
+
+  return (
+    <Tabs
+      variant={'outline'}
+      value={selectedBrowserId}
+      onChange={(id) => {
+        if (id) {
```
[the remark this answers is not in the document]
WeirdRubberDuck: Ah, right. Forgot that the Tabs have that feature.  👍

ylvaselling:
> And here are the content-related comments:
> 
>     1. The "world Wide Telescope" window should not be opened until a browser has been added, IMO
> 
>     2. The "Add browser" buttons needs some love/styling.  Maybe make it full-width / add some padding? Or at least center it and make it bigger?
> 
>     3. do we really want to keep using zero indexing for the skybrowsers?  At least in the UI, we should probably start with "Sky browser 1", not "Sky browser 0"
> 
>     4. We never want to simply rely on color for identifying something. this is currently a problem in the World Wide telescope Window. The name of the browser could maybe be added to the tab?  If I didn't see color, it would behard to know which sky browser was showing in the window here:
>        image
> 
>     5. I also found it very confusing that the same window was used to show all added browsers. What if I want to look at two parts of the sky at the same time??   We should open a new window for each browser, when clicking the "Open Telescope View", and maybe only ad the first one per default?  (or the subsequent ones as tabs in the same window or something). I do see how this may become complicated though, so it should be done with care
> 
>     6. We should add tooltips to all truncated texts, like these ones in the image list, that show the full name on hover.  Now it's hard to know how to find out what the name of the image actually is
>        image
> 
>     7. The image part of the items in the list in the image above is a bit small., especially compared to the parts with the buttons. can we maybe discuss and consider an alternative design, that makes the image more visibly dominant?
>        
>        * I have som other related thoughts on this card, such as the icons being too "cluttered" and the add and info button having the same visual focus, while the "add" should be considered the "main" action.  Let's maybe include that as part of the same discussion
> 
>     8. It look like the "read more" link here would expand the text, not open another browser window. maybe change the style and add the "open in browser" icon to make it clearer what happens on click?  I think a "real" button would make more sense than a link here, maybe
>        image   image
> 
>     9. See comment on the `TabButton` component aside, but I think ALL buttons in this UI should have tooltips. Now its confusing that most do, but a few (like the info, add skybrowser, and add/select image buttons) do not
> 
>     10. For the "Add image to current browser" button (maybe this is the txt it should have for the tooltip?), just a + icon would probably be enough. The current one with an image is a bit cluttered.  he same goes for the info icon - maybe we could just use an icon with an i, and not the circle?  Like the one we use for the inforboxes?
>         image
> 
>     11. The "Delete" button in setting does not seem to work.  I also miss the little "x" for closing a skybrowser, but perhaps that's just me. (We do expect our users to do this quite often though, since the target renderables stay visible when closing the skybrowser otherwise, so maybe the delete action should be more easily accessible?)
> 
>     12. Either way, Iäm not a fan of buttons that take more visual space than it needs to, like the delete button does now. Please make it so that it doesnät stretch across the entire width :) (It becomes waaay too visually dominant)
>         image
> 
>     13. I find it a bit weird that the cards in the grid filter list are not interactive aside from the buttons.  I don't know if it's habit, but I expected some type of action on the entire card. Maybe showing an enlarged version of the image and the info on hover, if we do not want them to be clickable?
> 
> 
> I think that was all. Sorry for the wall of text 😇

Making a list to keep track:

1. ✅ Fixed
2. ✅  Fixed
3. ✅  I agree but putting it in "future improvements" as it is a new feature
4. ❗ Needs discussion as how to implement this
5. ✅  I agree but putting it in "future improvements" as it is a new feature
6. 🔲  To be implemented
7. ❗ Needs discussion
8. 🔲 To be implemented
9. 🔲 To be implemented ❗ Discussion: do we want a delay for the tooltips to appear?
10. 🔲 To be implemented
11. ❗ Needs discussion. It's not possible to put them in the tab, hence why i put it where it is now. Where to put it instead? The bug is fixed.
12. ✅ Settings is not really a part of this PR anyway so I think I will leave this for later. 
13. ❗ Making the card clickable will have issues accessibility-wise, hence the current solution. It is possible to make the image clickable however. The images are actually in this weird narrow format so that comes from WWT. It is possible to "zoom" them but then we lose resolution.

WeirdRubberDuck (approved):

src/types/skybrowsertypes.d.ts
engbergandreas, line 1: these should be moved, but can be done with the other pr

engbergandreas (approved):

PR #39, opened 2025-02-03
Add loading state to focus and time button

WeirdRubberDuck:
Every time I open the UI I thought about that we should do this, so now I did it so I don't have to think about it anymore :) 

image

### DIFF
--- a/src/panels/OriginPanel/OriginPanelMenuButton.tsx
+++ b/src/panels/OriginPanel/OriginPanelMenuButton.tsx
@@ -1,4 +1,4 @@
-import { Button, Group, Stack, Text } from '@mantine/core';
+import { Button, Group, Skeleton, Stack, Text } from '@mantine/core';
 
 import { useGetPropertyOwner, useGetStringPropertyValue } from '@/api/hooks';
 import { AnchorIcon, FocusIcon, TelescopeIcon } from '@/icons/icons';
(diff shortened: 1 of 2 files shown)


engbergandreas:
Nice! Do we want to keep the disabled state on the button? If we do, all buttons should have it. I'm leaning towards letting the buttons be clickable and have the panels deal with the state (which most already do)

engbergandreas (approved):

WeirdRubberDuck:
> Nice! Do we want to keep the disabled state on the button? If we do, all buttons should have it. I'm leaning towards letting the buttons be clickable and have the panels deal with the state (which most already do)

In this case, I would personally prefer to have the buttons disabled. At least for these two buttons, since they are directly tied to an OpenSpace state. It simply does not make sense to be able to even open the panels, IMO, so intuitively it makes sense that they are disabled. to me at least

also:
1. the `Skeleton` components look atrocious on the active background (I guess this would not as much of a problem if we change the style of the buttons, but still), and I would not want to create a custom component to just make them look nice on a button that can't really be used anyways
2. then we do not have to deal with the "loading" state for these two panels, which I think makes sense since there wouldn't be any meaningful  information to show in these panels anyways

It looks a bit weird now since they "stick" out so much against the active buttons, but I think it will look better once we change the styling.

engbergandreas:
Sure make sense

> 2\. then we do not have to deal with the "loading" state for these two panels, which I think makes sense since there wouldn't be any meaningful  information to show in these panels anyways

We might still end up with a state where we will need some form of loading or handling invalid data though so we can't remove it altogether

ylvaselling:
Looks great

ylvaselling (approved):

PR #40, opened 2025-02-03
Feature/rc dock fork

engbergandreas:
This PR introduces the rc-dock fork from here: https://github.com/OpenSpace/OpenSpace-WebGui-WindowLayout
I think I've fixed the most crucial issues we had, especially with panels resizing when moved. However, there are probably cases that I did not find or in which a fix will cause bigger problems in another situation, had a few whack-a-mole moments. 

It should be enough to run `npm install` so that `rc-dock` points to the new repo instead of the one in `node_modules` 🤞 

Some fixes: 
Float window sizing 
Docking a float window uses a new default size
Docking a tab to another panel uses a new default size or the existing size
Removed the "deeper" docking so now it is only left, right, top, and bottom and not the ones where you dock to the edge
Removed flex-shrink on normal windows now we have to specify groups of windows that should shrink/grow

TODO: Go back to the floating window sizes when we've fixed them style wise


src/windowmanagement/data/MenuItems.tsx
```diff
@@ -42,6 +43,7 @@ export interface MenuItem {
   renderMenuButton?: (key: string, onClick: () => void) => React.JSX.Element; // Custom menu button to render
   renderIcon?: (size: IconSize) => React.JSX.Element; // Custom icon to render
   preferredPosition: WindowLayoutPosition; // Where this panel is instantiated
+  floatPosition?: FloatWindowPosition; // Preferred position and size of a floating window given in px
```
engbergandreas, line 46:
```suggestion
  floatPosition?: FloatWindowPosition; // Preferred position and size of a floating window given in px offset from lower right corner
```

ylvaselling (approved):
Good stuff

src/windowmanagement/WindowLayout/WindowLayoutProvider.tsx
```diff
@@ -67,18 +80,22 @@ export function WindowLayoutProvider({ children }: { children: React.ReactNode }
           else {
             tab.group = 'regularWindow';
             const panel: PanelData = {
-              tabs: [tab]
+              tabs: [tab],
+              size: 300
```
WeirdRubberDuck, line 84: Make a named variable for this? `SidePanelDefaultWidth` or something.  Could be global in this file?

src/windowmanagement/data/MenuItems.tsx
```diff
@@ -85,6 +88,7 @@ export const menuItemsData: MenuItem[] = [
     ),
     renderIcon: (size) => <CalendarIcon size={size} />,
     preferredPosition: 'float',
+    floatPosition: { offsetY: 100, offsetX: 370, width: 580, height: 520 },
```
WeirdRubberDuck, line 91: Merge in master (to get the polished time input) and make the default size of this panel a bit smaller, please :) Now it seems huge in comparison to the others

WeirdRubberDuck (approved):

PR #41, opened 2025-02-05
Refactor filterlist to make it easier to make more types

ylvaselling:
This PR adds a new component `SearchResults` for filterlist, with subtypes for the different types of results you might want to display. In this PR these are `VirtualList` and `Pagination`. Skybrowser will add `VirtualGrid`. I can also see we might want to add a regular list, without virtualization, and also potentially a grid without virtualization (thinking of the actions panel).

Some reasons for this refactor: 
1. Started debugging weird crash in the skybrowser PR which had to do with the FilterList. It was a bit hard to debug since the complexity of the different search result components have grown
2. The SkyBrowser is also adding another type of component for the search results and this refactor will make it easier to add more types

Has been done in discussion with @engbergandreas


src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,32 @@
+import { VirtualList } from '../../VirtualList/VirtualList';
```
engbergandreas, line 1: use @ import here

engbergandreas (approved):

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,32 @@
...
+  renderElement,
+  matcherFunc,
+  gap,
+  overscan
+}: VirtualizedProps<T>) {
+  const filteredElements = useSearch(matcherFunc, data);
+
+  if (filteredElements.length === 0) {
+    return <>No matches found. Try another search</>;
+  }
```
WeirdRubberDuck, line 22: Can this check be moved to the `SearchResults` component?  It should be done for the Paginated list as well (and the grid, once that exist)

src/components/FilterList/SearchResults/SearchResultsPagination.tsx
```diff
@@ -0,0 +1,29 @@
+import { PaginationList } from '@/components/PaginationList/PaginationList';
+
+import { useSearch } from './hooks';
+import { SearchResultsProps } from '@/types/types';
```
WeirdRubberDuck, line 4: These imports need to be sorted

```suggestion
import { PaginationList } from '@/components/PaginationList/PaginationList';
import { SearchResultsProps } from '@/types/types';

import { useSearch } from './hooks';
```

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,32 @@
+import { SearchResultsProps } from '@/types/types';
```
WeirdRubberDuck, line 1: Sort props to match linting

Also, I personally don't think that this Props component should live in the general types.d.ts file.  That file should include types that are relevant to all parts of the UI, and this type is only relevant to the `SearchResults` component. 

Create a local types folder? Or export it from the util file?  Thoughts @ylvaselling  and @engbergandreas =

src/components/FilterList/SearchResults/SearchResults.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { useFilterListProvider } from '../hooks';
+
+import { SearchResultsPagination } from './SearchResultsPagination';
+import { SearchResultsVirtualList } from './SearchResultsVirtualList';
+
+export function SearchResults({ children }: PropsWithChildren) {
+  const { showFavorites, isLoading } = useFilterListProvider();
+  if (!children) {
+    throw Error('FilterList.SearchResults should contain children!');
+  }
```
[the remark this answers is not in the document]
WeirdRubberDuck: Hm, after looking at the code a bit more I understand why this check might be useful - it acts as a guide to the person coding the component. However, I think some more guidance should be provided. For example: 

```suggestion
  if (!children) {
    throw Error('FilterList.SearchResults must contain children! Please provide a search result type component as the  child of this component);
  }
```
?

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
WeirdRubberDuck: Or should this be an optional component that can be added as a child to SearchResults? :)  Would match the "composite component" design style quite well, maybe?

src/components/FilterList/SearchResults/hooks.ts
```diff
@@ -0,0 +1,18 @@
+import { useMemo } from 'react';
+
+import { useFilterListProvider } from '../hooks';
+
+export function useSearch<T>(
```
WeirdRubberDuck, line 5:
```suggestion
export function useSearchResults<T>(
```
?  Makes it clearer what the hook does

src/components/FilterList/SearchResults/hooks.ts
```diff
@@ -0,0 +1,18 @@
...
+) {
+  const { searchString } = useFilterListProvider();
+
+  // Memoizing this function so we don't need to recreate it when
+  // the renderElement function changes
+  const filteredElements = useMemo(
+    () => data.filter((e) => matcherFunc(e, searchString)),
+    [searchString, matcherFunc, data]
+  );
+  return filteredElements;
```
WeirdRubberDuck, line 17:
```suggestion
  return useMemo(
    () => data.filter((e) => matcherFunc(e, searchString)),
    [searchString, matcherFunc, data]
  );
```

if the hook is renamed to `useSearchResults`, this variable is no longer needed for readability, since it becomes apparent that the this it returns is the results of the search :)

src/components/VirtualList/VirtualList.tsx
```diff
@@ -37,6 +38,7 @@ export function VirtualList<T>({
   return (
     <>
       {/* The scrollable element for your list */}
+
```
WeirdRubberDuck, line 41:
```suggestion
```

src/components/VirtualList/VirtualList.tsx
```diff
@@ -48,27 +50,33 @@ export function VirtualList<T>({
...
+            style={{
+              position: 'absolute',
+              top: 0,
+              left: 0,
+              width: '100%',
+              transform: `translateY(${items[0]?.start ?? 0}px)`
+            }}
+          >
+            {/* Only the visible items in the virtualizer,
           manually positioned to be in view */}
```
WeirdRubberDuck, line 68:
```suggestion
            {/* The visible items, manually positioned to be in view */}
```

WeirdRubberDuck (approved):

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
ylvaselling: My thoughts was also that it would make sense to have a types file in this folder but @engbergandreas asked me to put it in the global types file... I think there are pros and cons with both

src/components/FilterList/SearchResults/hooks.ts
ylvaselling: Imo it's more readable to keep the variable 🤔

WeirdRubberDuck: If you think it adds something, keep it.  I personally just think it's redundant if the hook is renamed :)

WeirdRubberDuck (approved):

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,21 @@
+import { VirtualList } from '../../VirtualList/VirtualList';
+
+import { useSearchResultProvider } from './hooks';
+
+interface VirtualizedProps {
```
engbergandreas, line 5:
```suggestion
interface Props {
```

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,21 @@
...
+  gap?: number; // Gap in pixels between items
+  overscan?: number; // How many items to preload when scrolling
+}
+
+export function SearchResultsVirtualList<T>({ gap, overscan }: VirtualizedProps) {
```
engbergandreas, line 10:
```suggestion
export function SearchResultsVirtualList<T>({ gap, overscan }: Props) {
```

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -0,0 +1,21 @@
+import { VirtualList } from '../../VirtualList/VirtualList';
```
engbergandreas, line 1: use @ import?

engbergandreas (approved):

PR #43, opened 2025-02-06
Feature/polish time panel

engbergandreas:
Fixed the major issue we had with the date input fields. Further code cleanup can happen when Emma implements her version of the input field. Then we can remove the prop `wrapStepControlButtons` and have the buttons default wrap the input field.

### DIFF
--- a/src/panels/TimePanel/StepControls.tsx
+++ b/src/panels/TimePanel/StepControls.tsx
@@ -8,12 +8,15 @@ import { ChevronDownIcon, ChevronUpIcon } from '@/icons/icons';
  * https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/NumberInput/NumberInput.tsx#L406
  */
 
+type StepDirection = 'up' | 'down';
+
 interface Props extends PropsWithChildren {
   stepHoldDelay?: number;
   stepHoldInterval?: number;
   onChange: (change: number, shiftKey: boolean) => void;
   step?: number;
 }
+
 export function StepControlButtons({
   stepHoldDelay,
   stepHoldInterval,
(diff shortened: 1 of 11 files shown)


ylvaselling (commented):
Some thoughts, maybe not related to this PR but to the time panel in general:

1. Mitigating the flickering behavior of the input (the caching problem). We have solved this in other components with having a local state that is changed with a useEffect that has the redux state in its dependency array
2. The buttons in the Seconds / second input field do not fit in the input field
3. Should the "cancel" button in the "lock" panel be gray maybe? Or somehow different from the other two buttons?
4. The day of the date is always 31 regardless of month which is a bit confusing.
5. Do we want to signal that values are not accepted more clearly? Now it's just not possible to enter the input but we can also mark the text in red, for example

src/panels/TimePanel/StepControls.tsx
```diff
@@ -27,19 +28,19 @@ export function HoldableButton({
   const shouldUseStepInterval =
     stepHoldDelay !== undefined && stepHoldInterval !== undefined;
 
-  function downHandler(event: KeyboardEvent) {
+  function downHandler(event: KeyboardEvent): void {
```
WeirdRubberDuck, line 31:
```suggestion
  function keyDownHandler(event: KeyboardEvent): void {
```
otherwise, it might just as well be a mousedown handler

src/panels/TimePanel/StepControls.tsx
```diff
@@ -27,19 +28,19 @@ export function HoldableButton({
...
     if (event.key === 'Shift') {
       shiftKeyRef.current = true;
     }
   }
 
-  function upHandler(event: KeyboardEvent) {
+  function upHandler(event: KeyboardEvent): void {
```
WeirdRubberDuck, line 37:
```suggestion
  function keyUpHandler(event: KeyboardEvent): void {
```

WeirdRubberDuck (approved):
Had some small comments but ended up making those changes directly instead. So 👍 from me

WeirdRubberDuck:
But yes we do need to add to a TODO to have a second look at the time input.  It have to be adapted to the OpenSpace use case with weird years, and to handle the number of days in a month a bit better. 

For example, if I set the date to Ovct 31, and then click the ⬇️ to decrease the month, I end up at Oct 1.  I would have expected the date to be the last of September

src/panels/TimePanel/TimeIncrementInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+              onChange={(value, shiftKey) => onInputChange(value, true, !shiftKey)}
+              step={step}
+            />
+          }
+        />
+      );
+    }
+  }
+
+  return displayInputField();
```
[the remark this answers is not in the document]
WeirdRubberDuck: good catch!  Did this now, since I was already making changes

WeirdRubberDuck:
> Some thoughts, maybe not related to this PR but to the time panel in general:
> 
> 1. Mitigating the flickering behavior of the input (the caching problem). We have solved this in other components with having a local state that is changed with a useEffect that has the redux state in its dependency array
> 2. The buttons in the Seconds / second input field do not fit in the input field
> 3. Should the "cancel" button in the "lock" panel be gray maybe? Or somehow different from the other two buttons?
> 4. The day of the date is always 31 regardless of month which is a bit confusing.
> 5. Do we want to signal that values are not accepted more clearly? Now it's just not possible to enter the input but we can also mark the text in red, for example

I'll take the liberty of answering these ones: 

1. Yes, this should be looked at - I can do it
2. ✅ This will be fixed in my new version of the component
3. ✅ Done. Made it light
4. I also find this a bit confusing.  
5. Yes! The numeric input has support for this and I eventually plan to add it for the properties as well.  It might be good to wait with this here until our custom input component is finished, however. Although, I think we will need to make som special solution for it anyways in this case since the space beneath the input field is too tight for a text message

PR #44, opened 2025-02-07
adjust scene tree filterlist to take all available space

engbergandreas:
Since the tabs are still there it will be off by 36 px (height of tabs) so the window component will be scrollable, but this wont be an issue when they are gone (Emmas branch) 

Also, I thought about it, but I don't know if this function needs to be memo'd? I did not want to memo if it is not necessary...


src/panels/Scene/SceneTree/SceneTree.tsx
```diff
@@ -103,8 +107,21 @@ export function SceneTree() {
     return nameA.toLocaleLowerCase().localeCompare(nameB.toLocaleLowerCase());
   });
 
+  function computeHeight(windowHeight: number): number {
```
WeirdRubberDuck, line 110: This seems like a completely copied version of the function in OriginPanel? Can we generalize it and put it in some utility function instead?

src/panels/Scene/SceneTree/SceneTree.tsx
```diff
@@ -27,7 +27,11 @@ import {
   sortTreeData
 } from './treeUtil';
 
-export function SceneTree() {
+interface Props {
+  siblingRef: React.RefObject<HTMLDivElement>;
```
WeirdRubberDuck, line 31: Is this the best name for this variable?  Maybe give it a comment that explains why it is the way it is?

WeirdRubberDuck: ALternatively, move the computeHeight function to the Scene.tsx file and pass the function instead of the ref?  Avoids having to look in another file to see what this ref actually is

WeirdRubberDuck (commented):
Works well, but see comments in the code. Some might need discussion

src/components/FilterList/hooks.ts
```diff
@@ -9,3 +9,21 @@ export function useFilterListProvider() {
   }
   return context;
 }
+
+export function useGenerateHeightFunction(minHeight: number, bottomMargin?: number) {
```
WeirdRubberDuck, line 13: Also please document this hook! :)

src/components/FilterList/hooks.ts
```diff
@@ -9,3 +9,33 @@ export function useFilterListProvider() {
   }
   return context;
 }
+
+/**
+ * Generates a function that calculates the available height for the `FilterList` component
+ *
+ * @param minHeight The minimum height the `FilterList` can be
+ * @param bottomMargin The amount of margin to the bottom of the panel, this value is
```
WeirdRubberDuck, line 17:
```suggestion
 * @param bottomMargin The margin to the bottom of the panel. This value is
```
But blease also clarify what "The margin to the bottom of the panel" means. The margin from what?? :)

PR #45, opened 2025-02-07
Feature/button styling

WeirdRubberDuck:
Make buttons and icons the `default` (gray) variant per default. 

Update some places where the styling looked weird after this change. 

Please go through and see if you think the button style should be changed or set anywhere

### DIFF
--- a/src/panels/OriginPanel/OriginPanel.tsx
+++ b/src/panels/OriginPanel/OriginPanel.tsx
@@ -196,7 +196,7 @@ export function OriginPanel() {
           <SegmentedControl
             value={navigationAction}
             withItemsBorders={false}
-            onChange={(v) => setNavigationAction(v as NavigationActionState)}
+            onChange={(value) => setNavigationAction(value as NavigationActionState)}
             data={[
               {
                 value: NavigationActionState.Focus,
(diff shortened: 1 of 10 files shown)


WeirdRubberDuck:
Might still want to make a semitransparent variant for the bottom bar

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -192,32 +193,40 @@ export function OriginPanel() {
...
-              size={'lg'}
-              style={actionIconStyle(NavigationActionState.Aim)}
-            >
-              <TelescopeIcon size={'70%'} />
-            </ActionIcon>
-          </Group>
+          <SegmentedControl
+            value={navigationAction}
+            withItemsBorders={false}
+            onChange={(v) => setNavigationAction(v as NavigationActionState)}
```
engbergandreas, line 199: I think we should avoid single character variable names unless they are for indices

WeirdRubberDuck: So: 
```suggestion
            onChange={(state) => setNavigationAction(state as NavigationActionState)}
```

ylvaselling (approved):

PR #46, opened 2025-02-07
Fix text wrapping

engbergandreas:
Fix text wrapping when searching for a place, change `Added Nodes` to use the SceneGraph header, this also fixes text wrapping in the exoplanets panel. 

Due to problems (probably with the ScrollArea) this does not fix text-wrapping in the scene tree.

### DIFF
--- a/src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/AddedCustomNodes.tsx
+++ b/src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/AddedCustomNodes.tsx
@@ -1,0 +1,47 @@
+import { ActionIcon, Box, Text } from '@mantine/core';
+
+import { SceneGraphNodeHeader } from '@/panels/Scene/SceneGraphNode/SceneGraphNodeHeader';
+import { Identifier } from '@/types/types';
+import { sgnUri } from '@/util/propertyTreeHelpers';
+import { useWindowSize } from '@/windowmanagement/Window/hooks';
+import { MinusIcon } from '@/icons/icons';
+
+interface Props {
+  addedNodes: Identifier[];
+  removeFocusNode: (identifier: Identifier) => void;
+}
+
+// This component is its own to reduce the amount of rerenders in
+// EarthPanel due to the `useWindowSize()` hook
+export function AddedCustomNodes({ addedNodes, removeFocusNode }: Props) {
+  // TODO anden88 2025-02-07 Using the windowSize like this causes some stuttering when
+  // moving the window panel. Investigation with Ylva we found that the culprit is most
+  // likely the `ScrollArea` wrapping the entire window, possible solution is make own
+  // scroll behaviour and instead wrap in e.g, `Box`
+  const { width } = useWindowSize();
+  const containerPadding = 36;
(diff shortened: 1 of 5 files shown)


src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/AddedCustomNodes.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { useWindowSize } from '@/windowmanagement/Window/hooks';
+import { MinusIcon } from '@/icons/icons';
+
+interface Props {
+  addedNodes: Identifier[];
+  removeFocusNode: (identifier: Identifier) => void;
+}
+
+// This component is its own to reduce the amount of rerenders in
+// EarthPanel due to the `useWindowSize()` hook
```
WeirdRubberDuck, line 15:
```suggestion
```
This comment is more confusing than helpful IMO :)

src/panels/Scene/SceneGraphNode/SceneGraphNodeHeader.tsx
```diff
@@ -56,7 +57,8 @@ export function SceneGraphNodeHeader({ uri, label, onClick }: Props) {
     <ThreePartHeader
       title={onClick ? titleButton : name}
       leftSection={
-        hasRenderable && <PropertyOwnerVisibilityCheckbox uri={renderableUri} />
+        leftSection ??
+        (hasRenderable && <PropertyOwnerVisibilityCheckbox uri={renderableUri} />)
```
WeirdRubberDuck, line 61: Hm, this would replace the checkbox, if a left section is specified.  This should be made clearer, e.g. using a comment above line 18

WeirdRubberDuck: Another option would be to include both the checkbox in a `Group`, but either way, the intention of `leftSection` needs to be communicated here

WeirdRubberDuck (approved):

src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/AddedCustomNodes.tsx
```diff
@@ -0,0 +1,47 @@
...
+export function AddedCustomNodes({ addedNodes, removeFocusNode }: Props) {
+  // TODO anden88 2025-02-07 Using the windowSize like this causes some stuttering when
+  // moving the window panel. Investigation with Ylva we found that the culprit is most
+  // likely the `ScrollArea` wrapping the entire window, possible solution is make own
+  // scroll behaviour and instead wrap in e.g, `Box`
+  const { width } = useWindowSize();
+  const containerPadding = 36;
+  const containerMaxWidth = 925;
```
ylvaselling, line 23: Some comments regarding these would be good

ylvaselling (approved):
Imo we should prioritize the proper way to fix this, which would be to investigate why ScrollArea doesn't work and either see if we can fix it or come up with our own ScrollComponent that works. This problem is going to come up in many panels so I think it's worth spending the time to make it work properly

PR #48, opened 2025-02-11
Refactor/types cleanup

engbergandreas:
Create types file locally in each panel instead

If a type is required in both a panel and redux we prefer to put it the panel folder


src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/types.ts
```diff
@@ -0,0 +1,28 @@
+// Incomplete type objects for the ArcGIS object we request for GeoLocationPanel
+// This was to avoid having to import the entire ArcGIS core since the @types file is
+// deprecated on npm.
```
WeirdRubberDuck, line 3: Add `anden` at hte end of this comment?  (Or, the id of whoever wrote it, so one knows who to ask about it?)

src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/types.ts
```diff
@@ -0,0 +1,28 @@
...
+export interface Location {
+  x: number;
+  y: number;
+}
+
+export interface Extent {
+  xmax: number;
+  xmin: number;
+  ymax: number;
+  ymin: number;
```
WeirdRubberDuck, line 27:
```suggestion
  xMax: number;
  xMin: number;
  yMax: number;
  yMin: number;
```
?

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -107,6 +107,7 @@ export function OriginPanel() {
     return aim !== '' && aim !== anchor;
   }
 
+  // TODO: anden88: @emma take a look at this
```
WeirdRubberDuck, line 110:
```suggestion
  // TODO: anden88: @emma take a look at this. Is the flight controller really needed for this?
```

src/redux/events/types.ts
```diff
@@ -1,99 +1,121 @@
-// Event types, these were manually typed from the event.cpp `toParameter`function:
-
-import { Identifier } from './types';
+import { Identifier, Uri } from '@/types/types';
 
+// Event types, these were manually typed from the event.cpp `toParameter`function:
 // https://github.com/OpenSpace/OpenSpace/blob/6dd3dd8ab88653fddee0b2e480934faf53d3c0f8/src/events/event.cpp#L332
-export type ParallelConnectionEvent = {
+type ParallelConnectionEvent = {
```
WeirdRubberDuck, line 5: Would it be too much to ask to have these event types sorted alphabetically..? :) 👼  And then also sort the `EventData` list at the end alphabetically as well

src/types/types.ts
```diff
@@ -1,9 +1,12 @@
 import { Property } from '@/components/Property/Property';
+
 import { PropertyVisibilityNumber } from './enums';
 
+/*--------------------------------------------------------------------------------------*/
```
WeirdRubberDuck, line 5: Hm, not a fan of these dividers, and don't fully agree with their placement either.  

If we have dividers, it should be made clear what type of types belong in each section

Personally, I would prefer to just remove them, I think

src/types/types.ts
```diff
@@ -119,31 +92,8 @@ export type Groups = {
   [key: string]: Group;
 };
 
+/*--------------------------------------------------------------------------------------*/
```
WeirdRubberDuck, line 95:
```suggestion
```

The CustomGroupOrdering is related to the "Groups"

src/redux/events/types.ts
ylvaselling: Linting rules for type files? But i guess that would apply to utils files and hooks files as well 🤔

ylvaselling (approved):

src/redux/events/types.ts
engbergandreas: not with chatgpt to the rescue it aint 😄

PR #50, opened 2025-02-14
Style updates

WeirdRubberDuck:
Various style-related changes that improved the visual look of the UI

- Use a non-transparent Mantine color for the window backgrounds
- Decrease font size for headings using Mantine theme setting
- Wrap text in SGN headers in the Scene Tree more nicely (using the CSS `pretty` textWrap setting)
- Fix button styling in top menu bar (and refactor the code for it a bit*)
- Update look of the rc-dock tabs and add close button to the windows upper right corner
- Add a custom button variant for the taskbar and top menu bar buttons (transparent with bright hover hint)
- Move Mantine theme settings to a separate file. The theme is quickly growing

*Note that it also restructures the top menubar code so that each menu is its own component and that the style only has to be set in one place (in the


src/panels/Menu/TopMenuBar/TopBarMenuWrapper.tsx
```diff
@@ -1,24 +1,28 @@
...
+  Text
 } from '@mantine/core';
 
 interface Props extends MenuProps, PropsWithChildren {
+  targetTitle: string | React.ReactNode;
   menuItemTabIndex?: -1 | 0;
   position?: FloatingPosition;
   offset?: number | FloatingAxesOffsets;
   withArrow?: boolean;
   arrowPosition?: ArrowPosition;
```
ylvaselling, line 18: No need to define these here, they are already in MenuProps

src/windowmanagement/WindowLayout/WindowLayout.tsx
```diff
@@ -55,19 +55,29 @@ export function WindowLayout() {
...
+    // This is the rc-dock group configuration we use for the transparent window in the
+    // center
+    headless: {
+      floatable: false,
+      maximizable: false,
+      tabLocked: true,
+      widthFlex: 1
+    },
+    // This is the rc-dock group configuration for all the other windows. Note that
+    // 'card' is a pre-existing rc-dock group that adds certina styling to the tab
```
ylvaselling, line 68:
```suggestion
    // 'card' is a pre-existing rc-dock group that adds certain styling to the tab
```

src/windowmanagement/WindowLayout/WindowLayout.tsx
```diff
@@ -55,19 +55,29 @@ export function WindowLayout() {
...
+    card: {
```
ylvaselling, line 69: I'm wondering if we should some how have a group for our regular windows as well. It seems a bit hacky to me to just modify the `card` group...

Like `"card regularWindow":`  ?

WeirdRubberDuck: I initially had the same thought, but became a bit unsure since "headless" (i.e. the "non-regular" window) is also a predefined rc-dock group. So then, maybe we should do something similar for that one as well. 

I ended up just using "card" since I thought that that was easier to understand than the double class names. But let's discuss what to do here :)

src/panels/Menu/TopMenuBar/TopBarMenuWrapper.tsx
WeirdRubberDuck: Good catch

src/panels/Menu/TopMenuBar/Menus/FileMenu.tsx
```diff
@@ -0,0 +1,48 @@
...
+import { useGetBoolPropertyValue, useOpenSpaceApi } from '@/api/hooks';
+import { AddFileIcon, ConsoleIcon, ExitAppIcon } from '@/icons/icons';
+
+import { TopBarMenuWrapper } from '../TopBarMenuWrapper';
+
+export function FileMenu() {
+  const luaApi = useOpenSpaceApi();
+
+  const [isConsoleVisible, setIsConsoleVisible] =
+    useGetBoolPropertyValue('LuaConsole.IsVisible');
```
engbergandreas, line 13: Do we want to use a key here instead?

src/panels/Menu/TopMenuBar/Menus/FileMenu.tsx
```diff
@@ -0,0 +1,48 @@
...
+  function toggleLuaConsole() {
+    setIsConsoleVisible(!isConsoleVisible);
+  }
+
+  function toggleShutdown() {
+    return modals.openConfirmModal({
+      title: 'Confirm action',
+      children: <Text>Are you sure you want to quit OpenSpace?</Text>,
+      labels: { confirm: 'Quit', cancel: 'Cancel' },
+      confirmProps: { color: 'red' },
```
engbergandreas, line 24: This button is no longer red (due to default button style(?)) add variant="filled"?

engbergandreas (approved):

src/panels/Menu/TopMenuBar/Menus/FileMenu.tsx
WeirdRubberDuck: I think not. There are multiple places where we are still using just the URI for the properties (e.g. in helpmenuHelpMenu, WwtProvider, SceneGraphNodeMetaInfo), so I think we can just leave it here for now

It's a future discussion to have, though. I personally find it easier to identify which property is begin  accessed when the URI is used rather than a key name

PR #53, opened 2025-02-18
Feature/scene tree properties

WeirdRubberDuck:
Adds implementation of the different property type views, as well as some styling for the menus with property owners.

**Needs testing!** 
There are definitely things that I have missed, so making this a draft PR. A test page can be accessed under "Windows", but note that this one does not cover all variants of the properties. 

Note that the built version runs much smoother than the dev one.


ylvaselling:
Overall I think it looks awesome!! ✨ I just jotted down some notes from looking at the properties. Lmk if you want to discuss 

1. General 
    1. Maybe we can make the titles less bold? They are a little hard to read now imo.

    1. I just had a thought - we could potientially go around the min max problem by making them as tiny input fields so you could set them yourself.

1. Numeric property 

    1. Overlap of text over the textbox 
image

    1. Is there a way to make these look more connected? When the window is small it is hard to see that these two are the same property. 

    1. Why does the slider have dots for these values? Seems 0.05 is always in there?

    1. Should we somehow let the user know when they are out of range? Like we did in the old UI 

    1. Seems the circle slider doesn't go all the way to the right? I think there might be an offset to the blue bar? image
image

    1.  It seems when the property is line wrapped it goes outside bounds? image

2. String list property
Can we somehow make this smaller? Like a dropdown?
image

8. Vec2 property
In the gui it took me a while to realize this was probably a vec2 property. I assume this is x and y? Can we maybe write that out somehow if that is the case? This might go for all vec properties - we might want to write out xyzw?
image

9. Selection property
Make a placeholder when empty?

4. Min max
Should we place the slider in the middle of the the number? Or the slider above the numbers, to mimic the numeric property when it is linewrapped?

3. Infobox
Can we place it so it is not in different places depending on how long the title is? Maybe to the far right?

5. Propertyowner
    1. Some padding on the right hand side? image
    2. Align about with the upper border of the about text?
image

7. Seems we have an issue with rendering
image

6. Wth 😆 We really like saying Earth is a special place
image

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,169 @@
...
+  const [shownValues, setShownValues] = useState(value);
+
+  const [isEditing, setIsEditing] = useState(false);
+  const [clickedItemIndex, setClickedItemIndex] = useState<number | undefined>(undefined);
+
+  const inputFieldRef = useRef<HTMLInputElement>(null);
+
+  useEffect(() => {
+    setShownValues(value);
+  }, [value]);
```
ylvaselling, line 43: Use the hook you made for this?

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -0,0 +1,84 @@
...
+  const exponent = additionalData.Exponent;
+
+  const isInt = step >= 1;
+
+  // When no min/max is set, the marks for the slider cannot be nicely computed
+  const hasNiceMinMax = isFinite(max - min);
+
+  // @TODO There still seems to be a bit of a stutter when dragging the slider.
+  // nvestigate. Is there somethign we could memo to make it better?
+  function onValueChange(newValue: number) {
```
ylvaselling, line 47: onValueEnd? or something (update the value when releasing the slider)

src/components/Property/Types/NumericProperty/NumericPropertySlider.tsx
```diff
@@ -0,0 +1,115 @@
...
+    // Only update the current value if the value is not currently being edited
+    if (!isEditingSlider) {
+      setCurrentValue(value);
+    }
+  }, [isEditingSlider, value]);
+
+  const scale = scalePow().exponent(exponent).domain([min, max]).range([min, max]);
+  const decimalPlaces = Math.max(0, -Math.floor(Math.log10(step)));
+
+  function markLabel(value: number, largeValuePrecision?: number): JSX.Element | string {
```
ylvaselling, line 39: Make this a component?

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -20,11 +23,24 @@ export interface VectorPropertyProps {
...
+    view = <ValueList {...props} />;
   }
 
-  return <ValueList {...props} />;
+  return (
+    <>
+      <PropertyLabel label={name} tip={description} />
+      {view}
+    </>
+  );
```
ylvaselling, line 45: Make this a switch that uses viewOptions and make each case return?

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -20,11 +23,24 @@ export interface VectorPropertyProps {
 }
 
 export function VectorProperty(props: VectorPropertyProps) {
-  const { value, viewOptions } = props;
+  const { value, viewOptions, name, description } = props;
+
+  const isColor = (value.length === 3 || value.length === 4) && viewOptions.Color;
+  const isMinMaxRange = value.length === 2 && viewOptions.MinMaxRange;
```
ylvaselling, line 29: Why do we need to check for this here? Can't we use only viewOption as the determining factor and throw an error in the component (ColorVector, MinMaxRange) if it is wrong? As it is now we allow the property to be set wrongly in openspace

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
```diff
@@ -0,0 +1,38 @@
+import { Flex, RGBA } from '@mantine/core';
+
+import { ColorPicker } from '@/components/ColorPicker/ColorPicker';
+import { VectorPropertyProps } from '@/components/Property/Types/VectorProperty/VectorProperty';
+
+import { ValueList } from './DefaultValueList';
+
+export function ColorVector(props: VectorPropertyProps) {
+  const { disabled, setPropertyValue, value } = props;
```
ylvaselling, line 9:
```suggestion
export function ColorVector({ disabled, setPropertyValue, value }: VectorPropertyProps) {

```

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
```diff
@@ -0,0 +1,38 @@
...
+  const hasAlpha = value.length === 4;
+
+  function setValue(newValue: number[]) {
+    setPropertyValue(newValue);
+  }
```
ylvaselling, line 14: I don't think we need this function

src/components/Property/Types/VectorProperty/ViewOptions/DefaultValueList.tsx
```diff
@@ -16,28 +13,29 @@ export function ValueList({
   const max = additionalData.MaximumValue;
   const step = additionalData.SteppingValue;
 
+  const isInt = step.every((s) => s >= 1);
```
ylvaselling, line 16: ?

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -0,0 +1,84 @@
...
+  setPropertyValue: (newValue: number) => void;
+  value: number;
+  additionalData: {
+    Exponent: number;
+    MaximumValue: number;
+    MinimumValue: number;
+    SteppingValue: number;
+  };
+  // TODO: view options in metadata
+}
```
ylvaselling, line 22: Can we extend Props from a Property type?

src/components/Property/Types/VectorProperty/VectorProperty.tsx
WeirdRubberDuck: This is simply copy-pasted from how it was implemented in the old UI, where it was simply a safeguard against if someone sets the wrong viewoption in C++ (which in general, we shouldn't), and cummincate in some way that this is the use case when each view option is supported on one line, I guess 😅  

But I agree, now that there are separate components for each view option, it would be nicer to check it in each component and throw of the number of values is wrong.  We should get away from the old tendency of silently ignoring errors

src/components/PropertyOwner/PropertyOwner.tsx
```diff
@@ -42,7 +42,9 @@ export function PropertyOwner({ uri, expandedOnDefault = false }: Props) {
       defaultOpen={expandedOnDefault}
       noTransition
     >
-      <PropertyOwnerContent uri={uri} />
+      <Box style={{ borderLeft: '2px solid var(--mantine-color-gray-8)' }}>
```
WeirdRubberDuck, line 45: Investigate using mantine's Divider for this instead

src/components/ColorPicker/ColorEdit.tsx
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  const warnAboutInvalidColor = !isColorValid(textEditValue);
+
+  return (
+    <Stack gap={'xs'}>
+      <MantineColorPicker
+        value={value}
+        // We don't use the selected format here, since changin the format would then
```
WeirdRubberDuck, line 53:
```suggestion
        // We don't use the selected format here, since changing the format would then
```

src/components/Property/Types/NumericProperty/NumericProperty.tsx
WeirdRubberDuck: Unfortunately, I donät think that is an option.  The user probably want to see the value bein updated while the slider is being used

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,169 @@
...
+  valueType: 'string' | 'int' | 'float';
+}
+
+export function ListProperty({
+  name,
+  description,
+  disabled,
+  setPropertyValue,
+  value,
+  valueType = 'string'
```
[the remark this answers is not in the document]
WeirdRubberDuck: It isn't. The default value was likely just left as a remnant from some old version.. :)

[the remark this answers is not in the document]
WeirdRubberDuck: Actually, nvm.  It needs to be there. 

The int number and float number versions of the list properties is just a specialization of the regular string-based one (i.e. this one).

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,169 @@
...
+          return `integer${count1}, integer${count2}, ...`;
+        case 'string':
+          return `item${count1}, item${count2}, ...`;
+        default:
+          throw new Error('Invalid value type');
+      }
+    },
+    [valueType]
+  );
+
```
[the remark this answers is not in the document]
WeirdRubberDuck: Hehe, yeah, maybe this placeholder is a bit overdesigned... But I did find it a useful way to communicate what type of value the list input expects. There is currently on way for the user to know this, otherwise. 

Also since the inputs also do the work of just showing what the value is, I only added the placeholder on edit. Otherwise I think it would be unnecessarily cluttered

But I'll remove the count, seems redundant

src/components/Property/Types/VectorProperty/VectorProperty.tsx
WeirdRubberDuck: Can't make a switch, but I'll make if statements for each view option, that directly returns

src/components/Property/Types/VectorProperty/ViewOptions/DefaultValueList.tsx
WeirdRubberDuck: 😅 A bit of a hacky way to check if the property type is an int... I'll make a new type instead, similar to the ListProperty

src/components/Property/Types/NumericProperty/NumericProperty.tsx
WeirdRubberDuck: I would have to create a new type, but yes, it is possible. 

However, the only things I can really put in that type are `name`, `description`, and `disabled`. All the other arguments will be specific to the given property type

WeirdRubberDuck: Would be good to discuss ideas for the property handling, if the current approach is confusing.

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
WeirdRubberDuck: Here I would prefer to keep the props object, to make it easy to spread to the value list view

The props handling for the vector types properties might need some discussion, in general

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -0,0 +1,88 @@
...
+
+  const min = additionalData.MinimumValue;
+  const max = additionalData.MaximumValue;
+  const step = additionalData.SteppingValue;
+  const exponent = additionalData.Exponent;
+
+  // When no min/max is set, the marks for the slider cannot be nicely computed
+  const extent = max - min;
+  // @TODO (202503-03, emmbr) This should be handled a better way...
+  const maxAllowedExtentForSlider = 10e12;
```
engbergandreas, line 43: Should it be 1e12 or 1e13 instead? 10e12 looks weird to me or am I forgetting something about scientific notations

src/components/Property/Types/NumericProperty/Slider/NumericPropertySlider.tsx
```diff
@@ -0,0 +1,71 @@
+import { MantineStyleProps, NumberFormatter, Slider, Stack } from '@mantine/core';
+
+import { usePropListeningState } from '@/api/hooks';
+import { useSliderScale } from '@/components/Property/SliderUtil/hook';
+
+import { SliderMinMaxLabels } from '../../../SliderUtil/SliderMinMaxLabels';
```
engbergandreas, line 6: We're getting deep on the nesting here, I get that this is technically in the same parent folder. But do we want to use the @ notation here?

src/components/Property/Types/StringProperty.tsx
```diff
@@ -1,27 +1,28 @@
...
+  if (disabled) {
+    return (
+      <Paper px={'sm'} py={5}>
+        <Text size={'sm'}>{value}</Text>
+      </Paper>
+    );
+  }
   return (
     <StringInput
       disabled={disabled}
```
engbergandreas, line 22: Why don't we want to use the string input if disabled? 
If we are to keep it this way we can remove line 22

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
```diff
@@ -0,0 +1,38 @@
+import { Flex, RGBA } from '@mantine/core';
+
+import { ColorPicker } from '@/components/ColorPicker/ColorPicker';
+import { VectorPropertyProps } from '@/components/Property/Types/VectorProperty/VectorProperty';
+
+import { ValueList } from './DefaultValueList';
+
+export function ColorVector(props: VectorPropertyProps) {
+  const { disabled, setPropertyValue, value, isInt } = props;
```
engbergandreas, line 9:
```suggestion
export function ColorVector({disabled, setPropertyValue, value, isInt} : VectorPropertyProps) {
```
??

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
```diff
@@ -0,0 +1,38 @@
...
+          b: Math.round(255 * value[2]),
+          a: hasAlpha ? value[3] : 1.0
+        }}
+        onChange={(rgbaColor: RGBA) => {
+          const newValue = [rgbaColor.r / 255, rgbaColor.g / 255, rgbaColor.b / 255];
+          if (hasAlpha) {
+            newValue.push(rgbaColor.a);
+          }
+          setPropertyValue(newValue);
+        }}
```
engbergandreas, line 34: check for min/max values here?

engbergandreas (commented):
Looking solid!! 🎉 
A few general things: 

1. Read only property grows larger than other properties 
image

2. Getting a bunch of redux warnings when making changes to many of the properties, don't know how big of an issue this actually is though.. 
image


3. The behaviour of sliders when either on the slider itself or using stepper controls are weird. Going down/back steps twice and forward/up nothing happens. Also, the focus highlight on the slider does not work in OpenSpace (browser is fine) 
https://github.com/user-attachments/assets/07ebb92d-7796-4191-8e08-4a3d70ba925f

4. We should have some limit on color inputs so that this is not valid, both picker and input: 
`rgb(30, 45, 2200)`

5. The text marker is not visible in input fields, here I'm positioned between 8 and 2
image

src/components/ColorPicker/ColorEdit.tsx
```diff
@@ -0,0 +1,88 @@
...
+          if (event.nativeEvent.code === 'Enter') {
+            event.currentTarget.blur();
+            if (isColorValid(textEditValue)) {
+              onColorChange(textEditValue);
+            } else {
+              setTextEditToCurrentFormat(value);
+            }
+          }
+        }}
+        onChange={setTextEditValue}
```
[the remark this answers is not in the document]
WeirdRubberDuck: The input fields should be able to take any type of input, but we should add warnings

The color however, do need some validation. Correct

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
WeirdRubberDuck: Unfortunately, no. This property type uses both the regular view for properties and for that one I need to spread all the inputted properties. And including the entire list here seemed less readable to me

Maybe it's just a bad design of the components. Would be open to discuss alternative solutions

WeirdRubberDuck:
> Looking solid!! 🎉 A few general things:
> 
> 1. Read only property grows larger than other properties
>    image
> 2. Getting a bunch of redux warnings when making changes to many of the properties, don't know how big of an issue this actually is though..
>    image
> 3. The behaviour of sliders when either on the slider itself or using stepper controls are weird. Going down/back steps twice and forward/up nothing happens. Also, the focus highlight on the slider does not work in OpenSpace (browser is fine)
>    https://github.com/user-attachments/assets/07ebb92d-7796-4191-8e08-4a3d70ba925f
> 4. We should have some limit on color inputs so that this is not valid, both picker and input:
>    `rgb(30, 45, 2200)`
> 5. The text marker is not visible in input fields, here I'm positioned between 8 and 2
>    image

1. ✅ Is fixed
2. ⛔ Yeah, not sure what to do about this. Requires some investigation into the redux I think. Separate from this PR.
3. Not sure what you mean by this one - can't reproduce the behavior you're describing
4. Yes! TODO
5. ✅ It is, but you need to use the OpenSpace branch for the UI update. That indicator depends on the fixes for focus in that branch

WeirdRubberDuck:
> Overall I think it looks awesome!! ✨ I just jotted down some notes from looking at the properties. Lmk if you want to discuss
> 
> 1. General
>    
>    1. Maybe we can make the titles less bold? They are a little hard to read now imo.
>    2. I just had a thought - we could potientially go around the min max problem by making them as tiny input fields so you could set them yourself.
> 2. Numeric property
>    
>    1. Overlap of text over the textbox
>       image
>    2. Is there a way to make these look more connected? When the window is small it is hard to see that these two are the same property.
>    3. Why does the slider have dots for these values? Seems 0.05 is always in there?
>    4. Should we somehow let the user know when they are out of range? Like we did in the old UI
>    5. Seems the circle slider doesn't go all the way to the right? I think there might be an offset to the blue bar? image
>       image
>    6. It seems when the property is line wrapped it goes outside bounds? image
> 3. String list property
>    Can we somehow make this smaller? Like a dropdown?
>    image
> 4. Vec2 property
>    In the gui it took me a while to realize this was probably a vec2 property. I assume this is x and y? Can we maybe write that out somehow if that is the case? This might go for all vec properties - we might want to write out xyzw?
>    image
> 5. Selection property
>    Make a placeholder when empty?
> 6. Min max
>    Should we place the slider in the middle of the the number? Or the slider above the numbers, to mimic the numeric property when it is linewrapped?
> 7. Infobox
>    Can we place it so it is not in different places depending on how long the title is? Maybe to the far right?
> 8. Propertyowner
>    
>    1. Some padding on the right hand side? image
>    2. Align about with the upper border of the about text?
>       image
> 9. Seems we have an issue with rendering
>    image
> 10. Wth 😆 We really like saying Earth is a special place
>     image

1. ✅ For now I changed the font weight of the input fields. I like the idea of min and max as input values, but we should save that for a future design iteration (maybe once we can set the min and max per object, even :) would be quite natural then) 
2. ✅ Essentially all of these points should be fixed. 
  - One remaining TODO is to provide error messages when the input value is out of range, like we did in the old UI. 
3. ✅ Gave this property type a max height and added a scrollbox
4. TODO: Improve vector properties
5. ✅ Done
6. TODO: Think about this and test out an alternative look
7. TODO: Discuss 
8.  i) ✅ Done ii) TODO
9. ✅ I don't remember what the cause for this was, but I fixed it :) 
10. ✅ Haha, indeed... We need to go through our descriptions ;)

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
WeirdRubberDuck: @engbergandreas also commented on this. Let's discuss it :)

src/components/ColorPicker/ColorEdit.tsx
```diff
@@ -0,0 +1,86 @@
...
+  color: RGBA;
+  onChange?: (color: RGBA) => void;
+  withAlpha: boolean;
+}
+
+export function ColorEdit({ color, onChange, withAlpha }: Props) {
+  const formats = withAlpha ? ['rgba', 'hexa', 'hsla'] : ['rgb', 'hex', 'hsl'];
+  const defaultFormat = withAlpha ? 'rgba' : 'rgb';
+
+  const [format, setFormat] = useState<ColorFormat | undefined>(defaultFormat);
```
ylvaselling, line 25: I don't think this can ever be undefined

src/components/ColorPicker/ColorEdit.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <MantineColorPicker
+        value={value}
+        // We don't use the selected format here, since changing the format would
+        // lead to the value changing, which would then trigger the onChange event
+        format={withAlpha ? 'rgba' : 'rgb'}
+        onChange={onColorChange}
+      />
+      <ColorInput
+        value={textEditValue}
+        withEyeDropper={false}
```
ylvaselling, line 60: I think we should enable this

src/components/ColorPicker/ColorEdit.tsx
```diff
@@ -0,0 +1,86 @@
...
+  return (
+    <Stack gap={'xs'}>
+      <MantineColorPicker
+        value={value}
+        // We don't use the selected format here, since changing the format would
+        // lead to the value changing, which would then trigger the onChange event
+        format={withAlpha ? 'rgba' : 'rgb'}
+        onChange={onColorChange}
+      />
+      <ColorInput
```
ylvaselling, line 58: Needs some way to focus on this popup with keyboard navigation

src/components/Property/Property.tsx
```diff
@@ -71,32 +72,38 @@ interface Props {
...
-    // console.error('Missing property type', property.description.type);
-    return null;
+    throw Error(`Missing property type: '${property.description.type}'`);
   }
 
-  // console.log(property);
+  // In most cases we want to show the label on top of the property input, but some types
+  // have a custom label built in in the input type. Keep track of which that is
+  const hasBuiltInLabel =
+    ConcreteProperty === BoolProperty || ConcreteProperty === TriggerProperty;
```
ylvaselling, line 94: These two properties have a built in label?

src/components/Property/PropertyLabel.tsx
```diff
@@ -1,16 +1,32 @@
...
+          {label}
+        </Text>
+        {isReadOnly && (
+          <Tooltip
+            maw={200}
+            multiline
+            label={`This property is read-only, meaning that it's not intended to be changed.`}
+          >
+            <Text span ml={'xs'} size={'xs'} c={'dimmed'}>
+              (Read-only)
```
ylvaselling, line 25: How about an icon for this?

src/components/Property/SliderUtil/hook.ts
```diff
@@ -0,0 +1,40 @@
...
+        },
+        { value: min + 1 * marksStep },
+        {
+          value: min + extent / 2
+        },
+        { value: min + 3 * marksStep },
+        {
+          value: max
+        }
+      ];
```
ylvaselling, line 25:
```suggestion
    :  [0, 1, 2, 3, 4].map((value, i) => { return { value: min + value * marksStep }});
```
Also this first array doesn't need the object with value in it:

```suggestion
    :  [0, 1, 2, 3, 4].map((value, i) => min + value * marksStep );
```

src/components/Property/SliderUtil/hook.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const scaledMarks = marks.map((mark) => ({
+    value: scale.invert(mark.value)
+  }));
```
ylvaselling, line 29:
```suggestion
  const scaledMarks = marks.map((mark) => ({
    value: scale.invert(mark)
  }));
```
This needs a comment - why is the scale inverting the mark?

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,153 @@
...
+  // The values to shown as a list of pills in the input
+  const {
+    value: shownValues,
+    setValue: setShownValues,
+    setIsEditing,
+    isEditing
+  } = usePropListeningState(value);
+
+  const [clickedItemIndex, setClickedItemIndex] = useState<number | undefined>(undefined);
```
ylvaselling, line 39: Group the useState hooks together

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+  const placeHolderText = useMemo(() => {
```
ylvaselling, line 41: This value is shadowing the state

src/components/Property/Types/ListProperty/ListProperty.tsx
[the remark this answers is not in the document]
ylvaselling: I would prefer it if there was one "base class" and then three "derived classes", i.e. make the StringListProperty component it's own component. It is not clear to me right now why the string is default 🤔 and if you have one specialized class for each property type you can place the specialized functions in them instead of having the switches

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,153 @@
...
+  return (
+    <PillsInput
+      disabled={disabled}
+      onBlur={() => stopEditing()}
+      aria-label={`List input for ${name}`}
+    >
+      <Pill.Group mah={100} style={{ overflowY: 'auto' }}>
+        {shownValues.map((v, i) => (
+          <Pill
+            key={`pill-${i}`}
```
ylvaselling, line 133: It is not advised to use index as the key

src/components/Property/Types/ListProperty/ListProperty.tsx
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  return (
+    <PillsInput
+      disabled={disabled}
+      onBlur={() => stopEditing()}
+      aria-label={`List input for ${name}`}
+    >
+      <Pill.Group mah={100} style={{ overflowY: 'auto' }}>
+        {shownValues.map((v, i) => (
```
ylvaselling, line 131:
```suggestion
        {shownValues.map((value, i) => (
```

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -0,0 +1,100 @@
...
+            <Text size={'sm'}>
+              <NumberFormatter
+                value={roundNumberToDecimalPlaces(currentValue, decimalPlaces)}
+              />
+            </Text>
+          </Paper>
+        ) : (
+          <NumericInput
+            value={currentValue}
+            valueLabel={(v) =>
```
ylvaselling, line 85:
```suggestion
            valueLabel={(value) =>
```

src/components/Property/Types/NumericProperty/Slider/NumericPropertySlider.tsx
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // @TODO (2025-02-05, emmbr): The slider value and resulting property value are not in
+  // sync, and this is something that the accessibility consultant commented on. We should
+  // make sure that the set value (i.e. including the scale) is property communicated to
+  // screen readers
+  return (
+    <Stack gap={0} {...styleProps}>
+      <Slider
+        label={(v) => roundNumberToDecimalPlaces(v, decimalPlaces)}
```
ylvaselling, line 57:
```suggestion
        label={(value) => roundNumberToDecimalPlaces(value, decimalPlaces)}
```

ylvaselling: Same for the rest of the code... I think we decided we weren't going to have one letter variables unless they were indices

src/components/Property/Types/NumericProperty/util.ts
```diff
@@ -0,0 +1,9 @@
+export function stepToDecimalPlaces(step: number): number {
+  const stepStr = step.toString();
+  const decimalPlaces = stepStr.includes('.') ? stepStr.split('.')[1].length : 0;
```
ylvaselling, line 3: Are you getting the last element of the array with the 1 index? how about `pop()` ?
```suggestion
  const decimalPlaces = stepStr.includes('.') ? stepStr.split('.')[1].length : 0;
```

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
ylvaselling: Yeah let's try redesign at some point :)

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import { ColorPicker } from '@/components/ColorPicker/ColorPicker';
+import { VectorPropertyProps } from '@/components/Property/Types/VectorProperty/VectorProperty';
+
+import { ValueList } from './DefaultValueList';
+
+export function ColorVector(props: VectorPropertyProps) {
+  const { disabled, setPropertyValue, value, isInt } = props;
+
+  if (value.length < 3 || value.length > 4 || isInt) {
```
ylvaselling, line 11:
```suggestion
  if (value.length !== 3 || value.length !== 4 || isInt) {
```

src/components/ColorPicker/ColorEdit.tsx
WeirdRubberDuck: Yep, noticed this yesterday too. As well as closing

src/components/Property/Property.tsx
WeirdRubberDuck: Yes, here we don't want to show the property label on top of the input, the same way we do for the others.  Coult try to clarify that in the comment?

src/components/ColorPicker/ColorEdit.tsx
WeirdRubberDuck: I would agree, and wondered why I disabled it.  It appears that unfortunately, it does not work in OpenSpace. And even if it did work, it would only work to pick colors form the UI itself, most likely. 

So I'd opt for removing it in all our color inputs

src/components/Property/Types/ListProperty/ListProperty.tsx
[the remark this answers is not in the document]
WeirdRubberDuck: 👍 that is a motivation to write it like that then. I'll change it

src/components/Property/Types/ListProperty/ListProperty.tsx
WeirdRubberDuck: Vell aware.  Not sure there is an alternative here, though?

src/components/Property/Property.tsx
ylvaselling: Aha I see now what you mean! I think the comment is fine actually now that i now what you mean

src/components/Property/SliderUtil/hook.ts
WeirdRubberDuck: Good catch!

src/components/Property/Types/VectorProperty/ViewOptions/ColorVector.tsx
WeirdRubberDuck: It would have to look like this: 
```
 if ((value.length !== 3 && value.length !== 4) || isInt) {
```
but sure

PR #57, opened 2025-02-18
Feature/dependency updates

WeirdRubberDuck:
Update all our dependencies that were marked as outdated to the latest major version - including React 19

Note however that it does not fix the severity vulnerabilities. I found no way around these, unfortunately
image

The React update commit can be reverted in case there are any issues, but I can't see any problematic things in the UI or when building. 

Please help test it! Making this a draft for now


ylvaselling (approved):

ylvaselling:
Mantine seems up to date to me

PR #60, opened 2025-02-19
Add property visibility to menu

ylvaselling:


engbergandreas (approved):

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -59,7 +60,14 @@ export function ViewMenu({ visibleMenuItems, setVisibleMenuItems }: Props) {
 
       <Menu.Item leftSection={<SaveIcon />}>Load/Save Layout</Menu.Item>
       <Menu.Divider />
-      <Menu.Item leftSection={<VisibilityIcon />}>User Visibility</Menu.Item>
+      <Menu.Label>
+        <Group>
+          <VisibilityIcon /> User Visibility
+        </Group>
+      </Menu.Label>
```
WeirdRubberDuck, line 67: Looks very busy with the eye icon. How about: 
```suggestion
      <Menu.Label>User Level</Menu.Label>
```
image

or even remove the label completely: 
```suggestion
```
image


If we want to apply it to other places than just properties it should be renamed anyways, so IMO the property label is probably sufficient

WeirdRubberDuck (approved):
Not a fan of the double

PR #63, opened 2025-02-20
Feature/layout component

ylvaselling:
* Adds new composite component layout 
* Has growing section and fixed section
* Both accept BoxProps
* Layout has ContainerProps


src/components/Layout/Layout.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const { ref: fixedContentRef, height: fixedContentHeight } = useElementSize();
+
+  return (
+    <LayoutContext.Provider
+      value={{
+        growingSizeHeight: parentHeight - fixedContentHeight,
+        ref: fixedContentRef
+      }}
+    >
+      <Container ref={parentRef} h={'100%'} {...props} px={'xs'} py={'md'}>
```
WeirdRubberDuck, line 20: I suggest we remove any build in padding from this component, and replace this with e.g: 
```suggestion
      <Boxref={parentRef} h={'100%'} {...props}>
```
Note that the container props should then be replaced with BoxProps

While it could be considered nice that there is already some styling applied to the Layout component, I think that we will end up in situations where we think it's limiting/annoying and will want to work around this. There is already a place where it's like that, in the scene menu.  Better to not include any padding at all in this component and leave i up to the user if a container should be added around it :)

WeirdRubberDuck (changes requested):
Loving it - simplifies the code a lot. 

I really only had one comment, but it will potentially lead to a cascade of small changes, so marking as "request changes" and will look at it again after that change to approve

WeirdRubberDuck:
Also, maybe apply this one to the ActionsPanel filter list? :)

src/components/FilterList/SearchResults/SearchResultsVirtualList.tsx
```diff
@@ -1,9 +1,11 @@
+import { MantineSpacing } from '@mantine/core';
+
 import { VirtualList } from '@/components/VirtualList/VirtualList';
 
 import { useSearchResultProvider } from './hooks';
 
 interface Props {
-  gap?: number; // Gap in pixels between items
+  gap?: MantineSpacing; // Gap in pixels between items
```
WeirdRubberDuck, line 8:
```suggestion
  gap?: MantineSpacing; // Gap between items
```

src/components/FilterList/FilterList.tsx
```diff
@@ -1,56 +1,19 @@
...
-  const calculatedHeight =
-    height ??
-    (heightPercent && windowHeight * (heightPercent / 100.0)) ??
-    (heightFunc && heightFunc(windowHeight)) ??
-    '100%';
-
+export function FilterList({ height = '100%', isLoading, children }: Props) {
   return (
-    <Stack style={{ height: calculatedHeight }}>
+    <Stack style={{ height }}>
```
WeirdRubberDuck, line 16:
```suggestion
    <Stack style={{ height }} gap={'xs'}>
```
:) ? Have been doing this change in some other PR I think, to reduce the cap between the filterlist parts. Would be nice to do it here too to reduce conflicting code

src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
```diff
@@ -57,7 +57,7 @@ export function FeaturedSceneTree() {
         data={featuredTreeData}
         renderNode={(payload) => <SceneTreeNode {...payload} />}
       />
-      <Divider my={'xs'} />
+      <Divider py={'xs'} />
```
WeirdRubberDuck, line 60: Unfortunately, it seems like the divider does not really support padding. This removed the spacing around the line...

 Try bringing back the margin? Did it cause problems?

WeirdRubberDuck (commented):
Here are some comments. Also, as we discussed, I can't quite approve this until the padding for the panels is fixed

WeirdRubberDuck (approved):

src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
WeirdRubberDuck: I made a fix for this using the `Space` component

WeirdRubberDuck:
After approval, I found a couple of issues:

1. The double scrollbars in the Skybrowser panel are now overlapping. Maybe the padding can not be added to the window after all... Seems like we might want the scroll area to go all the way to the edge in this case. 
image
  
 2. Expanding the top part of the exoplanets panel can lead to it being impossible to reach the bottom part (this was an issue before this PR though). ScrollArea needs to be added?  (would lead to the same isssue as for the Skybrowser panel above)
image

PR #64, opened 2025-02-20
Feature/friction buttons

engbergandreas:
add friction buttons to top menu bar

### DIFF
--- a/src/panels/Menu/TopMenuBar/Menus/FrictionMenu.tsx
+++ b/src/panels/Menu/TopMenuBar/Menus/FrictionMenu.tsx
@@ -15,7 +15,7 @@ export function FrictionMenu() {
         onChange={() => setRotation(!rotation)}
         variant={'light'}
         size={'xs'}
-        color="white"
+        color={"white"}
       >
         Rotation
       </Chip>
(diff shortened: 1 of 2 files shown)


WeirdRubberDuck (approved):

ylvaselling:
Can we move the scrollbar somehow? 
image

And add some padding to the right of the Roll button?

Also do we want to write that it is affecting friction somewhere?

WeirdRubberDuck:
> Can we move the scrollbar somehow? image
> 
> And add some padding to the right of the Roll button?
> 
> Also do we want to write that it is affecting friction somewhere?

We could maybe add an infobox to the very furthest right? 

image

Also, I think we should maybe consider setting the background of the top menu bar to something darker, for the contrast against the grays to be larger. Here's how it looks with `dark-9`
image

src/panels/Menu/TopMenuBar/Menus/FrictionMenu.tsx
```diff
@@ -0,0 +1,99 @@
...
+      </Chip>
+      <InfoBox
+        text={
+          <>
+            <Text size={'md'} fw={'bold'}>
+              Friction controls:
+            </Text>
+            <Text>
+              Enable or disable rotation , zoom , and roll friction respectively. If
+              checked, applies friction to camera movement.
```
WeirdRubberDuck, line 63:
```suggestion
              Enable or disable rotation, zoom, and roll friction. If checked, applies
              friction to camera movement.
```

WeirdRubberDuck (approved):

PR #65, opened 2025-02-21
parse negative years and years between 0-100 correctly

engbergandreas:


WeirdRubberDuck (commented):
Awesome!  Great improvement! But might need a bit more testing. 

Now it's possible to set the year, but when setting too large positive or negative years, the time part of the input stops working. Looks like maybe 6 digits may be the limit for positive numbers, and 4 digits the limit for negative ones. 

How about determining which ranges work for now and do some error checking for years outside the range that works?  I think we should deal with it in some way now while were looking at this anyways. Otherwise, we would have to remember what worked or not later on

Here's a video showing what I mean with the time part breaking. Note how when that happens, the time still ticks in OpenSpace when looking in the ImGui panel.  (Also, note how our UI is definitely superior to the old one 😉 NaN NaN NaN...)

https://github.com/user-attachments/assets/1aae2277-dafa-4204-a6a9-b0394ce29be3

engbergandreas:
> Now it's possible to set the year, but when setting too large positive or negative years, the time part of the input stops working. Looks like maybe 6 digits may be the limit for positive numbers, and 4 digits the limit for negative ones.

So the issue here is how we get the date string back from OpenSpace and is explained in the parser `redux/time/util`

> function parseBCYears(date: string) {
> // Year < 1000 B.C have a different format, B.C. YYYY MM DD HH:MM:SSS
>   // Due to the fixed length of the string we get from OpenSpace, we might also be
>   // missing information regarding hh:mm:ss for example:
>   // "B.C. 202969 JAN 09 16:5" and  "B.C. 3693 JUL 19 17:29:"
> 
>   // If we are unable to get month and day from the string chances are that the date is
>   // very far back in the past and that the month, day, and time is kind of meaningless
>   // We build a fake date that atleast uses the correct year to display *something*

I had a discussion with Alex about this, and rather than making major changes to the engine we came to the conclusion that setting hh:mm:ss to 0 would be sufficient for now, also since we are so far back maybe its not that relevant anymore?

ylvaselling:
Overall it looks amazing!! ✨ 

However I found two issues - maybe not related to this pr?

1) it is possible to enter a very large number (click enter). Then the number in the input field and the number in the bottom bar are different. Maybe we should reset the number when clicking enter?

https://github.com/user-attachments/assets/afdbe7f5-4d76-4d9a-90c9-ae5f8885ca1a

2) When holding the up/down button without shift, and then hold shift and click, the button doesn't register the input.

https://github.com/user-attachments/assets/3150661a-4e83-4cc7-a92d-c7338e6753a8

src/panels/TimePanel/util.ts
```diff
@@ -0,0 +1,42 @@
...
+    { limit: 12, factor: 12, unit: 'year' }
+  ];
+
+  for (const { limit, factor, unit: nextUnit } of timeUnits) {
+    if (increment < limit) {
+      break;
+    }
+    increment /= factor;
+    unit = nextUnit;
+  }
```
ylvaselling, line 27: This is a bit hard to follow... i tried to come up with a simpler way to write this but couldn't really hehe. Maybe we can come up with something together?

src/panels/TimePanel/TimePanelMenuButton.tsx
```diff
@@ -39,47 +42,19 @@ export function TimePanelMenuButton({ onClick }: TimePanelMenuButtonProps) {
...
-    return { increment, unit };
-  }
-
   return (
     <Button onClick={onClick} size={'xl'} variant={'menubar'} disabled={!isReady}>
       <Stack gap={0} align={'flex-start'}>
         {isReady ? (
           <>
-            <Text size={'lg'}>{timeLabel}</Text>
+            <Text size={'lg'}>{isValidDate ? timeLabel : backupTimeString}</Text>
```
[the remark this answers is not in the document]
engbergandreas: I believe this is to force the skeleton to a certain size? @WeirdRubberDuck

src/panels/TimePanel/TimeInput/TimeInput.tsx
```diff
@@ -159,10 +194,26 @@ export function TimeInput() {
     });
   }
 
+  function validateYear(value: number | string): void {
+    if (typeof value === 'string') {
+      return;
+    }
```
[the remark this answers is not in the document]
engbergandreas: I added a comment about it. In short, the onChange can give us either a string or a number. But our wrapper only handles numbers anyway so ignoring the case here

ylvaselling:
Looks pretty good overall! 🎉 

There are still some issues however: 
* still an issue with year -1000. If I set it to -1000, it becomes -1001, and i cant go "up" from there. Do we need a special case for the year 1000 in the parsing of the time string or something?
* Clicking "down" on any time control when we are in years before 1000 bc changes the day. "up" doesn't do anything. If we say that the time is not important before year -1000, maybe we can disable the time controls? And show an alert message? 
* Clicking "down" on the year -271820 or up on year 275760 doesn't do anything but also doesn't display an error message. Maybe we can disable the up and down for these cases? If we have these as min / max that should be pretty easy right?

I don't know how often people will run into these cases, but somehow it feels like it might be a while until someone will work on this again, so it might be nice to just communicate what is happening to the user :-)

src/components/Input/StringInput.tsx
```diff
@@ -14,8 +14,11 @@ export interface Props extends TextInputProps {
  * the value on ESCAPE.
  */
 export function StringInput({ onEnter, value, errorCheck, ...props }: Props) {
-  const { value: storedValue, set: setStoredValue } =
-    usePropListeningState<string>(value);
```
WeirdRubberDuck, line 18: Guessing that the changes in the Input component swill disappear if you merge in master in this branch?

src/redux/time/timeSlice.ts
```diff
@@ -14,6 +14,7 @@ export interface TimeState {
   nextDeltaTimeStep?: number;
   prevDeltaTimeStep?: number;
   deltaTimeSteps?: number[];
+  timeString?: string; // This is the raw OpenSpace time string we receive, can be used as backup if parsing fails.
```
WeirdRubberDuck, line 17:
```suggestion
  timeString?: string; // This is the raw OpenSpace time string we receive, can be used as
                       // backup if parsing fails
 ```

Comment (WeirdRubberDuck) [reply src/redux/time/timeSlice.ts:17]: ? We have various places where comments and strings overflow the allowed linewidth though...

File: src/util/time.ts [hunk 6]
```diff
@@ -2,11 +2,49 @@ export function dateToOpenSpaceTimeString(date: Date): string {
...
+    // For negative years we have to parse the string in B.C. epoch format for spice.
+    // https://naif.jpl.nasa.gov/pub/naif/toolkit_docs/C/cspice/str2et_c.html
+    else {
+      // (year <= 0)
+      const parsedUtcYear = parseInt(utcYear);
+
+      // A.D - B.C years does not have year 0 so we shift B.C by one. From str2et_c exmaples:
+      // The following date: 18 B.C. Jun 3, 12:29:28.291 becomes -017 Jun 03 12 29 28.291
+
+      // TODO anden88 2025-02-27: * Something * happens between 999 and 1001 BC where
```
WeirdRubberDuck, line 39:
```suggestion
      // @TODO (anden88 2025-02-27): *Something* happens between 999 and 1001 BC where
```

src/util/time.ts
```diff
@@ -2,11 +2,49 @@ export function dateToOpenSpaceTimeString(date: Date): string {
   // Spice, that is handling the time parsing in OpenSpace does not support
```
WeirdRubberDuck, line 2:
```suggestion
  // Spice, which is handling the time parsing in OpenSpace, does not support
```

src/panels/TimePanel/TimeInput/TimeInput.tsx
```diff
@@ -159,10 +206,29 @@ export function TimeInput() {
...
+    // Since our wrapper does not handle string values, we ignore them.
+    if (typeof value === 'string') {
+      return;
+    }
+
+    // Validate the new date to ensure the year is within the valid JavaScript Date range
+    const newTime = new Date(timeRef.current);
+    const ms = newTime.setUTCFullYear(value);
+
+    if (isNaN(ms)) {
```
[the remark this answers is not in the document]
engbergandreas: When setting the year/day/month w/e you get back the date object in milliseconds, but I've changed it to instead use the `isDateValid` function instead

src/panels/TimePanel/TimeInput/TimeInput.tsx
```diff
@@ -159,10 +206,29 @@ export function TimeInput() {
...
+    } else {
+      setErrorMessage('');
+    }
+  }
+
   if (cappedTime === undefined) {
     return (
-      <Center p={'xl'}>
-        <Text>Date out of range</Text>
+      <Center p={'xl'} style={{ flexDirection: 'column' }}>
```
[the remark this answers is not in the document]
engbergandreas: with: 
image

without: 
image

engbergandreas:
>     * still an issue with year -1000. If I set it to -1000, it becomes -1001, and i cant go "up" from there. Do we need a special case for the year 1000 in the parsing of the time string or something?

This is a tricky one, and I wonder if this should instead be made an issue in the OpenSpace repo instead. I haven't tested it yet, but one solution might be to check specifically the year 1000 B.C and whether or not we are trying to go up or down from that year, and based on that, add the necessary offset. However, this changes the surrounding functions more than I'd like..
 
>     * Clicking "down" on any time control when we are in years before 1000 bc changes the day. "up" doesn't do anything. If we say that the time is not important before year -1000, maybe we can disable the time controls? And show an alert message?

This issue is because the time string is cut from OpenSpace and we manually set time to 00:00:00, this will be fixed if we can parse the string correctly. Alex is tasked to extend the buffer so we can get the full time string with 6 digit years

>     * Clicking "down" on the year -271820 or up on year 275760 doesn't do anything but also doesn't display an error message. Maybe we can disable the up and down for these cases? If we have these as min / max that should be pretty easy right?

Added error message when trying to go above / below the range, I have ideas on how to disable the inputs, but they are quite nasty, so I'd rather avoid doing that

PR #67, opened 2025-02-25
Don't update/rerender input field while editing

engbergandreas:


src/api/hooks.ts
```diff
@@ -326,11 +326,18 @@ export function useSetOpenSpaceTime() {
  * Hook that listens to a prop and updates the local state when the prop changes.
```
WeirdRubberDuck, line 326: This comment could use an update to explain the isEditing feature

WeirdRubberDuck (changes requested):
It doesn't work in Cef :(  Notice how after the value is edited, it stops updating based on value changes 

https://github.com/user-attachments/assets/8b993106-e3c7-4581-9bd9-5bf890c28d0f

engbergandreas:
> > It doesn't work in Cef :( Notice how after the value is edited, it stops updating based on value changes
> > Guessing that blur is maybe not handled correctly...
> > 2025-02-26.09-54-07.mov
> 
> Relaized that this may be fixed when using the OpenSpace branch that fixes our issues with focus. Will investigate

Yeah, can't seem to reproduce on my end using the new OpenSpace branch

WeirdRubberDuck (approved):
It is indeed fixed in the new OpenSpace branch

PR #69, opened 2025-02-26
Add hints for how many layers there are in each layer group

WeirdRubberDuck:
Including active layers

image

### DIFF
--- a/src/util/propertyTreeHelpers.ts
+++ b/src/util/propertyTreeHelpers.ts
@@ -136,15 +136,15 @@ export function isPropertyOwnerHidden(uri: Uri, properties: Properties) {
   return isHidden || false;
 }
 
+export function isPropertyOwnerActive(uri: Uri, properties: Properties): boolean {
+  const enabledValue = properties[enabledPropertyUri(uri)]?.value as boolean | undefined;
+  const fadeValue = properties[fadePropertyUri(uri)]?.value as number | undefined;
+  return checkVisiblity(enabledValue, fadeValue) || false;
+}
+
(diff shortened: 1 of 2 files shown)


engbergandreas (approved):

src/util/propertyTreeHelpers.ts
```diff
@@ -138,8 +138,12 @@ export function isPropertyOwnerHidden(uri: Uri, properties: Properties) {
...
 export function isSceneGraphNodeVisible(uri: Uri, properties: Properties): boolean {
   const renderableUri = sgnRenderableUri(uri);
-  const enabledValue = properties[enabledPropertyUri(renderableUri)]?.value as boolean;
-  const fadeValue = properties[fadePropertyUri(renderableUri)]?.value as number;
+  return isPropertyOwnerActive(renderableUri, properties);
+}
+
+export function isPropertyOwnerActive(uri: Uri, properties: Properties): boolean {
+  const enabledValue = properties[enabledPropertyUri(uri)]?.value as boolean;
+  const fadeValue = properties[fadePropertyUri(uri)]?.value as number;
```
ylvaselling, line 146: Do we need to handle if this returns undefined? What does undefined cast to number mean? 0?

Otherwise, in this file I think it would be better if the functions were defined before they were used. Here they appear in reverse order (functions use functions that are defined below them)

WeirdRubberDuck: Good catch!  Just copied these rows, but there is an `| undefined` missing after the `as boolean`

PR #70, opened 2025-02-26
Feature/actions panel bug fix

ylvaselling:
Fixes a bug reported by Adam that when you had folders names that started with the same word, the breadcrumbs didn't work. Also the actions didn't show up in the correct folder.

I tried to refactor and simplify this code to make it easier to understand and more robust


engbergandreas (approved):
looks great, I feel like its a lot easier to follow along

PR #71, opened 2025-02-26
Feature/keybinds listview

engbergandreas:
Adds a list view for the key binds, use search for name, key, modifiers, or documentation

### DIFF
--- a/src/panels/KeybindsPanel/KeybindButtons.tsx
+++ b/src/panels/KeybindsPanel/KeybindButtons.tsx
@@ -20,7 +20,7 @@ export function KeybindButtons({ modifiers, selectedKey }: KeybindButtonsProps) 
   const allKeys = [...modifiers, displayKey].filter((kbd) => kbd);
 
   return (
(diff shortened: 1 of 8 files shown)


ylvaselling (approved):

ylvaselling:
Some suggestions:

* Highlight the selected keybind
* When clicking the keybind again, deselect
* Have the modifier buttons next to input field with the modifiers, either as a filter selection like in the scene menu or like buttons (see below)


image

engbergandreas:
Discussed with Emma the multiselect version, which looked way worse. Decided on removing super as a filter option since its rarely used anyway and went back to the chip version

ylvaselling:
Looks awesome!! I think it would look even better with a bit more padding at the top and bottom. Maybe the same as between the action card and the list?
image

engbergandreas:
See what you think of the new padding @ylvaselling

ylvaselling:
Looks great! 🎉

PR #74, opened 2025-02-28
Add touch and focus support for infobox tooltips

WeirdRubberDuck:
Realized that Mantine already had support for accessible tooltips, and for touch!  
https://mantine.dev/core/tooltip/#accessibility

Applied those settings for our infoboxes, but we should check if it's good enough for our touch use-case. 

Let me know if you think this solution is good enough, or if we should do something more advanced

### DIFF
--- a/src/components/InfoBox/InfoBox.tsx
+++ b/src/components/InfoBox/InfoBox.tsx
@@ -9,8 +9,14 @@ interface Props {
 
 export function InfoBox({ text, w = 220 }: Props) {
   return (
-    <Tooltip label={text} multiline w={w} offset={{ mainAxis: 5, crossAxis: 100 }}>
-      <ThemeIcon radius={'xl'} size={'xs'}>
+    <Tooltip
+      label={text}
+      multiline
+      w={w}
+      offset={{ mainAxis: 5, crossAxis: 100 }}
+      events={{ hover: true, focus: true, touch: true }}
+    >
+      <ThemeIcon tabIndex={0} radius={'xl'} size={'xs'}>
         <InformationIcon />
       </ThemeIcon>
     </Tooltip>


src/components/InfoBox/InfoBox.tsx
```diff
@@ -9,8 +9,14 @@ interface Props {
...
 export function InfoBox({ text, w = 220 }: Props) {
   return (
-    <Tooltip label={text} multiline w={w} offset={{ mainAxis: 5, crossAxis: 100 }}>
-      <ThemeIcon radius={'xl'} size={'xs'}>
+    <Tooltip
+      label={text}
+      multiline
+      w={w}
+      offset={{ mainAxis: 5, crossAxis: 100 }}
+      events={{ hover: true, focus: true, touch: true }}
```
ylvaselling, line 17: Do we want it to appear on focus? Or on enter when we have focused? It might be a lot of information popping up when you tab through the interface if you are using a screenreader

WeirdRubberDuck: Yeah I felt unsure of that as well.  Seems like it might need some discussion

engbergandreas: I think it would be better to make them toggleable, that way we also solve the issue with unclickable links in the tooltip description

engbergandreas: Or perhaps that should be its tooltip version? Otherwise, we will get into trouble anytime we put a tooltip in a button..

WeirdRubberDuck: This is the infobox, not the general tooltip. But yeah, tooltips for buttons is a different situation entirely that we also need to think about. I think it might be fine to not have these tooltips accessible using touch, though. 

But I'll close this PR for now then. Too simple fix for the problem :)

PR #75, opened 2025-03-03
add mission selection if multiple ones are loaded

engbergandreas:
Requires latest OpenSpace gui branch [here](https://github.com/OpenSpace/OpenSpace/commit/9fa57a47b8b9142a97d2f25934dafda5edd57642)

You can now load multiple mission files and select which one you want to view. 
Remembers which mission was last viewed if the panel is closed
Keep the same mission viewed if more are loaded at runtime
Fallback to first mission file in list if the one viewed is removed

### DIFF
--- a/src/panels/MissionsPanel/MissionContent.tsx
+++ b/src/panels/MissionsPanel/MissionContent.tsx
@@ -23,6 +23,7 @@ export function MissionContent({ missionOverview }: Props) {
     data: undefined
   });
 
+  // Reset phases when selected mission is changed
   useEffect(() => {
     // When the mission is changed, display overview again
     setDisplayedPhase({
--- a/src/panels/MissionsPanel/MissionContent.tsx
+++ b/src/panels/MissionsPanel/MissionContent.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState } from 'react';
+import { useEffect, useMemo, useState } from 'react';
 import { Box, Button, Group, Switch, Text, Title } from '@mantine/core';
 
 import { useSubscribeToTime } from '@/api/hooks';
@@ -22,6 +22,20 @@ export function MissionContent({ missionOverview }: Props) {
     type: undefined,
     data: undefined
   });
+
+  useEffect(() => {
+    // When the mission is changed, display overview again
+    setDisplayedPhase({
+      type: DisplayType.Overview,
+      data: missionOverview
+    });
+    // Avoid potentially showing information from a previous mission
+    setLastDisplayedPhase({
+      type: DisplayType.Overview,
+      data: missionOverview
+    });
+    setDisplayCurrentPhase(false);
+  }, [missionOverview]);
 
   const now = useSubscribeToTime();
 
@@ -104,14 +118,14 @@ export function MissionContent({ missionOverview }: Props) {
   }
 
   return (
-    <Group wrap={'nowrap'} align={'start'} h={'100%'} gap={0}>
+    <Group wrap={'nowrap'} align={'start'} gap={'xs'}>
       <TimeLine
         allPhasesNested={allPhasesNested}
         displayedPhase={displayedPhase}
         missionOverview={missionOverview}
         setDisplayedPhase={setPhaseManually}
       />
-      <Box px={'md'} h={'100%'} style={{ overflow: 'auto' }}>
+      <Box flex={1}>
         <Group justify={'space-between'} mb={'md'}>
           <Title order={2}>{missionOverview.name}</Title>
           <Button
--- a/src/panels/MissionsPanel/MissionsPanel.tsx
+++ b/src/panels/MissionsPanel/MissionsPanel.tsx
@@ -5,15 +5,19 @@ import { RocketLaunchIcon } from '@/icons/icons';
 import { useAppDispatch, useAppSelector } from '@/redux/hooks';
 import { setSelectedMission } from '@/redux/missions/missionsSlice';
 
-import { useSelectedMission } from './hooks';
 import { MissionContent } from './MissionContent';
 
 export function MissionsPanel() {
-  const { data, selectedMissionIdentifier } = useAppSelector((state) => state.missions);
+  const { isInitialized, missions, selectedMissionIdentifier } = useAppSelector(
+    (state) => state.missions
+  );
 
   const luaApi = useOpenSpaceApi();
-  const { hasMission, mission } = useSelectedMission();
   const dispatch = useAppDispatch();
+
+  // Grab the mission last viewed or if it does not exist anymore, take the first one
+  const mission = missions[selectedMissionIdentifier] ?? Object.values(missions)[0];
+  const hasMission = isInitialized && mission !== undefined;
 
   function onMissionSelected(identifier: string | null) {
     if (!identifier) {
@@ -39,12 +43,12 @@ export function MissionsPanel() {
 
   return (
     <Box>
-      {data.missions.length > 1 && (
+      {Object.keys(missions).length > 1 && (
         <Select
           label={'Selected mission'}
           placeholder={'Select a mission'}
-          data={data.missions.map((mission) => {
-            return { value: mission.identifier, label: mission.name };
+          data={Object.entries(missions).map(([identifier, mission]) => {
+            return { value: identifier, label: mission.name };
           })}
           value={selectedMissionIdentifier}
           onChange={onMissionSelected}
--- a/src/panels/MissionsPanel/MissionsPanel.tsx
+++ b/src/panels/MissionsPanel/MissionsPanel.tsx
@@ -15,9 +15,7 @@ export function MissionsPanel() {
   const luaApi = useOpenSpaceApi();
   const dispatch = useAppDispatch();
 
-  // Grab the mission last viewed or if it does not exist anymore, take the first one
-  const mission = missions[selectedMissionIdentifier] ?? Object.values(missions)[0];
-  const hasMission = isInitialized && mission !== undefined;
+  const hasMission = isInitialized && Object.values(missions).length > 0;
 
   function onMissionSelected(identifier: string | null) {
     if (!identifier) {
@@ -27,7 +25,7 @@ export function MissionsPanel() {
     luaApi?.setCurrentMission(identifier);
   }
 
-  if (!hasMission || !mission) {
+  if (!hasMission) {
     return (
       <Stack h={'100%'} w={'100%'} ta={'center'} align={'center'} p={'lg'}>
         <Title order={2}>No mission loaded</Title>
@@ -40,6 +38,9 @@ export function MissionsPanel() {
       </Stack>
     );
   }
+
+  // Grab the mission last viewed or if it does not exist anymore, take the first one
+  const mission = missions[selectedMissionIdentifier] ?? Object.values(missions)[0];
 
   return (
     <Box>
--- a/src/panels/MissionsPanel/MissionsPanel.tsx
+++ b/src/panels/MissionsPanel/MissionsPanel.tsx
@@ -23,7 +23,7 @@ export function MissionsPanel() {
     luaApi?.setCurrentMission(identifier);
   }
 
-  if (!hasMission) {
+  if (!hasMission || !mission) {
     return (
       <Stack h={'100%'} w={'100%'} ta={'center'} align={'center'} p={'lg'}>
         <Title order={2}>No mission loaded</Title>
@@ -52,7 +52,7 @@ export function MissionsPanel() {
           allowDeselect={false}
         />
       )}
-      <MissionContent missionOverview={mission!} />
+      <MissionContent missionOverview={mission} />
     </Box>
   );
 }
--- a/src/panels/MissionsPanel/MissionsPanel.tsx
+++ b/src/panels/MissionsPanel/MissionsPanel.tsx
@@ -1,13 +1,41 @@
-import { Stack, Text, ThemeIcon, Title } from '@mantine/core';
+import { Box, Select, Stack, Text, ThemeIcon, Title } from '@mantine/core';
 
+import { useOpenSpaceApi } from '@/api/hooks';
 import { RocketLaunchIcon } from '@/icons/icons';
-import { useAppSelector } from '@/redux/hooks';
+import { useAppDispatch, useAppSelector } from '@/redux/hooks';
+import { setSelectedMission } from '@/redux/missions/missionsSlice';
 
 import { MissionContent } from './MissionContent';
 
 export function MissionsPanel() {
-  const hasMission = useAppSelector((state) => state.missions.isInitialized);
-  const mission = useAppSelector((state) => state.missions.data.missions[0]);
+  const {
+    isInitialized: hasMission,
+    data,
+    selectedMissionIdentifier
+  } = useAppSelector((state) => state.missions);
+
+  const luaApi = useOpenSpaceApi();
+  const dispatch = useAppDispatch();
+
+  let [mission] = data.missions;
+  // If a new missions is added, the missions order might have changed so we try to find
+  // the last one viewed
+  if (hasMission && selectedMissionIdentifier) {
+    const selectedMission = data.missions.find(
+      (_mission) => _mission.identifier === selectedMissionIdentifier
+    );
+    if (selectedMission) {
+      mission = selectedMission;
+    }
+  }
+
+  function onMissionSelected(identifier: string | null) {
+    if (!identifier) {
+      return;
+    }
+    dispatch(setSelectedMission(identifier));
+    luaApi?.setCurrentMission(identifier);
+  }
 
   if (!hasMission) {
     return (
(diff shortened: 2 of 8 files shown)


src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -104,14 +118,14 @@ export function MissionContent({ missionOverview }: Props) {
...
   return (
-    <Group wrap={'nowrap'} align={'start'} h={'100%'} gap={0}>
+    <Group wrap={'nowrap'} align={'start'} gap={'xs'}>
       <TimeLine
         allPhasesNested={allPhasesNested}
         displayedPhase={displayedPhase}
         missionOverview={missionOverview}
         setDisplayedPhase={setPhaseManually}
       />
-      <Box px={'md'} h={'100%'} style={{ overflow: 'auto' }}>
```
engbergandreas, line 114: I couldn't see any noticeable differences with the height and overflow css here, so I removed it (is it legacy from before @ylvaselling  refactor?) and moved the padding to the parent gap prop instead

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -1,13 +1,41 @@
...
+    isInitialized: hasMission,
+    data,
+    selectedMissionIdentifier
+  } = useAppSelector((state) => state.missions);
+
+  const luaApi = useOpenSpaceApi();
+  const dispatch = useAppDispatch();
+
+  let [mission] = data.missions;
+  // If a new missions is added, the missions order might have changed so we try to find
```
engbergandreas, line 21:
```suggestion
  // If a new mission is added, the missions order might have changed, so we try to find
```

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -1,13 +1,41 @@
...
+  const {
+    isInitialized: hasMission,
+    data,
+    selectedMissionIdentifier
+  } = useAppSelector((state) => state.missions);
+
+  const luaApi = useOpenSpaceApi();
+  const dispatch = useAppDispatch();
+
+  let [mission] = data.missions;
```
ylvaselling, line 20: Personally I would prefer using this type of restructuring for hooks 
```suggestion
  let mission = data.missions.at(0);
```

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -1,13 +1,41 @@
...
+  // If a new missions is added, the missions order might have changed so we try to find
+  // the last one viewed
+  if (hasMission && selectedMissionIdentifier) {
+    const selectedMission = data.missions.find(
+      (_mission) => _mission.identifier === selectedMissionIdentifier
+    );
+    if (selectedMission) {
+      mission = selectedMission;
+    }
+  }
```
ylvaselling, line 30: It seems like the selectedMission takes precedence over getting the first mission, so how about we make this a function / hook and return the index 0 in an else statement? That would make it clearer that we first check for the selected mission and then as a fallback the first element in the array. Plus the hook could return { hasMission, mission } and it would look kinda neat I think
```suggestion
  if (hasMission && selectedMissionIdentifier) {
    const selectedMission = data.missions.find(
      (_mission) => _mission.identifier === selectedMissionIdentifier
    );
    if (selectedMission) {
      mission = selectedMission;
    }
    else {
        mission = data.missions.at(0);
    }
  }
```

src/redux/events/types.ts
```diff
@@ -64,10 +64,20 @@ type InterpolationFinishedEvent = {
...
+
 type MissionEventReachedEvent = {
   Event: 'MissionEventReached';
 };
 
+type MissionRemovedEvent = {
+  Event: 'MissionRemoved';
+  MissionName: string;
+};
+
```
ylvaselling, line 80: Place these two next to each other?

src/redux/missions/missionsMiddleware.ts
```diff
@@ -12,12 +16,15 @@ interface MissionData {
...
+async function getMissions(callback: (missions: { missions: Phase[] } | null) => void) {
   const missionTopic = api.startTopic('missions', {});
   const { value } = (await missionTopic.iterator().next()) as MissionData;
-  if (value.missions !== null) {
+  if (value.missions) {
     callback({ missions: value.missions });
+  } else {
+    callback(null);
   }
+
```
ylvaselling, line 27: Can we place the code in the `getMissions` function inside of the `effect`? This is kind of hard to read atm

src/redux/missions/missionsMiddleware.ts
```diff
@@ -26,7 +33,22 @@ export const addMissionsListener = (startListening: AppStartListening) => {
...
+  startListening({
+    actionCreator: refreshMissions,
+    effect: (_, api) => {
+      getMissions((missions) => {
+        if (missions) {
+          api.dispatch(initializeMissions(missions));
+        } else {
+          api.dispatch(clearMissions());
+        }
```
ylvaselling, line 51: Why don't the openconnection and the refreshmissions have the same callback? Shouldn't they both do the same thing?

src/redux/missions/missionsSlice.ts
```diff
@@ -44,11 +46,31 @@ export const missionsSlice = createSlice({
...
+
+      // If no mission was loaded or if the previously selected mission was removed,
+      // automatically select first available mission from the updated list
+      const isSelectedMissionLoaded = state.data.missions.some(
+        (mission) => mission.identifier === state.selectedMissionIdentifier
+      );
+      if (!isSelectedMissionLoaded) {
+        state.selectedMissionIdentifier = action.payload.missions[0].identifier;
+      }
+      return state;
```
ylvaselling, line 58: Now that it is possible to have multipe missions, maybe we want to store them as a map with a key first? that way the finds are easier

src/panels/MissionsPanel/MissionContent.tsx
```diff
@@ -23,6 +23,20 @@ export function MissionContent({ missionOverview }: Props) {
     data: undefined
   });
```
ylvaselling, line 25: Make a comment here that describes that this resets the phases when the selected mission has changed 🙏

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -23,5 +51,22 @@ export function MissionsPanel() {
...
-  return <MissionContent missionOverview={mission} />;
+  return (
+    <Box>
+      {data.missions.length > 1 && (
+        <Select
+          label={'Selected mission'}
+          placeholder={'Select a mission'}
+          data={data.missions.map((_mission) => {
+            return { value: _mission.identifier, label: _mission.name };
+          })}
```
ylvaselling, line 62: Personally I'm not a huge fan of the `_variableName` style. It looks to me like the member variables in OpenSpace, and for some reason I also find it hard to read... I would prefer a shorter name here like `entry`, `mis` , `element` or whatever

src/panels/MissionsPanel/types.ts
```diff
@@ -17,6 +17,8 @@ export interface Phase {
   link: string;
   milestones: Milestone[];
   capturetimes: string[];
+  identifier: string; // TOOD anden88 2025-03-03: The identifier only exists on the outermost
```
ylvaselling, line 20: Should this be of type Identifier?

ylvaselling (changes requested):

ylvaselling:
Great work! 🎉 

Do we want to show the drop down at all times like in the geo position panel? It looks a bit jumpy when the new mission is added. Maybe the drop down menu could serve as the title and we could get rid of the title? Also there is quite a lot of padding on top of the menu. and we should unify the text above the dropdown with how it looks in the geo panel

ylvaselling:
I saw that there was no way to comment on the code in the engine so jotting down some notes here:

* Should the identifier be the name - rename "identifier" to "name"? I think we have special rules for identifiers (such as no spaces etc). Here you could also create an object with keys instead of the array as suggested above.
* In `event.cpp` should we set the "MissionName" variable to just "Name"? I think Mission is implied already in the event

src/panels/MissionsPanel/MissionsPanel.tsx
engbergandreas: at first I did `data.missions[0]` but the linter screamed at me, I know we've done this in several other places so we should be consistent if we want to do the `at(0)` version

src/redux/events/types.ts
engbergandreas: Sorted in alphabetical order

src/panels/MissionsPanel/MissionsPanel.tsx
engbergandreas: Checkout the new hook pls and thx

src/redux/missions/missionsSlice.ts
engbergandreas: I hear ya, I'm wondering if it is worth the extra overhead when receiving the data from OpenSpace since we will, for the most part, have 1 mission loaded. If users have loaded multiple ones, I doubt it will be more than 2-3 at most?

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -23,5 +37,22 @@ export function MissionsPanel() {
...
+          data={data.missions.map((mission) => {
+            return { value: mission.identifier, label: mission.name };
+          })}
+          value={selectedMissionIdentifier}
+          onChange={onMissionSelected}
+          my={'xs'}
+          allowDeselect={false}
+        />
+      )}
+      <MissionContent missionOverview={mission!} />
```
WeirdRubberDuck, line 55:
```suggestion
      <MissionContent missionOverview={mission} />
```

and move the check to the if statement above? See other comment

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -1,13 +1,27 @@
...
+
+  function onMissionSelected(identifier: string | null) {
+    if (!identifier) {
+      return;
+    }
+    dispatch(setSelectedMission(identifier));
+    luaApi?.setCurrentMission(identifier);
+  }
 
   if (!hasMission) {
```
WeirdRubberDuck, line 26:
```suggestion
  if (!hasMission || !mission) {
```
and remove exclamation mark below

src/redux/missions/missionsSlice.ts
```diff
@@ -5,13 +5,15 @@ import { Phase } from '@/panels/MissionsPanel/types';
...
+  selectedMissionIdentifier: string;
 }
 
 const initialState: MissionState = {
   isInitialized: false,
   data: {
     missions: []
-  }
+  },
+  selectedMissionIdentifier: ''
```
WeirdRubberDuck, line 16: do we want to use empty or undefined here?

src/redux/missions/missionsSlice.ts
```diff
@@ -44,11 +46,31 @@ export const missionsSlice = createSlice({
...
       state.isInitialized = true;
       state.data.missions = action.payload.missions.map(convertPhaseToUTC);
+
+      // If no mission was loaded or if the previously selected mission was removed,
+      // automatically select first available mission from the updated list
+      const isSelectedMissionLoaded = state.data.missions.some(
+        (mission) => mission.identifier === state.selectedMissionIdentifier
+      );
+      if (!isSelectedMissionLoaded) {
+        state.selectedMissionIdentifier = action.payload.missions[0].identifier;
```
WeirdRubberDuck, line 56: What happens if there are no missions... ?  E.g. if all the mission assets were deinitialized

WeirdRubberDuck (approved):
Nice!  Eventually, we should consider adding a popup or something when a new missing file is loaded, so the user becomes aware of the change. But this is already a great improvement :) 

I saw Ylva's comment about potentially having the dropdown at the top at all files. Personally, I would prefer not to, since loading multiple missions files could still quite an unusual/special situation, and it takes unnecessary screen real-estate

src/panels/MissionsPanel/hooks.ts
```diff
@@ -39,3 +42,31 @@ export function useJumpToTime(fadeTime: number = 1): ReturnType {
...
+  const { isInitialized, data, selectedMissionIdentifier } = useAppSelector(
+    (state) => state.missions
+  );
+
+  let mission: Phase | undefined = undefined;
+
+  // If a new mission is added, the missions order might have changed so we try to find
+  // the last one viewed
+  if (isInitialized && selectedMissionIdentifier) {
+    const selectedMission = data.missions.find(
```
WeirdRubberDuck, line 59: This speaks to using a map instead, IMO. I would rather add some complexity in the redux store than have to search the list of missions. 

And it would be nice to avoid having to use zero-indexing to get the missing if there is only one

src/panels/MissionsPanel/hooks.ts
```diff
@@ -39,3 +42,31 @@ export function useJumpToTime(fadeTime: number = 1): ReturnType {
...
+      (mission) => mission.identifier === selectedMissionIdentifier
+    );
+    if (selectedMission) {
+      mission = selectedMission;
+    } else {
+      mission = data.missions.at(0);
+    }
```
WeirdRubberDuck, line 66:
```suggestion
    mission = selectedMission ?? data.missions.at(0);
```
?

WeirdRubberDuck: Can we end up here if there are no missions?  Getting and index zero seems a bit dangerous if there's no check against if `data.missions` have enough values

engbergandreas:
>     * Should the identifier be the name - rename "identifier" to "name"? I think we have special rules for identifiers (such as no spaces etc). Here you could also create an object with keys instead of the array as suggested above

So this is kind of weird, but the name that we get is in fact the identifier, see `MissionManager::loadMission`. So it should probably be changed to  `for (auto const& [identifier, mission] : missions) .. ` instead? Also, the phase already reserves a key "name" that is the readable name of the mission. 

Changing the topic to send it as a map as suggested though.

WeirdRubberDuck:
Nice with the change to map 👍

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -1,15 +1,33 @@
...
+
+  function onMissionSelected(identifier: string | null) {
+    if (!identifier) {
+      return;
+    }
+    dispatch(setSelectedMission(identifier));
+    luaApi?.setCurrentMission(identifier);
+  }
+
+  if (!hasMission || !mission) {
```
ylvaselling, line 30: When there are no missions, `Object.values(missions)` creates an empty array, and we then take the first element in that empty array (going out of bounds). This is when mission is undefined on the next line. Although we don't get a warning or error for this, going out of bounds on an array should be avoided, so how about something like this instead? (with next comment)

```suggestion
  const luaApi = useOpenSpaceApi();
  const dispatch = useAppDispatch();

  const hasMission = isInitialized && Object.values(missions).length > 0;

  function onMissionSelected(identifier: string | null) {
    if (!identifier) {
      return;
    }
    dispatch(setSelectedMission(identifier));
    luaApi?.setCurrentMission(identifier);
  }

  if (!hasMission) {
```

src/panels/MissionsPanel/MissionsPanel.tsx
```diff
@@ -23,5 +41,22 @@ export function MissionsPanel() {
...
+          })}
+          value={selectedMissionIdentifier}
+          onChange={onMissionSelected}
+          my={'xs'}
+          allowDeselect={false}
+        />
+      )}
+      <MissionContent missionOverview={mission} />
+    </Box>
+  );
```
ylvaselling, line 61:
```suggestion
const mission = missions[selectedMissionIdentifier] ?? Object.values(missions)[0];
  return (
    <Box>
      {Object.keys(missions).length > 1 && (
        <Select
          label={'Selected mission'}
          placeholder={'Select a mission'}
          data={Object.entries(missions).map(([identifier, mission]) => {
            return { value: identifier, label: mission.name };
          })}
          value={selectedMissionIdentifier}
          onChange={onMissionSelected}
          my={'xs'}
          allowDeselect={false}
        />
      )}
      <MissionContent missionOverview={mission} />
    </Box>
  );
```

src/redux/events/eventsMiddleware.ts
```diff
@@ -42,6 +44,12 @@ export const setupEventsSubscription = createAsyncThunk(
         case 'ActionRemoved':
           thunkAPI.dispatch(removeAction(data.Uri));
           break;
+        case 'MissionAdded':
+          thunkAPI.dispatch(refreshMissions());
+          break;
+        case 'MissionRemoved':
+          thunkAPI.dispatch(refreshMissions());
+          break;
```
ylvaselling, line 52: These two cases could be one

```suggestion
case: "MissionAdded":
case: "MissionRemoved":
...
```

ylvaselling (changes requested):
Overall looks great! Just found one out of bounds in an array otherwise 👍

PR #77, opened 2025-03-04
Make button header a button so we can focus it

ylvaselling:


engbergandreas:
I'm not entirely sure about this solution, now you have to tab twice to get away from the tab header, first the outer and then the inner. Feels weird.. autofocus seems to be a normal css prop so maybe we can add it to rc-dock button directly? 


https://github.com/user-attachments/assets/568f5fca-5a66-4ee7-99df-c3acf678e2a6

WeirdRubberDuck:
Yeah, seems like this problem might require an alternative solution. I'm also not quite convinced that this is the way

PR #78, opened 2025-03-04
Create interface that extends TreeNodeData and add the guipath to it

ylvaselling:
* Now we can search for subfolders

The drawback of this is that we get a lot more search results... I'm not really sure what I think about it. One one hand, it is nice, because we can easier find things, but it also is weird that when you search for "Earth", it shows up second. On the other hand, this behavior could be fixed with a "matching score", which we might want to implement anyway - it is not certain that we display things in a good order anyway.

One option is to only search on the leaf folder, but that seems a bit weird to me. 

Do we also want to be able to search for folder in the navigation panel?

What do you guys think?

### DIFF
--- a/src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
+++ b/src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
@@ -1,4 +1,4 @@
-import { Tree, TreeNodeData } from '@mantine/core';
+import { Tree } from '@mantine/core';
 
 import { useGetStringPropertyValue } from '@/api/hooks';
 import { useAppSelector } from '@/redux/hooks';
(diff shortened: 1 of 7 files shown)


WeirdRubberDuck:
Yeah as you may have noticed, I'm personally a bit skeptical to introducing more matches, and deliberately did not include the GUI path when implementing this search the first time. 

If it's not handled "gracefully", e.g. with a sorting that makes sense, I think it makes the search harder to use rather than being helpful

engbergandreas:
>  but it also is weird that when you search for "Earth", it shows up second. 

To be fair, our current search also puts Earth second, not that it should be used as the ground truth or anything
image

I think with some search organization and a better matcher, it will be practical. I like the idea of searching for Earth and getting all the satellites, for example. But then maybe the search result should display it as Earth/Satellites > Geostationary, GPS etc. Just something to make it clear where it's coming from and why it's shown. 

> One option is to only search on the leaf folder, but that seems a bit weird to me.

Yeah, this feels weird, depending on what the search results would look like

> Do we also want to be able to search for folder in the navigation panel?
Maybe? Same as above, we need to change how we display the search results first

engbergandreas (approved):

PR #80, opened 2025-03-10
Feature/reorder taskbar

engbergandreas:
Added reordering to the `view menu` so that you can customize the taskbar. This change also included making the `MenuItemsData` into a map instead of an array. I'm not a big fan of having to retype the component ID as its key, but I also don't want to remove the `componentID` from the interface so 🤷 don't know what else to do here

Edit: It was a mistake to bring the other commits to this PR, so disregard them, I'll make a separate PR for those. Just look at the last commit `add ability to customize task bar order`


ylvaselling (approved):
Works like a charm! 🎉

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -38,7 +35,7 @@ export function TaskBar({ visibleMenuItems }: Props) {
               borderTopRightRadius: isLastItem ? 'var(--mantine-radius-md)' : undefined
             }}
           >
-            <TaskBarMenuButton item={item} />
+            <TaskBarMenuButton item={item} disabled={!itemConfig.enabled} />
```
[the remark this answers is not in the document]
engbergandreas: Oh right, so if you disable a module or if a mission inte loaded, the Button will be disabled. I'm not sure if we want it that way though.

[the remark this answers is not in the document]
ylvaselling: Aha, yeah, not sure either. There seems to be many ways now that panels can be disabled. Maybe it requires some discussion

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -44,18 +33,28 @@ export function ViewMenu({ visibleMenuItems, setVisibleMenuItems }: Props) {
...
+              <Menu.Item
+                key={item.componentID}
+                leftSection={item.renderIcon?.(IconSize.xs)}
+                rightSection={<CheckboxIndicator checked={itemConfig.visible} />}
+                onClick={() => setMenuItemVisible(itemConfig.id, !itemConfig.visible)}
+              >
+                {item.title}
+              </Menu.Item>
+            );
+          }}
```
WeirdRubberDuck, line 54: It looks a bit busy and weirdly aligned with the dragHandle on the left side. How about moving it to the right? And maybe also move the checkbox to the very left, as requested by someone in the team during the user test? 

This would help with parsing each line (i..e is X enabled), and makes the reordering seem more like a secondary action than the primary one 

image


Code suggestion to make it look like this (with some prop reordering) below

```suggestion
          id={'viewMenu'}
          data={menuItems}
          dragHandlePosition={'right'}
          onDragEnd={({ updatedData }) => {
            dispatch(updateMenuItemsOrder(updatedData));
          }}
          renderFunc={(itemConfig) => {
            const item = menuItemsData[itemConfig.id];
            return (
              <Menu.Item
                key={item.componentID}
                leftSection={
                  <Group>
                    <CheckboxIndicator checked={itemConfig.visible} />
                    {item.renderIcon?.(IconSize.xs)}
                  </Group>
                }
                onClick={() => setMenuItemVisible(itemConfig.id, !itemConfig.visible)}
              >
                {item.title}
              </Menu.Item>
            );
          }}
```

WeirdRubberDuck: For reference, this is how it looks like now: 
image

The drag handle being the first thing per line makes it harder to "parse" when identifying if a certain panel is disabled/enabled

src/panels/Menu/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface TaskbarItemConfig {
+  id: string;
+  visible: boolean;
```
WeirdRubberDuck, line 3: Add comments decsribing what "visible" and "enabled" means? So that future us remember the difference?

src/panels/OriginPanel/AnchorAimView/AnchorAimListEntry.tsx
```diff
@@ -24,10 +24,11 @@ export function AnchorAimListEntry({
   isCurrentAnchor,
   isCurrentAim,
   onSelectAnchor,
-  onSelectAim
+  onSelectAim,
+  ...props
```
WeirdRubberDuck, line 28:
```suggestion
  ...styleProps
```
?  To make it clearer what these props contain

src/panels/OriginPanel/AnchorAimView/AnchorAimView.tsx
```diff
@@ -151,7 +152,7 @@ export function AnchorAimView({ favorites, searchableNodes, matcherFunction }: P
         )}
         matcherFunc={matcherFunction}
       >
-        <FilterList.SearchResults.VirtualList />
+        <FilterList.SearchResults.VirtualList gap={3} />
```
WeirdRubberDuck, line 155: I'm a bit sad about this custom gap in both the focus view and here. As well as the added `mb`

3 is a new number 🥲 how would it look with 2?

WeirdRubberDuck (commented):
Reordering works great!  Had a comment about changing the order for the elements in the menu

And this PR should maybe be cleaned off any commits that do not belong to the reordering. Let me know if you want help with that

src/redux/local/localSlice.ts
```diff
@@ -27,11 +33,40 @@ export const localSlice = createSlice({
...
+      state,
+      action: PayloadAction<{ id: string; visible: boolean }>
+    ) => {
+      const item = state.taskbarItems.find((item) => item.id === action.payload.id);
+      if (item) {
+        item.visible = action.payload.visible;
+      }
+      return state;
+    },
+    setEnabledMenuItem: (
```
WeirdRubberDuck, line 47: tiny thing.  Maybe this should be `setMenuItemEnabled` and `setMenuItemVisible`?

`

PR #82, opened 2025-03-12
Feature/scroll component

ylvaselling:
* Adds new component ScrollBox with custom styling
* Using this for all scroll in the ui, including topmenubar and taskbar


src/components/ScrollBox/ScrollBox.tsx
```diff
@@ -0,0 +1,32 @@
...
+      ref={ref}
+      style={{
+        overflowX: horizontal ? 'auto' : 'hidden',
+        overflowY: vertical ? 'auto' : 'hidden'
+      }}
+      className={styles.scroller}
+      onWheel={(event) => {
+        if (horizontal) {
+          event.currentTarget.scrollLeft += event.deltaY;
+        }
```
engbergandreas, line 25: I don't think we have a panel that scrolls horizontally and vertically right now, but this behaviour will be quite strange if a panel has both. Cause then you would end up in the bottom right corner when scrolling down

engbergandreas (commented):
I know you don't like firefox, but this is breaking a lil to hard I think 😭  I think also we need to increase the size somewhat for firefox, looks great in OpenSpace though!

https://github.com/user-attachments/assets/8c4ee1e3-c8cd-4e12-a8ef-887ec0377423

src/components/ScrollBox/ScrollBox.module.css
```diff
@@ -0,0 +1,22 @@
+.scroller::-webkit-scrollbar {
```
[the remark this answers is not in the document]
WeirdRubberDuck: That is correct. Firefox uses separate classes

src/components/ScrollBox/ScrollBox.module.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.scroller:hover::-webkit-scrollbar {
+  background-color: var(--mantine-color-gray-9);
+}
+
+.scroller:hover::-webkit-scrollbar-thumb {
+  border-radius: 100px;
+  background-color: var(--mantine-color-gray-2);
+}
```
WeirdRubberDuck, line 22: Some wishes here: 

1. Use `mantine-color-dark` instead of `mantine-color-gray` ?  Don't know if it's just me, but I think the cold gray looks quite ugly against the otherwise warmer gray from the "dark" colors
2. Please sort the CSS so that the thumb-related styles ar next to each other. E.g:
```suggestion

.scroller:hover::-webkit-scrollbar {
  background-color: var(--mantine-color-gray-9);
}

.scroller::-webkit-scrollbar-thumb {
  border-radius: 100px;
  background-color: var(--mantine-color-gray-5);
  border: 3px solid transparent;
  background-clip: content-box;
}

.scroller:hover::-webkit-scrollbar-thumb {
  border-radius: 100px;
  background-color: var(--mantine-color-gray-2);
}
```
3. Make the thumb a little darker? It's current very bright!  Could maybe be solved by using e.g. `var(--mantine-color-dark-2)` for the default background and  `var(--mantine-color-dark-1)` for hover
4. Now I've only seen the sliders on the darkest background and there is looks great, but how will they look on e.g. a `Paper` or `Card` ? Do we need to think about that? E.g. what should the `hovercolor` for the track be? Now I tried ` var(--mantine-color-dark-7)`, but maybe it should really be a transparent bright color..?  (seems a bit dangerous in a 50 shades of gray territory, though....)

WeirdRubberDuck (commented):
She's so beautiful! 🤩

Had some comments about the colors. I would personally prefer if we used mantine's "dark" color instead of the cold gray, and if the slider thumb was made to be a little less white. With the current colors it sort of becomes the main focus in any view, which is not great IMO

Here's a version where I've reduced the brightness slightly and used the warmer grays of "dark" instead . 

https://github.com/user-attachments/assets/46d87a5a-eaa4-4af3-b6e9-34753fd08415

Thoughts? Do you agree? Also see the comment in the CSS file.

src/components/ScrollBox/ScrollBox.module.css
ylvaselling: This is a Box component so I don't think having it on a paper or a card will be an issue. We could just implement the styling for that when we need it?

WeirdRubberDuck: I will need it :) 

image

WeirdRubberDuck: But yeah, we could merge it first and then look at how it looks in this case

engbergandreas (approved):

src/components/ScrollBox/ScrollBox.module.css
```diff
@@ -0,0 +1,24 @@
...
+    border-radius: 100px;
+  }
+
+  .scroller:hover::-webkit-scrollbar {
+    background-color: var(--mantine-color-dark-8);
+  }
+
+  .scroller::-webkit-scrollbar-thumb {
+    border-radius: 100px;
+    background-color: var(--mantine-color-dark-2);
```
WeirdRubberDuck, line 15:
```suggestion
    background-color: var(--mantine-color-dark-3);
```
? :) I still think it "pops" a bit much against the other elements

WeirdRubberDuck (approved):

PR #83, opened 2025-03-12
Skybrowser close button

ylvaselling:
* Adds close button to skybrowser tab
* Changes the styling of the skybrowser tab 
    * This has the benefit of showing the color in a neater way, and not looking confusingly similar to the window tabs

### DIFF
--- a/src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
+++ b/src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
@@ -1,9 +1,9 @@
-import { Button, ColorInput, Stack, Title } from '@mantine/core';
+import { ColorInput, Stack, Title } from '@mantine/core';
 
 import { useOpenSpaceApi } from '@/api/hooks';
 import { useAppSelector } from '@/redux/hooks';
 
-import { useBrowserColorString, useSkyBrowserIds } from '../hooks';
+import { useBrowserColorString } from '../hooks';
 
 interface Props {
   id: string | undefined;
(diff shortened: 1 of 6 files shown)


src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -24,31 +23,18 @@ export function Settings({ id }: Props) {
...
-      const otherBrowsers = browserIds.filter((b) => b !== id);
-      luaApi?.skybrowser.setSelectedBrowser(otherBrowsers[0]);
-    }
-    luaApi?.skybrowser.removeTargetBrowserPair(id);
-  }
-
   return (
     <Stack gap={5} my={'lg'}>
       <Title order={2}>Settings</Title>
+      {/* Just using this variable here so the linter is happy */}
```
WeirdRubberDuck, line 29: Why is this comment needed? :) I would suggest to remove it

WeirdRubberDuck (approved):
Looks great to me!

One thought: Do we want a confirm modal to pop up when deleting the browser?  Or would that be annoying?

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
ylvaselling: It's to make it easier to merge when i merge the skybrowser settings pr

ylvaselling:
> Looks great to me!
> 
> One thought: Do we want a confirm modal to pop up when deleting the browser? Or would that be annoying?

I don't think it's necessary with a modal tbh, we can see if the user test requests one?

WeirdRubberDuck:
> > Looks great to me!
> > One thought: Do we want a confirm modal to pop up when deleting the browser? Or would that be annoying?
> 
> I don't think it's necessary with a modal tbh, we can see if the user test requests one?

Sounds good.

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
WeirdRubberDuck: Aha, so the idea is to remove it on merge of that one?

engbergandreas (approved):

PR #85, opened 2025-03-12
Feature/open close panels

engbergandreas:
close panels using shift click or right click, on touch it should work to press and hold for ~ half a second to close. I have only tested it using Chrome dev tools so would be great if anyone can test on an actual touch screen, otherwise I'll do it when I'm back


src/panels/Menu/hooks.ts
```diff
@@ -2,7 +2,10 @@ import { useCallback, useEffect } from 'react';
 
 import { useGetBoolPropertyValue } from '@/api/hooks';
 import { useAppDispatch, useAppSelector } from '@/redux/hooks';
-import { setMenuItemEnabled, setMenuItemVisible } from '@/redux/local/localSlice';
+import {
+  setMenuItemEnabled,
+  setMenuItemVisible as setMenuItemVisibleRedux
```
ylvaselling, line 7: Why do we need the Redux here? all actions that are dispatched go to redux

ylvaselling (approved):

src/panels/Menu/hooks.ts
WeirdRubberDuck: There is a naming conflict with the action `setMenuItemVisible` and a function that is being returned by the `useMenuItems` hook

src/panels/Menu/hooks.ts
```diff
@@ -41,7 +44,7 @@ export function useMenuItems() {
   }, [isSkyBrowserEnabled, enablePanel]);
 
   function setMenuItemVisible(id: string, value: boolean) {
```
WeirdRubberDuck, line 46: See comment thread above, but I suggest that we remove this function entirely and just use the `dispatch` in the code

Then there is no need for the renaming of the function

WeirdRubberDuck (commented):
Works great for mouse - but yeah it would be good to try for touch.  I kind of expected click+hold to work on mouse as well but that it doesn't is no big deal, IMO

Did we want to add some styling to the menuitems that are currently open? E.g. like a border at the bottom of the button?

engbergandreas:
> Works great for mouse - but yeah it would be good to try for touch. I kind of expected click+hold to work on mouse as well but that it doesn't is no big deal, IMO

I'm not sure I have the same expectation tbh since we support right-click for that anyway, and a long press using a mouse is not a common thing you see... Another reason to try and avoid is that it will add a bunch of IMO complex logic to see whether the user is holding the mouse button or not. 


> Did we want to add some styling to the menuitems that are currently open? E.g. like a border at the bottom of the button?

Yes, thinking I will add that as a separate thing instead of this PR

PR #86, opened 2025-03-13
Make highlight of items be inside the object

ylvaselling:
* Fixes the focus issue in the bottom bar

### DIFF
--- a/src/theme/mantineTheme.ts
+++ b/src/theme/mantineTheme.ts
@@ -12,6 +12,7 @@ import {
 } from '@mantine/core';
 
 import styles from './mantineTheme.module.css';
+
 /**
  * This functions resolves colors for different variants, and adds new variants, for
  * Mantine components.
--- a/src/theme/mantineTheme.ts
+++ b/src/theme/mantineTheme.ts
@@ -11,6 +11,7 @@ import {
   VariantColorsResolver
 } from '@mantine/core';
 
+import styles from './mantineTheme.module.css';
 /**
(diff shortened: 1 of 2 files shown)


src/theme/mantineTheme.ts
```diff
@@ -9,6 +9,7 @@ import {
   VariantColorsResolver
 } from '@mantine/core';
 
+import styles from './mantineTheme.module.css';
 /**
```
WeirdRubberDuck, line 13:
```suggestion

/**
```

WeirdRubberDuck (approved):
❤️ Nice

Interesting that the Mantine `Tree` component is not included in the focus styling from the class. However, I don't consider that a problem. We want to add some CSS to the tree anyways (like hover behaviors) so can try to figure that out separately

engbergandreas:
I'm thinking if maybe we should do something special for the taskbar and let everything else be default mantine behavior?
For example, here I'm focused on `Next`... Otherwise we have to come up with a completely new focus color.
image

ylvaselling:
> I'm thinking if maybe we should do something special for the taskbar and let everything else be default mantine behavior? For example, here I'm focused on `Next`... Otherwise we have to come up with a completely new focus color. image

Good catch... maybe simple solution is just to use white? Or the text highlight color

WeirdRubberDuck (approved):
👍 I like how it looks in white. 

There are still a couple of places that get a blue focus indicator (like the friction buttons in the top right), but this is good enough to merge IMO

PR #88, opened 2025-03-13
Add skybrowser settings

ylvaselling:
* Adds skybrowser settings
    * Note that it doesn't look great because of that the properties are not in master 
* Removes "AAS" from worldwidetelescope and removed their logo - replaced it with an icon like in missions panel

Needs testing 🙏


src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+          additionalData={{
+            MinimumValue: [-10, -10, -10],
+            MaximumValue: [10, 10, 10],
+            SteppingValue: [1, 1, 1],
+            Exponent: 1
+          }}
+          viewOptions={{}}
+        />
+      </Collapsable>
+      <Group w={'100%'}>
```
WeirdRubberDuck, line 46:
```suggestion
      <Group grow>
```

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <Button
+          onClick={() => {
+            luaApi?.skybrowser.addDisplayCopy(id, noOfCopies, newPosition);
+          }}
+          flex={1}
```
WeirdRubberDuck, line 51:
```suggestion
```
solved by replacing 100% width with grow

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+        >
+          Add
+        </Button>
+        <Button
+          onClick={() => {
+            luaApi?.skybrowser.removeDisplayCopy(id);
+          }}
+          flex={1}
```
WeirdRubberDuck, line 59:
```suggestion
```
solved by replacing 100% width with grow

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+  const displayCopies = useAppSelector(
+    (state) => state.skybrowser.browsers[id]?.displayCopies
+  );
+
+  return (
+    <>
+      <Property uri={`ScreenSpace.${id}.Scale`} />
+      <Property uri={`ScreenSpace.${id}.FaceCamera`} />
+      <Property uri={`ScreenSpace.${id}.UseRadiusAzimuthElevation`} />
+      <Collapsable title={'Add New Display Copy Settings'}>
```
WeirdRubberDuck, line 27: Add some padding/margin at the bottom

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,114 @@
...
+        min={-90}
+        max={90}
+        onEnter={setDeclination}
+      />
+      <Property uri={`Scene.${targetId}.Renderable.ApplyRoll`} />
+      <Property uri={`ScreenSpace.${id}.PointSpacecraft`} />
+      <Collapsable title={'Display Copies'}>
+        <SettingsDisplayCopies id={id} />
+      </Collapsable>
+      <Collapsable title={'General Settings'}>
```
WeirdRubberDuck, line 102: We need to add some hierarchy for the different sections here. First, use titles for the collapsables. 

See other code suggestions for ordering - these will be based on that this change is done
```suggestion
      <Collapsable title={<Title order={3}>Display Copies</Title>}>
        <SettingsDisplayCopies id={id} />
      </Collapsable>
      <Collapsable title={<Title order={3}>General Settings</Title>}>
```

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+        <Button
+          onClick={() => {
+            luaApi?.skybrowser.removeDisplayCopy(id);
+          }}
+          flex={1}
+        >
+          Remove
+        </Button>
+      </Group>
+      <Title order={2} mt={'md'} mb={'sm'}>
```
WeirdRubberDuck, line 64:
```suggestion
      <Title order={4} mt={'md'} mb={'sm'}>
```
Note that the "Settings" label is order 2, and we want this to be below that. We also want it to be below the suggested title I added for the Collapsable

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
WeirdRubberDuck, line 27:
```suggestion
      <Collapsable title={<Title order={4}>Add New Display Copy Settings</Title>}>
```
following the other suggested hierarchy

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,75 @@
...
+        Added Display Copies
+      </Title>
+      {Object.entries(displayCopies).map(([key, entry]) => (
+        <Fragment key={key}>
+          <Property uri={`ScreenSpace.${id}.${entry.idShowProperty}`} />
+          <Property uri={`ScreenSpace.${id}.${key}`} />
+        </Fragment>
```
WeirdRubberDuck, line 71: How about grouping these in some component that makes it clear that they belong together, visually?  The current flat list is quite hard to parse, even when grouped in a common parent element. 

image

It was much clearer when each were on one single line, like in the old UI. Maybe we want to make something custom for these instead of using the `Property` components?
image

In general, I feel like the interface for the display copies may need some discussion for it to be useful

Also, I never thought I'd miss the old sliders, but for the display copies there were so much more interesting to use 🥲 Need to come up with a good way of editing vector properties..

WeirdRubberDuck (changes requested):
Seems to work just fine, but I had some comments

1. Do we maybe want to use the logo they have on their website? 

image

2. When setting the border radius, the are outside the radius turns white. I don't think we want this behavior. It will for example be problematic when used in an otherwise dark dome setup

image

3. These inputs also have the issue of getting long "ugly" values from OpenSpace. You can however ignore this for now - I've added a feature to fix it in the scene tree branch I hop to get to merging that branch tomorrow or so
image

4. This settings does nothing anymore? 
image

5. This menu is in great need of some titles and text hierarchy, both to be understandable by screen readers and by the visible eye. I added some code suggestions

6. The end of the settings menu gets quite busy and hard to read. Part of it is due to the doubly displayed properties, but we should either way maybe create some type of visual grouping for the different menus. Let's discuss?

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,114 @@
...
         label={'Color'}
         placeholder={'Set browser color...'}
         format={'rgb'}
         defaultValue={color}
         onChange={setColor}
       />
-      <Button variant={'outline'} color={'red'} onClick={deleteBrowser} mt={'lg'}>
-        Delete browser
-      </Button>
+      <NumericInput
```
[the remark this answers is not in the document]
ylvaselling: 👍 I think it would be nice to have the same look as a numeric property, is that possible?

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
ylvaselling: I agree with you, however I think these properties have the same problem that many of our other properties have, which is the "enabled property" thing, doesn't really make sense to me to only implement a custom thing in only one place. We should implement the "enabled property" and then switch to that.

These properties need to be fixed anyway, currently they switch between cartesian and spherical coordinate systems, and there has been some discussion that people want both, so that needs another interface anyway

ylvaselling: Potentially we could place the enabled property as the header in a collapse? That could be a quick fix. Still it would differ from how the UI looks in other places

ylvaselling:
> Seems to work just fine, but I had some comments
> 
>     1. Do we maybe want to use the logo they have on their website?
> 
> 
> image
> 
>     2. When setting the border radius, the are outside the radius turns white. I don't think we want this behavior. It will for example be problematic when used in an otherwise dark dome setup
> 
> 
> image
> 
>     3. These inputs also have the issue of getting long "ugly" values from OpenSpace. You can however ignore this for now - I've added a feature to fix it in the scene tree branch I hop to get to merging that branch tomorrow or so
>        image
> 
>     4. This settings does nothing anymore?
>        image
> 
>     5. This menu is in great need of some titles and text hierarchy, both to be understandable by screen readers and by the visible eye. I added some code suggestions
> 
>     6. The end of the settings menu gets quite busy and hard to read. Part of it is due to the doubly displayed properties, but we should either way maybe create some type of visual grouping for the different menus. Let's discuss?
> 
>     7. It also looks like this setting still needs to be implemented:
>        image



> Seems to work just fine, but I had some comments
> 
>     1. Do we maybe want to use the logo they have on their website?
> 
> 
> image
> 
>     2. When setting the border radius, the are outside the radius turns white. I don't think we want this behavior. It will for example be problematic when used in an otherwise dark dome setup
> 
> 
> image
> 
>     3. These inputs also have the issue of getting long "ugly" values from OpenSpace. You can however ignore this for now - I've added a feature to fix it in the scene tree branch I hop to get to merging that branch tomorrow or so
>        image
> 
>     4. This settings does nothing anymore?
>        image
> 
>     5. This menu is in great need of some titles and text hierarchy, both to be understandable by screen readers and by the visible eye. I added some code suggestions
> 
>     6. The end of the settings menu gets quite busy and hard to read. Part of it is due to the doubly displayed properties, but we should either way maybe create some type of visual grouping for the different menus. Let's discuss?
> 
>     7. It also looks like this setting still needs to be implemented:
>        image

1. Don't really know if that logo is valid - just went by what Megan said, to remove the logo. Maybe check with her if we can use the logo you're suggesting?
2. Yeah, making a ticket for it in notion. Not part of this pr - think it needs some attention on the server
3. -
4. ✅ True, I guess we can just remove that property in the engine.
5. ✅ 
6. Sure, yeah it was a bit hard to tell with the double properties so I just left it as it was in the old ui
7. True! making a ticket for it

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
WeirdRubberDuck: As long as we "visually group" thw two properties for each display copies somehow, I think the current solution is good enough

WeirdRubberDuck: Love the new styling!  Resolving this comment

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,119 @@
...
+      <Property uri={`ScreenSpace.${id}.Scale`} />
+      <Property uri={`ScreenSpace.${id}.FaceCamera`} />
+      <Property uri={`ScreenSpace.${id}.UseRadiusAzimuthElevation`} />
+
+      <Collapsable
+        title={
+          <>
+            Added Display Copies{' '}
+            <Badge variant={'default'}>{Object.keys(displayCopies).length}</Badge>
+          </>
```
WeirdRubberDuck, line 106:
```suggestion
          <Group>
            Added Display Copies
            <Badge variant={'default'}>{Object.keys(displayCopies).length}</Badge>
          </Group>
```

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,119 @@
...
+                  Number of copies
+                </Text>
+              </InputLabel>
+              <InfoBox>
+                {`This sets the position of the first copy. The additional copies will be evenly
+          spread out on the Azimuth, if isUsingRadiusAzimuthElevation is enabled,
+          otherwise it will spread on the Y axis.`}
+              </InfoBox>
+            </Group>
+            <NumericInput value={noOfCopies} onEnter={setNoOfCopies} />
```
WeirdRubberDuck, line 63: 1. The infobox text here is wrong
2. Is this setting even needed?? It's so easy to just click the "Add" button two times

WeirdRubberDuck: But I guess it is since people want to automatically spread out multiple copies at once?  I personally find this feature quite confusing. Do we still need it? If so, the settings menu might need a bit more explaining of what the purpose of these settings are

Let's discuss, maybe?

engbergandreas (commented):
1. Do we want to have min/max values shown for the Border Radius setting - including the slider? 
2. Setting the Declination to 90 or -90 will change the Right Ascension as well. To reproduce: set Right Ascension to 30, change Declination to 90 => Right ascension will be set to ~170

src/components/SettingsPopout/SettingsPopout.tsx
```diff
@@ -1,17 +1,30 @@
...
+  popoverWidth?: number;
+  title?: string;
+}
+
+export function SettingsPopout({
+  popoverWidth,
+  title,
+  children,
+  ...styleProps
+}: Props & PropsWithChildren & MantineStyleProps) {
```
WeirdRubberDuck, line 16: I'm not necessarily against this way of doing this, but please let's stick to the way we've done it in all other components
```suggestion
interface Props extends MantineStyleProps, PropsWithChildren {
  popoverWidth?: number;
  title?: string;
}

export function SettingsPopout({ popoverWidth, title, children, ...styleProps }: Props) {
```

WeirdRubberDuck (changes requested):
Loving the new look!  Have some thing sin the code I would like fixed, though :) 

Also, can we please fix this issue with the border radius: 
https://github.com/user-attachments/assets/dada2e0a-59fc-4c68-ab8f-c9e4c6f5d7fc



There is also an issue with the RA changing when the declination is edited, but I assume this might be harder to fix?
https://github.com/user-attachments/assets/ea478d4f-3e14-4309-a5bf-e7c093295571

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,193 @@
...
+            <NumericPropertySlider
+              disabled={false}
+              value={fov}
+              min={0.01}
+              max={70.0}
+              step={1}
+              onInput={setVerticalFov}
+              flex={2}
+            />
+            <NumericInput
```
WeirdRubberDuck, line 122: I suggest you set a ` decimalScale={x}` on these inputs.  Maybe limiting it to 3-4 decimals?

WeirdRubberDuck: Goes for RA and Dec as well

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,193 @@
...
+              clampBehavior={'strict'}
+              onEnter={setVerticalFov}
+              flex={1}
+            />
+          </Group>
+
+          <Text size={'sm'} mb={'xs'} mt={'sm'} fw={'500'}>
+            Right Ascension
+          </Text>
+          <Group grow my={'sm'}>
```
WeirdRubberDuck, line 135:
```suggestion
          <Group grow my={'xs'}>
```

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,193 @@
...
+              step={1}
+              onInput={setDeclination}
+            />
+            <NumericInput
+              value={dec}
+              min={-90}
+              max={90}
+              clampBehavior={'strict'}
+              onEnter={setDeclination}
+              pb={'sm'}
```
WeirdRubberDuck, line 172:
```suggestion
```

engbergandreas (approved):

src/panels/SkyBrowserPanel/BrowserTabs/Settings.tsx
```diff
@@ -1,54 +1,193 @@
...
+            label={`Settings for Display Copies, which are a way to render the WorldWide
+                Telescope view as a ScreenSpaceRenderable.`}
+          >
+            <Tabs.Tab value={'Display'}>Display Copies</Tabs.Tab>
+          </Tooltip>
+          <Tooltip label={`General settings for all browsers.`}>
+            <Tabs.Tab value={'General'}>General</Tabs.Tab>
+          </Tooltip>
+        </Tabs.List>
+        <Tabs.Panel value={'Settings'} py={'sm'}>
```
WeirdRubberDuck, line 92: Now you add a `py={'sm'}` to all panels individually. This easily becomes out of sync

How about, instead, wrapping all the `Tabs.Panel` components in a box or something that adds that padding to all of them?

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,119 @@
...
+                </Text>
+              </InputLabel>
+              <InfoBox>
+                {`This sets the position of the first copy. The additional copies will be evenly
+          spread out on the Azimuth, if isUsingRadiusAzimuthElevation is enabled,
+          otherwise it will spread on the Y axis.`}
+              </InfoBox>
+            </Group>
+            <Box pb={'sm'}>
+              <DefaultView
```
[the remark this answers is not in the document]
ylvaselling: Actually I would like all our Properties to just be wrappers of input components :) I think that would be awesome as that could unify the look of our UI

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
ylvaselling: What is wrong about the text?

I'm just making the ui look like before 😄 I've had similar thoughts about this setting but discussed removing it with alex and we ended up having it the way it was.

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
```diff
@@ -0,0 +1,119 @@
...
+          }}
+          leftSection={<PlusIcon />}
+          flex={1}
+        >
+          Add
+        </Button>
+        <Button
+          onClick={() => {
+            luaApi?.skybrowser.removeDisplayCopy(id);
+          }}
```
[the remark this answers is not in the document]
ylvaselling: Technically these are not the same though. The lambda function will return the value, whereas the curly braces will not return anything. In this particular case I don't think it makes a difference since the functions return void but having some global setting for this would be dangerous I think

ylvaselling:
> Loving the new look! Have some thing sin the code I would like fixed, though :)
> 
> Also, can we please fix this issue with the border radius: https://github.com/user-attachments/assets/d3545147-2f17-4edf-9ae1-cce02fc0edd4
> 
> There is also an issue with the RA changing when the declination is edited, but I assume this might be harder to fix? https://github.com/user-attachments/assets/9c034b30-6e02-4e1b-9dc8-cefc706cae7f

The links you posted lead to 404 so not sure what you mean :)

ylvaselling:
The declination to -90 is a bug in the engine that was present in the old ui as well, so I don't think that is in the scope of this pr.

src/panels/SkyBrowserPanel/WorldWideTelescope/WorldWideTelescopeView.tsx
```diff
@@ -112,7 +115,11 @@ export function WorldWideTelescopeView() {
         allowFullScreen
         height={height}
         width={width}
-        style={{ borderWidth: 0, pointerEvents: 'none' }}
+        style={{
+          border: '0px solid transparent',
+          borderRadius: radius * 1000,
```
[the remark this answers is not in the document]
WeirdRubberDuck: AH, this might be what's causing the weird behavior I sent a video for

WeirdRubberDuck:
> > Loving the new look! Have some thing sin the code I would like fixed, though :)
> > Also, can we please fix this issue with the border radius: https://github.com/user-attachments/assets/d3545147-2f17-4edf-9ae1-cce02fc0edd4
> > There is also an issue with the RA changing when the declination is edited, but I assume this might be harder to fix? https://github.com/user-attachments/assets/9c034b30-6e02-4e1b-9dc8-cefc706cae7f
> 
> The links you posted lead to 404 so not sure what you mean :)

Don't know what happened with the videos there... Updated so the first video should show now. Something broke when I uploaded both in one message

Here's the second one, showing how changing declination leads to some odd (? are they?) RA changes. But as I said, not sure if this is something we can fix? 
https://github.com/user-attachments/assets/1a0f593a-4f65-46fb-bc12-245692376326

src/panels/SkyBrowserPanel/WorldWideTelescope/WorldWideTelescopeView.tsx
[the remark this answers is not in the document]
ylvaselling: Yeah for some reason the background of the transparent iframe in our new ui is white. In our previous ui it was transparent. I tried to remedy it this hacky way but it doesn't work very well... the proper solution is to fix this behavior on the server which is another repo. I think for now I'll leave the background white and do a new ticket for the radius

ylvaselling:
> > > Loving the new look! Have some thing sin the code I would like fixed, though :)
> > > Also, can we please fix this issue with the border radius: https://github.com/user-attachments/assets/d3545147-2f17-4edf-9ae1-cce02fc0edd4
> > > There is also an issue with the RA changing when the declination is edited, but I assume this might be harder to fix? https://github.com/user-attachments/assets/9c034b30-6e02-4e1b-9dc8-cefc706cae7f
> > 
> > 
> > The links you posted lead to 404 so not sure what you mean :)
> 
> Don't know what happened with the videos there... Updated so the first video should show now. Something broke when I uploaded both in one message
> 
> Here's the second one, showing how changing declination leads to some odd (? are they?) RA changes. But as I said, not sure if this is something we can fix?
> 2025-03-26.16-39-52.mov

See above comment:)

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
WeirdRubberDuck: Ok, sure, if you've already discussed it leave it like it is. Was only a suggestion :) 

But the text for the infobox is a copy of the one that describes the "Position of the first copy". This should describe the "Number of copies" setting 😉

ylvaselling: Ah, good catch. 

I'm happy make to discuss ways how to make this interface better, but I think there are plenty of things in the skybrowser that can be improved and I would like to merge this PR, so maybe we can do an overhaul another time :)

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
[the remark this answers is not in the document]
WeirdRubberDuck: Yeah that would definitely be something to strive for. Maybe we can do another rewrite at some point

Have started poking a bit on the precision issue and had a similar thought

src/panels/SkyBrowserPanel/BrowserTabs/SettingsDisplayCopies.tsx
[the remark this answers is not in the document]
WeirdRubberDuck: Ok sure, I agree with your point on the global setting. 

This is, however, how we have written essentially all of our onclick function call, and I think we should be consistent. It is also a lot cleaner looking :)

WeirdRubberDuck (approved):
Looks great with the new changes!  Love the tabs in the settings. 

One thing - it would be good if the settings view is loading while the rest of the skybrowser menus are still loading. Now the settings are accessible before they are really working

PR #89, opened 2025-03-13
Add hover behaviors for collapsable headers

WeirdRubberDuck:
Please let me know what you think


ylvaselling:
Looks nice! Add the same hover to the property owner headers?

WeirdRubberDuck:
> Looks nice! Add the same hover to the property owner headers?

It's already applied to those as well :) They use the same collapsable component

engbergandreas (approved):
didnt realise how much this would be missing, nice

PR #90, opened 2025-03-17
Create a way to save a file that stores the bottom bar info

ylvaselling:
* Creates a way to store information for the bottom bar
* Similarly to how this is done, we could save the layout of the windows as well. It would be pretty much the exact same code
* Alternatives:
    * We could make one directory where all the layout files must go and then walk that directory and display the files. Pro: Browser agnostic. Con: less flexible 

TODO:
* Somehow we need to communicate the default file name per profile to the UI. This is not solved yet. Thoughts / discussion? 
    * To use the default layout you need to create a PATH in the config called USER_UI_TASKBAR, and place a file in that folder that is called "layout.json". Obviously that is not what we want long term, but I think using a path in the config is kind of neat as we don't need any properties. We still need the file name somehow...
    
Edit: works on firefox, but the downloading will not open a file dialog, just download to "Downloads" folder


WeirdRubberDuck (commented):
Tried it out a little bit, but took a pause now.  One major question:  Shouldn't the save/load from the UI also include the order?  Or what are the numbers in the .json file for? Anyhow, now it doesn't

src/redux/profile/profileMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { setMenuItemVisible } from '../local/localSlice';
+
+export interface ProfileState {
+  uiPanelVisibility: {
+    // The key here should match the id of the menu item,
+    // else will be ignored
+    [key: string]: boolean;
+  };
+  markNodes: string[];
```
WeirdRubberDuck, line 15: Just realized - how about adding the name of the started profile to this topic as well?  

It would be nice to add that information somewhere in the UI - to answer the question of "Oh, I wonder which profile I have running"

src/redux/profile/profileMiddleware.ts
```diff
@@ -0,0 +1,42 @@
+import { createAsyncThunk } from '@reduxjs/toolkit';
+
+import { api } from '@/api/api';
+import { onOpenConnection } from '@/redux/connection/connectionSlice';
+import { AppStartListening } from '@/redux/listenerMiddleware';
+
+import { setMenuItemVisible } from '../local/localSlice';
+
+export interface ProfileState {
```
WeirdRubberDuck, line 9: I just realized that there is only a middleware for this. Why is that? 

It would match the the other redux code more if this data structure was located in a file `profileSlice.ts`

ylvaselling: We don't need to store the profile data because the data is actually handled by other slices so this middleware is really only for getting the topic. It is possible that we might want to add things in the future that we want to store but I think we can add that when needed

ylvaselling:
> Tried it out a little bit, but took a pause now. One major question: Shouldn't the save/load from the UI also include the order? Or what are the numbers in the .json file for? Anyhow, now it doesn't

Good catch, added a fix for that now

src/redux/profile/profileMiddleware.ts
ylvaselling: Making a ticket in notion

src/panels/Menu/hooks.ts
```diff
@@ -49,3 +53,51 @@ export function useMenuItems() {
...
+    if (!content || Object.keys(content).length === 0) {
+      console.error('File is empty');
+      return;
+    }
+    const newLayout = Object.values(content) as TaskbarItemConfig[];
+    if (newLayout.length !== menuItems.length) {
+      console.error('Invalid layout file. Length does not match');
+      return;
+    }
+    // We have to ensure that all id's are valid before we can set
```
WeirdRubberDuck, line 75:
```suggestion
    // We have to ensure that all ids are valid before we can set
```

src/panels/Menu/types.ts
```diff
@@ -1,5 +1,6 @@
 export interface TaskbarItemConfig {
   id: string;
+  name: string; // The name of the menu item. This will show up in the OpenSpace launcher.
```
WeirdRubberDuck, line 3: Move this comment to on top of the variable. A change like that is incoming for the other variables as well :)

src/util/fileIOhooks.ts
```diff
@@ -0,0 +1,87 @@
...
+    try {
+      const content = await contents?.[0]?.text();
+      if (!content) {
+        return;
+      }
+      const json = JSON.parse(content);
+      handlePickedFile(json);
+    } catch (e) {
+      // TODO: do we want to throw here?
+      console.error('Error parsing file', e);
```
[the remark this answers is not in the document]
WeirdRubberDuck: Maybe for now, add TODO comments for all the console.logs that you have?  To make this part of some notification system in the future?

And I think personally, I would prefer to `throw` here. Seems like it makes no difference, but the solution would probably involve dealing with that throw somehow

And console logs leads to annoying warnings when linting :)

src/util/fileIOhooks.ts
```diff
@@ -0,0 +1,85 @@
...
+          description: 'Text Files',
+          accept: {
+            'text/plain': ['.json']
+          }
+        }
+      ]
+    };
+    // For some reason typescript doesn't recognize this as a function
+    // It is an experimental feature of the chromium browser
+    const fileHandle = await window.showSaveFilePicker(options);
```
WeirdRubberDuck, line 51: This is sad :( but yeah, looks like there's nothing we can do

At least the linter does not complain

WeirdRubberDuck (approved):

src/util/fileIOhooks.ts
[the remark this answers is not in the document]
ylvaselling: A throw will lead to the ErrorBoundary showing though, whereas a console.error won't disrupt the user experience, so I don't think a throw is an option in this case.

I'll add some todo comments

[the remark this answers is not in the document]
WeirdRubberDuck: I tried, but didn't get any error visible for the user in this case 🤔 Anyhow, since we want to deal with it later anyways, it doesn't really matter :)

src/redux/profile/profileMiddleware.ts
```diff
@@ -0,0 +1,42 @@
...
+      listenerApi.dispatch(getProfile());
+    }
+  });
+
+  startListening({
+    actionCreator: getProfile.fulfilled,
+    effect: async (action, listenerApi) => {
+      Object.entries(action.payload.uiPanelVisibility).forEach(([key, value]) => {
+        listenerApi.dispatch(setMenuItemVisible({ id: key, visible: value }));
+      });
```
engbergandreas, line 38: Do we want to do something explicitly here if the ID from OpenSpace is invalid? It will be silently ignored in `setMenuItemVisible`, so if a user has changed the json file there is no way to tell unless they later try to load that file using the load/save buttons

src/util/fileIOhooks.ts
engbergandreas, line 51: This needs to be wrapped in a try/catch if the users cancels the window dialog an error is thrown. Don't think we need to handle it in anyway just catch it and continue with our day

PR #91, opened 2025-03-17
Feature/properties refactor

ylvaselling:
Refactor the properties

Aim:
* Make everything strongly typed -  no `any`
* Minimize prop drilling
* Minimize prop spreading

Bigger changes:
* Make each property handle their own state
* Utilize renderFunction match the patterns in the rest of the UI
* Removed some wrapper components

Lmk what you think!


src/components/Property/Types/OptionProperty.tsx
```diff
@@ -1,58 +1,44 @@
...
-  // select component
-  function valueToString(value: number): string | undefined {
-    return Object.keys(options).find((key) => options[key] === value);
+  if (!description || value === undefined) {
+    return <></>;
   }
+  const data: Option[] = description.additionalData.Options;
+
+  // Get the name of the option, e.g. "Option 1"
+  const optionsStrings = data.map((option) => Object.values(option)[0]);
```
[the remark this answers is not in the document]
WeirdRubberDuck: Also, I would like to keep some kind of TODO comment here.  The `[0]` part is a bit nasty and we should at some point refactor so that the value for the options is a nice list.  

It should really just be a list of strings :)

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -1,58 +1,44 @@
...
 
   function onChange(option: string | null) {
-    if (option) {
-      setPropertyValue(options[option]);
+    if (option && optionsStrings.indexOf(option) !== -1) {
+      // Now we need to find the number key of the option
+      // which is the same as its index in the optionsStrings array
+      const index = optionsStrings.indexOf(option);
```
WeirdRubberDuck, line 29:
```suggestion
    if (option && options.indexOf(option) !== -1) {
      // Now we need to find the number key of the option
      // which is the same as its index in the options array
      const index = options.indexOf(option);
```
PLease 😇

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -1,58 +1,44 @@
...
     <Select
       aria-label={`${name} option input`}
       placeholder={'Choose an option'}
-      disabled={disabled}
-      data={Object.keys(options)}
-      value={valueToString(value)}
-      onChange={(_value) => onChange(_value)}
+      disabled={description.additionalData.isReadOnly}
+      data={optionsStrings}
+      value={optionsStrings[value]}
```
WeirdRubberDuck, line 40:
```suggestion
      data={options}
      value={options[value]}
```

src/components/Property/Types/SelectionProperty.tsx
```diff
@@ -1,30 +1,38 @@
...
+    string[] | undefined
+  >(value);
+
+  const description = useGetPropertyDescription(uri);
+
+  if (!value || !description || currentValue === undefined) {
+    return <></>;
+  }
+
+  const options = description.additionalData.Options as string[];
```
WeirdRubberDuck, line 22: Same here, make a type for this `additionalData` structure? Similarly to `AdditionalDataNumber` and `AdditionalDataVectorMatrix`?

src/components/Property/Types/SelectionProperty.tsx
```diff
@@ -1,30 +1,38 @@
...
-  setPropertyValue,
-  value,
-  additionalData
-}: Props) {
-  const options = additionalData.Options;
   return (
     <MultiSelect
       aria-label={`${name} multi-select`}
-      disabled={disabled}
+      disabled={description.additionalData.isReadOnly}
```
WeirdRubberDuck, line 32: Same here!  It's really an issue that this does not lead to an error. We should look at the type handling for `additionalData`

```suggestion
      disabled={description.metaData.isReadOnly}
```

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -1,36 +1,101 @@
...
+            const newValue = [rgbaColor.r / 255, rgbaColor.g / 255, rgbaColor.b / 255];
+            if (hasAlpha) {
+              newValue.push(rgbaColor.a);
+            }
+            // @TODO (emmbr26, 2025-03-14) Check for min max values. color values should be
+            // between 0 and 1.
+            setPropertyValue(newValue);
+          }}
+        />
+      </Flex>
```
WeirdRubberDuck, line 74: Why was the component for this removed? To me it seems much more clear to have each view option in a separate component

Now it's a bit confusing that there is a component in the viewoptions folder for MinMaxRange, but not for color

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -1,36 +1,101 @@
...
+    );
   }
 
   if (viewOptions.MinMaxRange) {
-    return <MinMaxRange {...props} />;
+    if (value.length !== 2) {
+      throw Error('Invalid use of MinMaxRange view option!');
+    }
```
WeirdRubberDuck, line 81: Move this check back into the `MinMaxRange` component?

src/components/Property/Types/MatrixProperty.tsx
```diff
@@ -1,32 +1,37 @@
...
-  const step = additionalData.SteppingValue;
+  if (!description || !value) {
+    return <></>;
+  }
+
+  const { additionalData } = description as {
+    additionalData: AdditionalDataVectorMatrix;
+  };
+  const { MinimumValue: min, MaximumValue: max, SteppingValue: step } = additionalData;
+  const disabled = description.metaData.isReadOnly;
```
WeirdRubberDuck, line 28: There is a bit of inconsistency as to how this setting is handled. Sometimes it's called readonly, and sometimes disabled. I personally think that's a bit messy/confusing. 

Also, I would like to discuss this.  Could we maybe move this setting into the general `PropertyProps` and set it at the very top level?  That would get rid of inconsistency issues, and make it clearer that all the property type have to deal with these settings. 

As part of the rewrite I did, I tried to move as many of the common props as possible to the top level, to get rid of inconsistency issues like this.  That work has sort of been undone with this rewrite.

I don't want us to end up with the same mess as before, where we add some setting to properties and forget to handle it for some of the property types. Let's discuss, please

src/components/Property/types.ts
```diff
@@ -1,6 +1,18 @@
 // These props are the same for all property types
-export interface ConcretePropertyBaseProps {
-  name: string;
-  description: string;
-  disabled: boolean;
+export interface PropertyProps {
```
WeirdRubberDuck, line 2: Is this name too generic?  I.e. is there a risk that we mix it up with the types in the top level types file?

src/components/Property/types.ts
```diff
@@ -1,6 +1,18 @@
...
+  MinimumValue: number;
+  SteppingValue: number;
+}
+
+export interface AdditionalDataVectorMatrix {
+  Exponent: number; // TODO: handle the exponent
+  MaximumValue: number[];
+  MinimumValue: number[];
+  SteppingValue: number[];
 }
```
WeirdRubberDuck, line 18: Add all the other AdditionalData types as well! 

However, I'm wondering if this file is needed or if it would make sense to have the additionaldata struct in the file for the specific property type?

src/components/Property/types.ts
ylvaselling: Any suggestions for another name?

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -1,58 +1,44 @@
 import { Select } from '@mantine/core';
 
-import { ConcretePropertyBaseProps } from '../types';
+import { useGetOptionPropertyValue, useGetPropertyDescription } from '@/api/hooks';
+import { PropertyProps } from '@/components/Property/types';
 
+// In OpenSpace the options are represented like so:
+// { 0 : "Option 1"}, { 1: "Option 2"}
```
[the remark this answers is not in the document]
ylvaselling: Actually the array starts at 0 if you print it to the console

src/components/Property/Types/StringProperty.tsx
```diff
@@ -1,23 +1,23 @@
...
     return (
       <Paper px={'sm'} py={5}>
         <Text size={'sm'}>{value}</Text>
       </Paper>
     );
   }
-  return (
-    <StringInput onEnter={setPropertyValue} value={value} aria-label={`${name} input`} />
-  );
+  return <StringInput onEnter={setValue} value={value} aria-label={`${name} input`} />;
```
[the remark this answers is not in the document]
ylvaselling: Now i got to remove the useDescription hook 😢 but i can take it back for the aria label...

src/components/Property/Types/OptionProperty.tsx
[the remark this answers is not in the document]
WeirdRubberDuck: Ok, ignore this then :)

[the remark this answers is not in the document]
WeirdRubberDuck: Ok 👍 
```suggestion
// { 0: "Option 1"}, { 1: "Option 2"}
```

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -1,58 +1,51 @@
...
+  if (!description || value === undefined) {
+    return <></>;
   }
 
+  const { Options: data } = description.additionalData as AdditionalDataOptions;
+
+  // Get the name of the option, e.g. "Option 1"
+  // Flatten the array as otherwise each element is an array
+  // @TODO (ylvse 2025-03-18): Change the property format
+  const options = data.map((option) => Object.values(option)).flat();
```
WeirdRubberDuck, line 30: Nice solution 👍

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
```diff
@@ -1,23 +1,38 @@
...
+export function ColorView({
+  readOnly,
+  setPropertyValue,
+  value,
+  additionalData,
+  isInt
+}: Props) {
   if ((value.length !== 3 && value.length !== 4) || isInt) {
-    throw Error('Invalid use of Color view option!');
+    throw Error('Invalid use of Color View option!');
```
WeirdRubberDuck, line 22:
```suggestion
    throw Error('Invalid use of Color view option!');
```

src/api/hooks.ts
```diff
@@ -22,7 +22,13 @@ import {
 } from '@/redux/sessionrecording/sessionRecordingMiddleware';
 import { subscribeToTime, unsubscribeToTime } from '@/redux/time/timeMiddleware';
 import { ConnectionStatus } from '@/types/enums';
-import { Property, PropertyOwner, PropertyValue, Uri } from '@/types/types';
+import {
+  Property,
```
WeirdRubberDuck, line 26:
```suggestion
```

or maybe just do a lint-fix

src/components/Property/types.ts
```diff
@@ -1,6 +1,19 @@
...
-export interface ConcretePropertyBaseProps {
-  name: string;
-  description: string;
-  disabled: boolean;
+export interface PropertyProps {
+  uri: string;
+  readOnly: boolean;
+}
+
+export interface AdditionalDataNumber {
```
WeirdRubberDuck, line 7: Move this to NumericProperty.tsx?

src/components/Property/types.ts
```diff
@@ -1,6 +1,19 @@
...
+  Exponent: number;
+  MaximumValue: number;
+  MinimumValue: number;
+  SteppingValue: number;
+}
+
+export interface AdditionalDataVectorMatrix {
```
WeirdRubberDuck, line 14: Move this to VectorProperty.tsx and export it from there?

src/components/Property/types.ts
WeirdRubberDuck: Nope, and feel free to leave it as is

WeirdRubberDuck (approved):
Thanks for addressing the comments! This looks good enough to me :)

src/components/Property/types.ts
engbergandreas: We shouldn't export interfaces from .tsx files though, so I would rather see us move all the other `AdditionalData` types here instead

src/types/types.ts
```diff
@@ -41,10 +45,14 @@ export interface PropertyMetaData {
   needsConfirmation: boolean;
 }
 
+export type AdditionalData =
+  | AdditionalDataNumber
+  | AdditionalDataVectorMatrix
+  | AdditionalDataOptions
+  | AdditionalDataSelection;
+
```
engbergandreas, line 53: Would probably move this to the Property types file instead and import it here

engbergandreas (commented):
looks good

src/components/Property/PropertyLabel.tsx
```diff
@@ -1,21 +1,26 @@
...
-  tip?: string;
-  isReadOnly: boolean;
-}
+import { PropertyProps } from './types';
+
+export function PropertyLabel({ uri }: PropertyProps) {
+  const description = useGetPropertyDescription(uri);
+
+  if (!description) {
+    return <></>;
```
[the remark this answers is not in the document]
ylvaselling: I think this mainly is for when the interface is loading, before we have received all the data from openspace, so this will be undefined all the time, therefore it is not really an error

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -1,42 +1,58 @@
...
+  'ULongProperty',
+  'IntProperty',
+  'UIntProperty'
+];
+
+export interface AdditionalDataNumber {
+  Exponent: number;
+  MaximumValue: number;
+  MinimumValue: number;
+  SteppingValue: number;
```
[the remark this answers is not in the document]
ylvaselling: Here you and @WeirdRubberDuck have to fight it out 😉

src/components/Property/Types/NumericProperty/NumericProperty.tsx
```diff
@@ -1,42 +1,58 @@
...
+const propertyTypes = [
+  'FloatProperty',
+  'DoubleProperty',
+  'ShortProperty',
+  'UShortProperty',
+  'LongProperty',
+  'ULongProperty',
+  'IntProperty',
+  'UIntProperty'
+];
```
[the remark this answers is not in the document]
ylvaselling: Needs discussion for how to implement this 🤔

engbergandreas (approved):
Looks good to me, I did some tidying with the properties hooks. Felt there were too many empty lines

WeirdRubberDuck (approved):

PR #93, opened 2025-03-17
Increase infobox tooltip text width

WeirdRubberDuck:
Set the line width to correspond to roughly 50 character per line => good readability.

Seems to look ok even for shorter texts

Before: 
image

After:
image

Shorter text:
image

### DIFF
--- a/src/components/InfoBox/InfoBox.tsx
+++ b/src/components/InfoBox/InfoBox.tsx
@@ -7,7 +7,7 @@ interface Props {
   w?: number;
 }
 
-export function InfoBox({ text, w = 300 }: Props) {
+export function InfoBox({ text, w = 320 }: Props) {
   return (
     <Tooltip label={text} multiline maw={w} offset={{ mainAxis: 5, crossAxis: 100 }}>
       <ThemeIcon radius={'xl'} size={'xs'}>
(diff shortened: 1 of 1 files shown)


src/components/InfoBox/InfoBox.tsx
```diff
@@ -7,7 +7,7 @@ interface Props {
   w?: number;
 }
 
-export function InfoBox({ text, w = 220 }: Props) {
+export function InfoBox({ text, w = 320 }: Props) {
   return (
     <Tooltip label={text} multiline maw={w} offset={{ mainAxis: 5, crossAxis: 100 }}>
```
ylvaselling, line 12: Why use maw and not w? Seems they use w in the offical docs https://mantine.dev/core/tooltip/#multiline

ylvaselling (approved):

src/components/InfoBox/InfoBox.tsx
WeirdRubberDuck: To look nicer for short texts. 

With `maw`,  i.e. max width:
image

VS with `w`, i.e. a fixed width
image

engbergandreas (approved):
nice

PR #94, opened 2025-03-19
Feature/scene refactor

ylvaselling:
Aim
* Compartmentalize complexity
* Use existing components

Bigger changes
* Moved scene tree data structure from redux store to hook with all it's utility functions inside that hook file. This was because the tree was recomputed in the component anyway, so it didn't seem necessary to store it in redux. This way only the hook file needs to "see" the utility functions and we have a clean interface to get the tree data structure.
* Lifted state up from scene filters as the state was duplicated
* Changed the expanded state from being computed to use mantines built in functionality
* Moved usePropertyVisibilty hook to api/hooks where it seemed more at home
* Removed leftSection part of the SceneGraphNodeHeader as it was only used in one place and if I recall correctly, was meant to solve the issue of line breaking text in the GeoPositionPanel. It seemed to add too much functionality to the SceneGraphNodeHeader imo so I instead solved the line breaking issue with the text and removed the leftSection. This would've caused an issue if we both wanted to show the checkbox and a left section so that issue is also now fixed
* ...and some smaller cleanup / refactor

Let me know what you think!


src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
```diff
@@ -33,29 +30,20 @@ export function SceneGraphNodeView({ uri }: Props) {
...
-  const translation = getSgnTransform(
-    propertyOwner,
-    TransformType.Translation,
-    propertyOwners
-  );
-  const rotation = getSgnTransform(propertyOwner, TransformType.Rotation, propertyOwners);
-
-  // Group the transforms under one tab, in the following order. Only show the transforms
-  // that are actually present
-  const transforms = [scale, translation, rotation].filter((t) => t !== undefined);
```
WeirdRubberDuck, line 49: Note that the order here was intentional, but maybe alphabetical makes more sense

My impression is that users most commonly fiddle with the scale propertyowner, which is why it was displayed first

WeirdRubberDuck: I personally think it's worth it, since they're open per default :)

Another alternative would be to use the order in which the property owners are added, i.e. not sort them at all. This _should_ match the order in which we want them presented in the UI (at least it does for all other subproperty owners), and if it doesn't we can add it on the C++ side

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -62,15 +55,19 @@ export function SceneTreeFilters({ onFilterChange }: Props) {
             <Checkbox
               label={'Show only visible'}
               checked={showOnlyVisible}
-              onChange={(event) => setshowOnlyVisible(event.currentTarget.checked)}
+              onChange={(event) =>
+                setSpecificFilter('showOnlyVisible', event.currentTarget.checked)
+              }
```
WeirdRubberDuck, line 60: Hm ,not a fan of the need for a string key here. How about simply removing the `setSpecificFilter` function and just doing: 
```suggestion
              onChange={(event) =>
                setFilter({ ...filter, showOnlyVisible: event.currentTarget.checked })
              }
```

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -62,15 +55,19 @@ export function SceneTreeFilters({ onFilterChange }: Props) {
...
             />
             <InfoBox text={'Visible = Enabled and not faded out'} />
           </Group>
           <Group>
             <Checkbox
               label={'Show objects with GUI hidden flag'}
               checked={showHiddenNodes}
-              onChange={(event) => setShowHiddenNodes(event.currentTarget.checked)}
+              onChange={(event) =>
+                setSpecificFilter('showHiddenNodes', event.currentTarget.checked)
```
WeirdRubberDuck, line 69: Same here
```suggestion
                setFilter({ ...filter, showHiddenNodes: event.currentTarget.checked })
```

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -83,7 +80,7 @@ export function SceneTreeFilters({ onFilterChange }: Props) {
           <MultiSelect
             data={sortedTags}
             value={selectedTags}
-            onChange={setSelectedTags}
+            onChange={(newTags) => setSpecificFilter('tags', newTags)}
```
WeirdRubberDuck, line 83: aand here
```suggestion
            onChange={(newTags) => setFilter({ ...filter, tags: newTags })}
```

src/panels/Scene/SceneTree/hooks.ts
```diff
@@ -1,24 +1,96 @@
...
+  isGroupNode,
+  SceneTreeGroupPrefixKey,
+  treeDataForSceneGraphNode
+} from './treeUtils';
+import { SceneTreeFilterSettings, SceneTreeNodeData } from './types';
+
+/**
+ * Create the data for the Scene tree from the groups information.
+ */
+function sceneTreeDataFromGroups(
```
WeirdRubberDuck, line 27: I think this is the first time we add a large number of helper functions in a hooks.ts file?  Is this something we want to do?  

It gets quite hard to find the hook in the file, but it's also nice that it's all located in one place.... And does it still make sense when/if we add more hooks to this file?

No strong opinion, just thoughts

src/panels/Scene/SceneTree/treeUtils.ts
```diff
@@ -0,0 +1,27 @@
+import { SceneTreeNodeData } from './types';
+
+// This key is added to the group path values in the tree data structure, to mark which
+// tree nodes correspond to groups (as opposed to scene graph nodes).
+export const SceneTreeGroupPrefixKey = '/groups/';
+/**
```
WeirdRubberDuck, line 6:
```suggestion

/**
```

src/util/groupsHelpers.ts
```diff
@@ -0,0 +1,51 @@
+import { Group, Groups, Properties, PropertyOwners } from '@/types/types';
```
WeirdRubberDuck, line 1: Do we really want this file in the main src folder?  It will only ever be relevant to the groups slice.  Maybe put it in the `src\redux\groups? folder?

WeirdRubberDuck: We have types files for some of these folders, so I don't see a problem of also adding files with helper functions

src/util/propertyTreeHelpers.ts
```diff
@@ -108,6 +108,12 @@ export function isRenderable(uri: Uri): boolean {
   return uri.endsWith('.Renderable');
 }
 
+export function isTransform(uri: Uri): boolean {
+  return (
+    uri.endsWith(ScaleKey) || uri.endsWith(TranslationKey) || uri.endsWith(RotationKey)
```
WeirdRubberDuck, line 113: I would prefer to make this check safer.  "Scale" is, for example, a common ending for many property owners that are not necessarily a transform. 

Could for example check that the URI includes two dots, i.e. that these keys are on the third level in the URI

WeirdRubberDuck (approved):
Very nice! 

Life is so much better when one understands how to use hooks and `useMemo`s :D Glad you took the time to clean up this code of the Emma that did not know better

Had a couple of thoughts/comments but nothing major. Looks great

src/util/propertyTreeHelpers.ts
WeirdRubberDuck: oculd maybe also update the funciton name to `isSgnTransform` if we want to make it even more clear, as well.

src/panels/Scene/SceneTree/hooks.ts
ylvaselling: Yeah I wasn't sure about this either, but the reason I still went ahead and did it was because it seemed like compartmentalizing these functions had two benefits:

1. We know for sure no other file is dependent on these functions as they aren't exported
2. We know exactly what code the hook is dependent on

If we ever want to modify the tree data structure this hook it will make it pretty straightforward for the above reasons (no other code will need the util functions so they will not be changed for other reasons than that the hook changes).

Since the work done by this hook is quite complex, it felt nice to compartmentalize it like this to me 🤔 I understand what you mean though, but I'm wondering if this is not a pretty nice structure to use. Makes dependencies easy 🤔 

We could place the hook at the top of the file if the linter allows it. It should be able to see the functions with capturing.

If we want to add more hooks to this file, we might want to consider starting naming our hooks files? And then we can have a folder called "hooks"

src/panels/OriginPanel/FocusView/FocusView.tsx
```diff
@@ -25,17 +24,15 @@ interface Props {
 }
 
 export function FocusView({ favorites, searchableNodes, matcherFunction }: Props) {
-  const propertyOwners = useAppSelector((state) => state.propertyOwners.propertyOwners);
   const engineMode = useSubscribeToEngineMode();
 
-  const [anchor] = useGetStringPropertyValue(NavigationAnchorKey);
+  const anchorNode = useGetAnchorNode();
```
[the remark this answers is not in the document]
ylvaselling: We do need it in this file though, as we pass the anchorNode to the FocusEntry

src/panels/Scene/SceneGraphNode/SceneGraphNodeMoreMenu.tsx
```diff
@@ -23,7 +19,7 @@ interface Props {
 
 export function SceneGraphNodeMoreMenu({ uri }: Props) {
   const propertyOwner = useGetPropertyOwner(uri);
-  const [anchor] = useGetStringPropertyValue(NavigationAnchorKey);
+  const anchorNode = useGetAnchorNode();
```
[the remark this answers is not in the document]
ylvaselling: Yeah here I changed it for readabililty. I could change it back though, don't have strong opinions on this one

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
ylvaselling: Lol tried it and the order is literally the reverse of what we want. We could display it in reverse hehe?
image

src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
```diff
@@ -1,61 +1,61 @@
...
+      )
     });
   }
 
   return (
-    <Tree
-      data={featuredTreeData}
-      renderNode={(payload) => <SceneTreeNode {...payload} />}
-    />
+    <>
```
engbergandreas, line 47:
```suggestion
```

src/panels/Scene/SceneTree/FeaturedSceneTree.tsx
```diff
@@ -1,61 +1,61 @@
...
+        renderNode={({ node, expanded, elementProps, tree }) => (
+          <SceneTreeNode
+            node={node}
+            expanded={expanded}
+            elementProps={elementProps}
+            tree={tree}
+          />
+        )}
+      />
+    </>
```
engbergandreas, line 59:
```suggestion
```

src/panels/Scene/SceneTree/hooks.ts
engbergandreas: I think for the above reason this is ok, but we should move the hook to the top if possible

engbergandreas (commented):
no real comments, great rewrite!

engbergandreas (approved):

PR #96, opened 2025-03-21
Feature/properties min max

ylvaselling:
Adds error display when property numbers are outside of range.

Might need some discussion if this is the behavior we want.

Built on top of properties refactor so only last commit is interesting


src/components/Input/NumericInput/NumericInput.tsx
```diff
@@ -126,6 +135,7 @@ export function NumericInput({
           />
         )
       }
+      error={isOutsideRange ? `Value outside range [${min}, ${max}]` : undefined}
```
WeirdRubberDuck, line 138: how about only showing the error message when editing?  IIRC that's how our last UI worked at least, but it might be worth discussing if that's what we really want. 
```suggestion
      error={isEditing && isOutsideRange ? `Value outside range [${min}, ${max}]` : undefined}
```

WeirdRubberDuck (commented):
I like it!  Also a very clean solution to an otherwise complex problem :) 

See one comment - we should discuss if we want the error message to always be visible, or only visible when editing. 

And if we think that it looks too "serious" with the red, versus what was previously shown in yellow. 

For simplicity, I would prefer to keep it red, but IIRC there were some comments when I implemented the feature in the last UI so we should maybe check with the team.

src/components/Input/NumericInput/NumericInput.tsx
engbergandreas: I think that would be good, if I've knowingly set a value outside the range a red warning text would be quite annoying to see all the time. I could however see that we add some kind of exclamation or something in yellow/orange to indicate which ones have values outside

ylvaselling: Tried something lmk what you think

src/components/Input/NumericInput/NumericInput.tsx
```diff
@@ -125,6 +129,14 @@ export function NumericInput({
       max={max}
       step={step}
       hideControls={hideControls}
+      leftSection={
+        isOutsideRange &&
```
WeirdRubberDuck, line 133: I love this!   :D Looks great. 

Could we maybe add a `Tooltip` that says something like "The set value is outside of the min/max range"?  So that one can hover over the icon to see what it means?

It does add to an already existing issue of there not being enough space for Vector properties, but we should look at improving those anyways and since this is a feature that most people shouln't see, I don't think it's necessarily a problem

image

src/components/Input/NumericInput/NumericInput.tsx
```diff
@@ -116,6 +129,14 @@ export function NumericInput({
...
+        !isEditing && (
+          <ThemeIcon color={'orange.4'} variant={"transparent"}>
+            <MdWarning size={IconSize.sm} />
```
WeirdRubberDuck, line 136: Yeah I think making this smaller would be nice
```suggestion
            <MdWarning size={IconSize.xs} />
```

WeirdRubberDuck (approved):

engbergandreas:
Looks great!

PR #97, opened 2025-03-21
Feature/focusable info icon

ylvaselling:
Make info tooltip a button for accessibility

Might need some discussion w regards to if this is what we want


engbergandreas (approved):

WeirdRubberDuck:
Not sure how to review this PR, as it both needs to get master emerged in and is also the base for another PR

PR #98, opened 2025-03-21
Feature/make top menu focusable

ylvaselling:
Makes the top menu focusable (the user visibility).

To discuss: should we do something clever with the now repeated code in the option property or leave it as is?
We could make some kind of helper function to put the repeated code in. Maybe even in the hook for the option property. But that will likely hide the problem which is that the property itself is in kind of a weird format (the additional data).

Built on top of the property refactor rewrite so only last commit is interesting


src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -13,9 +14,32 @@ import { TopBarMenuWrapper } from '../TopBarMenuWrapper';
...
+  const options = data.map((option) => Object.values(option)).flat();
+
+  function onChange(option: string | null) {
+    if (option && options.indexOf(option) !== -1) {
+      // Now we need to find the number key of the option
+      // which is the same as its index in the optionsStrings array
+      const index = options.indexOf(option);
+      setPropertyVisibility(index);
+    }
+  }
```
engbergandreas, line 41: This is off by one, the map from OpenSpace is: 

0: Object { 1: "Novice User" }
1: Object { 2: "User" }
2: Object { 3: "Advanced User" }
3: Object { 4: "Developer" }
4: Object { 5: "Everything" }

But with the `options` we've mapped it 0-4 instead so Everything -> Developer, Developer -> Advanced User which also leads to this error when setting Novice User 
(E) OptionProperty       Could not find an option for value '0'
(W) Logging              Number of Errors: 1

engbergandreas: Would also put the index on line 38 outside the if so we don't do the same `option.indexOf(..)`

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -62,15 +86,34 @@ export function ViewMenu() {
...
+        {options.map((option, i) => (
+          <Menu.Item
+            key={option}
+            leftSection={
+              <Group>
+                <CheckboxIndicator checked={i === propertyVisibility} />
+              </Group>
+            }
+            onClick={() => onChange(option)}
+          >
```
engbergandreas, line 112: Should this be Radio Buttons instead? I was thinking if there is any difference for a screen reader, otherwise I think it looks good so no biggie

engbergandreas: looks good

engbergandreas (approved):

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -13,11 +14,19 @@ import { TopBarMenuWrapper } from '../TopBarMenuWrapper';
...
+  const [propertyVisibility, setPropertyVisibility] = useOptionProperty(
+    'OpenSpaceEngine.PropertyVisibility'
+  );
+  const description = usePropertyDescription('OpenSpaceEngine.PropertyVisibility');
   const dispatch = useAppDispatch();
 
+  if (!description) {
+    return <></>;
+  }
+  const { Options: data } = description.additionalData as AdditionalDataOptions;
```
WeirdRubberDuck, line 26: This component inslcudes more than just the user level, so please give the list a more specific name. For example
```suggestion
  const { Options: userLevelOptions } = description.additionalData as AdditionalDataOptions;
```

Also, maybe we can add a hook that gets the list of options from an option property? Then we can remove the duplicated part and `description` etc which isn't really relevant for this component

WeirdRubberDuck (changes requested):
Getting this error when trying to open the visibility settings, so can't try it :(  Want to see how it looks like with the Radio input.  

image

Also, are we sure that we want to hide this setting behind an extra click?  Might make it harder for the user to see which setting they currently have on. It would be good if that was easily accessible

Also, since we're handling the option property explicitly here, maybe we could use the Lua API to set the value in this case? Then we could use the string value, since that is now supported 😄 

Also added a suggestion for what to do about the "duplicated option property code"

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
engbergandreas: > Also, are we sure that we want to hide this setting behind an extra click? Might make it harder for the user to see which setting they currently have on. It would be good if that was easily accessible

I think its fine that it's behind an extra click, it's not like this setting is changed a lot and I think most people are not gonna fiddle with it anyway

WeirdRubberDuck (commented):
I was instructed that I needed the latest commit in the engine for it to work :)  Please remember to specify that in the PR next time

Looks good to me, but I noticed that it's not possible to change the option using the keyboard? We should maybe look at fixing that

WeirdRubberDuck:
Oh, and the message of this PR should include that this includes a breaking change of the format for the option property data! As well as a link to the commit in the OpenSpace repo that changes it

This might be useful when investigating changes in the future, so please add it

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -13,28 +13,16 @@ export function OptionProperty({ uri, readOnly }: PropertyProps) {
...
     <Select
       aria-label={`${description.name} option input`}
       placeholder={'Choose an option'}
       disabled={readOnly}
-      data={options}
-      value={options[value]}
-      onChange={onChange}
+      data={Object.entries(data).map(([value, label]) => {
+        return { value: value, label: label };
+      })}
```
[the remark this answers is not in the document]
WeirdRubberDuck: Actually, I looked a bit at the change in the engine. Why do we return an object with the values as keys, and not just a list of strings?  While we're changing the format, can't we just return a list of strings?  (with the caveat that it will still be an object with the indexes as keys)

A drawback of having the value as a key is that the sorting will be off if we have more than 10 options, since the order in the object will be alphabetical based on the keys.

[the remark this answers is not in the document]
ylvaselling: I didn't want to rewrite the entire property since maybe there was a reason it was implemented the way it was. Seems to me it would be fine to rewrite it as a list of strings, but the property itself is an int, so that would require rewriting the option property completely, as you mentioned. 

Wait, you want have the list of strings as an object with the indices as keys? That's exactly what it is now?

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
ylvaselling: I thought about adding a hook for it but it seemed neater to me to change the property itself in the engine rather than having an extra hook hiding the "ugly" code. Also making it a hook will break the pattern that we now have with the additional data

ylvaselling:
> I was instructed that I needed the latest commit in the engine for it to work :) Please remember to specify that in the PR next time
> 
> Looks good to me, but I noticed that it's not possible to change the option using the keyboard? We should maybe look at fixing that

It is possible, but it is not intuitive. You need to tab into the menu and then change the radio with the <-- and --> buttons. You could try out the commit before that (and the previous commit in the engine branch) to try out another option with checkboxes and see which one you prefer

WeirdRubberDuck:
> > I was instructed that I needed the latest commit in the engine for it to work :) Please remember to specify that in the PR next time
> > Looks good to me, but I noticed that it's not possible to change the option using the keyboard? We should maybe look at fixing that
> 
> It is possible, but it is not intuitive. You need to tab into the menu and then change the radio with the <-- and --> buttons. You could try out the commit before that (and the previous commit in the engine branch) to try out another option with checkboxes and see which one you prefer

Ohhh, I tried everything but those keys 😅 Yes, not intuitive?  

Checkboxes feels worse for this use case - the should not be used for cases where only one option can be selected.   Was there a reason why we did not want to use the combobox?

ylvaselling:
> > > I was instructed that I needed the latest commit in the engine for it to work :) Please remember to specify that in the PR next time
> > > Looks good to me, but I noticed that it's not possible to change the option using the keyboard? We should maybe look at fixing that
> > 
> > 
> > It is possible, but it is not intuitive. You need to tab into the menu and then change the radio with the <-- and --> buttons. You could try out the commit before that (and the previous commit in the engine branch) to try out another option with checkboxes and see which one you prefer
> 
> Ohhh, I tried everything but those keys 😅 Yes, not intuitive?
> 
> Checkboxes feels worse for this use case - the should not be used for cases where only one option can be selected. Was there a reason why we did not want to use the combobox?

The combobox was not possible to select with keyboard, which is the issue I'm trying to solve with this PR hehe

src/components/Property/Types/OptionProperty.tsx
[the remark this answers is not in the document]
WeirdRubberDuck: There's a slight _conceptual_ difference, since now you create the object with the key as index, while what I'm suggesting would mean to just return a list of strings, with no explicit information about the indices. If our javascript API worked the way it _should_, it would then only return an ordered list of options, i.e. in the format `["Option 1", "Option 2", ...]`

But it's not a big deal, especially since we might want to reimplement this property type down the 

And yea, I was not suggesting to rewrite the property type as part of this and don't think it makes sense to do so. That's a future task :)

src/components/Property/Types/OptionProperty.tsx
```diff
@@ -11,30 +11,20 @@ export function OptionProperty({ uri, readOnly }: PropertyProps) {
...
       placeholder={'Choose an option'}
       disabled={readOnly}
-      data={options}
-      value={options[value]}
-      onChange={onChange}
+      // For each entry in the options object, the numeric value is the key, and the
+      // label is the value
+      data={Object.entries(options).map(([value, label]) => {
+        return { value: value, label: label };
+      })}
```
WeirdRubberDuck, line 25: We could write it like this, actually. With the comment I think it's clear enough what's going on and this si shorter and more straightforward :) 
```suggestion
      data={Object.entries(options).map(([value, label]) => ({ value, label }))}
```

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -62,15 +73,39 @@ export function ViewMenu() {
...
+            <InfoBox
+              text={`Controls what settings will be exposed in the interface. Increase the
+                level to reveal more advanced settings.`}
+            />
+          </Group>
+        </Menu.Label>
+        <Radio.Group
+          value={propertyVisibility?.toString()}
+          onChange={(newValue) => setPropertyVisibility(parseInt(newValue))}
+        >
```
WeirdRubberDuck, line 101: The reason why the up/down keys do not work is because the event "bubbles" and gets handled by the top level. if you add `onKeyDown={(event) => event.stopPropagation()}`, it works. :) 

The accessibility guidelines for radio inputs requires up/down to work as well as left/right: https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-rating/

```suggestion
        <Radio.Group
          value={propertyVisibility?.toString()}
          onChange={(newValue) => setPropertyVisibility(parseInt(newValue))}
          onKeyDown={(event) => event.stopPropagation()}
        >
```

WeirdRubberDuck (approved):

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
WeirdRubberDuck: Guessing we had the same issue with the ComboBox then, so if we liked it more, we could bring it back with the same fix. 

Personally, I think it more clearly shows the current state. But, the radio buttons more clearly show the states that are available.

PR #99, opened 2025-03-21
Feature/copy uri

ylvaselling:
Changes the tooltip to a popover and adds the copy button. This should also solve the issue that links in info areas are not clickable - they wont be clickable now either, but they will be selectable, which i think might be good enough. It also solves the issue that the transition effect looked weird in CEF.

Lmk what you think:-)

Built on top of the pr focusable info icon so only last commit is interesting


engbergandreas (approved):

src/components/InfoBox/InfoBox.tsx
```diff
@@ -1,18 +1,47 @@
...
+import { useState } from 'react';
+import { ActionIcon, Code, Group, Popover } from '@mantine/core';
 
 import { InformationIcon } from '@/icons/icons';
 
+import { CopyToClipboardButton } from '../CopyToClipboardButton/CopyToClipboardButton';
+
 interface Props {
   text: React.ReactNode;
+  uri?: string;
```
WeirdRubberDuck, line 10: Add a comment describing this prop?  Not super clear what it's used for, since info-boxes are completely separate from properties. 

Honestly, I would personally prefer to move this out of the InfoBox component, and maybe make a separate component for the properties that add the URI part. And have the InfoBox know nothing about URIs.  Thoughts? 

I.e, that component would do this: 
```
      <InfoBox
        text={
          <>
            {description}
            <Group pt={'sm'}>
              <Code>Copy URI:</Code>
              <CopyToClipboardButton value={uri} />
            </Group>
          </>
        }
      />
```
Thoughts?

WeirdRubberDuck (approved):
It's so beautiful!  And feel really intuitive!  

One comment about moving the URI out of the `InfoBox` component, and keep that component more generic

src/components/InfoBox/InfoBox.tsx
ylvaselling: Made InfoBox more general, now it accepts children. Lmk what you think

WeirdRubberDuck: 👍 Good solution! Makes perfect sense :)

PR #101, opened 2025-03-25
Fix outline confusion in time panel

ylvaselling:
Changes the outline in the time panel when pressing the lock icon to another type of outline

### DIFF
--- a/src/windowmanagement/data/MenuItems.tsx
+++ b/src/windowmanagement/data/MenuItems.tsx
@@ -113,7 +113,7 @@ export const menuItemsData: Record<string, MenuItem> = {
     ),
     renderIcon: (size) => <CalendarIcon size={size} />,
     preferredPosition: 'float',
-    floatPosition: { offsetY: 100, offsetX: 370, width: 410, height: 600 },
+    floatPosition: { offsetY: 100, offsetX: 370, width: 410, height: 520 },
     defaultVisible: true
   },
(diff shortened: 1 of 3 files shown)


src/windowmanagement/data/MenuItems.tsx
```diff
@@ -113,7 +113,7 @@ export const menuItemsData: Record<string, MenuItem> = {
     ),
     renderIcon: (size) => <CalendarIcon size={size} />,
     preferredPosition: 'float',
-    floatPosition: { offsetY: 100, offsetX: 370, width: 410, height: 520 },
+    floatPosition: { offsetY: 100, offsetX: 370, width: 410, height: 600 },
```
engbergandreas, line 116: Why this massive height increase? Tried it on 1080p monitor and there is a lot of extra space at the bottom now?

engbergandreas (approved):

src/windowmanagement/data/MenuItems.tsx
ylvaselling: For when you press the lock to not get the scrollbar. We could be ok with the scrollbar though, thats ok with me.

engbergandreas: I think that feature is so rarely used i would say ita fine. I'd be more annoyed with the extra dead space att the bottom than the scrollbar appearing once in a blue moon

PR #103, opened 2025-03-25
Fixed width and update style for More/Less button in filterlist

WeirdRubberDuck:
Making this a PR since I want a thumbs up for the updated styling fo the button. The old one blended in with the input field and all the surrounding buttons in a way I didn't like. 

Other suggestions are welcome!

Now VS before
imageimage

### DIFF
--- a/src/components/FilterList/FilterListInputField.tsx
+++ b/src/components/FilterList/FilterListInputField.tsx
@@ -24,7 +24,7 @@ function InputButton({ showMoreButton }: InputButtonProps) {
 
   return (
     showMoreButton && (
-      <Button w={80} onClick={toggleShowDataInstead}>
+      <Button w={80} variant={'light'} color={'gray'} onClick={toggleShowDataInstead}>
         {showDataInstead ? 'Less' : 'More'}
       </Button>
     )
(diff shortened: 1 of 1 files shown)


engbergandreas:
Definitely necessary but this does not pass WCAGG AA level for UI components; its at 1.22.1 contrast ratio

WeirdRubberDuck:
> Definitely necessary but this does not pass WCAGG AA level for UI components; its at 1.22.1 contrast ratio

You mean the white text against the gray?  Then we have bigger problems...

ylvaselling:
I agree with the contrast issue, it seems to me the button is a bit "washed out". Maybe the subtle variant could be nice (the one that doesn't have any background at all)? For these two I prefer the old one:)

Alternatively we could make the filterlist input field darker, like the buttons up to the top left in the screenshots. That would resolve the contrast issues

WeirdRubberDuck:
Here's how it looks with the "subtle" variant. I think I prefer this over the old one :) How about you?

I see your point on the contrast. Although, the old button border against the gray background probably has the same issue. 

If we change the input field to a darker color we probably risk not having enough contrast between the input field itself and the background. And messing with this color seems dangerous - Mantine has probably set it in a way so it should look ok in most cases (or so I would like to think, at least)

image

PR #104, opened 2025-03-25
Feature/highlight open panels

engbergandreas:
highlights open panels in the taskbar and windows menu, and also add a 'x' button to close windows immediately from the windows


ylvaselling:
Looks great! 🎉 

I know we went over this code and this solution is fine, but now that I looked at it again it seemed to me the most straightforward solution would be to pass `item` (or maybe just `item.id`) to the `TaskBarMenuButton`, and in that component place the event handlers and get the `isOpen` state from redux. I'm not sure if we talked about this solution and decided against it for some reason? If we have the id of the menuButton we should be able to do everything inside the `TaskBarMenuButton` component. 🤔 That way, if we decide we need more functionality in the `TaskBarMenuButton`, it will be easier to change

src/panels/Menu/TaskBar/TaskBar.tsx
```diff
@@ -61,9 +61,13 @@ export function TaskBar() {
...
               {item?.renderMenuButton ? (
-                item.renderMenuButton(eventHandlers(item))
+                item.renderMenuButton(eventHandlers(item), itemConfig)
               ) : (
-                <TaskBarMenuButton {...eventHandlers(item)} aria-label={item.title}>
+                <TaskBarMenuButton
+                  {...eventHandlers(item)}
+                  isOpen={itemConfig.isOpen}
+                  aria-label={item.title}
+                >
```
ylvaselling, line 70: This here is what I'm referring to in my other comment. We could just pass in item.id here, and do all the work in the `TaskBarMenuButton`. Or am I missing something? We could even place the aria-label in there which would mean we only need to pass 1 prop. That would be neat!

src/panels/OriginPanel/OriginPanelMenuButton.tsx
```diff
@@ -41,12 +42,14 @@ export function OriginPanelMenuButton({ eventHandlers }: Props) {
       aimName={aimNode?.name}
       isOpenSpaceReady={isReady}
       eventHandlers={eventHandlers}
+      isOpen={isOpen}
```
ylvaselling, line 45: I guess if we went by that change this would need to be extracted from redux an extra time

src/windowmanagement/data/MenuItems.tsx
```diff
@@ -18,15 +18,15 @@ import {
...
 import { OriginPanelMenuButton } from '@/panels/OriginPanel/OriginPanelMenuButton';
 import { TempPropertyTest } from '@/panels/Scene/TempPropertyTest';
 import { SessionRecordingMenuButton } from '@/panels/SessionRecordingPanel/SessionRecordingMenuButton';
 import { TimePanelMenuButton } from '@/panels/TimePanel/TimePanelMenuButton';
 import { IconSize } from '@/types/enums';
 import { MenuItemEventHandlers } from '@/types/types';
 
-import { FloatWindowPosition } from '../WindowLayout/types';
-import { WindowLayoutPosition } from '../WindowLayout/WindowLayout';
+import { FloatWindowPosition, WindowLayoutPosition } from '../WindowLayout/types';
```
ylvaselling, line 29: use @ notation?

src/windowmanagement/data/MenuItems.tsx
```diff
@@ -50,7 +50,10 @@ export interface MenuItem {
   title: string; // Title of the rc-dock tab
   componentID: string; // Unqiue ID to identify this component among the rc-dock tabs
   content: React.JSX.Element; // Content to render inside the rc-dock tab
-  renderMenuButton?: (eventHandlers: MenuItemEventHandlers) => React.JSX.Element; // Custom menu button to render
+  renderMenuButton?: (
+    eventHandlers: MenuItemEventHandlers,
+    config: TaskbarItemConfig
+  ) => React.JSX.Element; // Custom menu button to render
```
ylvaselling, line 56: So in the above scenario, this function would only take an id

ylvaselling (approved):

src/panels/Menu/TopMenuBar/Menus/WindowsMenu.tsx
```diff
@@ -17,20 +17,37 @@ export function WindowsMenu() {
...
-                title: item.title,
-                position: item.preferredPosition,
-                id: item.componentID,
-                floatPosition: item.floatPosition
-              });
-            }}
-          >
-            {item.title}
-          </Menu.Item>
+          <Group>
```
WeirdRubberDuck, line 20: How about reducing the cap here? Maybe either to `'xs'` or `5`?
```suggestion
          <Group gap={'xs'}>
```

src/panels/Menu/types.ts
```diff
@@ -6,4 +6,5 @@ export interface TaskbarItemConfig {
   // the variable since we can otherwise repurpose it for whether a panel is open/closed
   enabled: boolean; // Whether the menu item is enabled, for example the missions panel
   // is enabled = false when a mission is not loaded or if a specific module is disabled
+  isOpen: boolean;
```
WeirdRubberDuck, line 9: Add a comment for this variable as well?  Just to be clear

Also, I would like it if we moved long comments like this to start avobe the variable, like this: 
```
  // Whether the menu item should be visible in the taskbar.
  // @TODO anden88 2025-03-12: determine if we will be using the enabled property to
  // to disable using a certain menu, e.g., disable Sky browser in a dome. Leaving the
  // the variable since we can otherwise repurpose it for whether a panel is open/closed
  visible: boolean; 
  // Whether the menu item is enabled, for example the missions panel
  // is enabled = false when a mission is not loaded or if a specific module is disabled
  enabled: boolean; 
```
not it's quite hard to see where the comment starts

src/panels/Menu/TopMenuBar/Menus/WindowsMenu.tsx
```diff
@@ -17,20 +17,37 @@ export function WindowsMenu() {
...
+              key={item.componentID}
+              leftSection={item.renderIcon?.(IconSize.xs)}
+              onClick={() => {
+                addWindow(item.content, {
+                  title: item.title,
+                  position: item.preferredPosition,
+                  id: item.componentID,
+                  floatPosition: item.floatPosition
+                });
+              }}
```
WeirdRubberDuck, line 31:
```suggestion
              onClick={() =>
                addWindow(item.content, {
                  title: item.title,
                  position: item.preferredPosition,
                  id: item.componentID,
                  floatPosition: item.floatPosition
                })
              }
```

WeirdRubberDuck (approved):
Nice!  This turned our really well

Code looks good, but I think I like the solution that Ylva is suggesting even more :) Would make it a little less cluttered

src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
```diff
@@ -13,6 +21,10 @@ export function TaskBarMenuButton({ onClick, onRightClick, children, ...props }:
       size={'xl'}
       variant={'menubar'}
       {...props}
+      style={{
+        borderBottom: isOpen ? '4px solid var(--mantine-primary-color-filled)' : '',
```
WeirdRubberDuck, line 25: do we want to make a variable for this style?  Now it's used in three places :) (including the selected node in the scene tree)

ylvaselling: Is it possible to add something to the theme?

WeirdRubberDuck: Not that I'm aware of, at least

WeirdRubberDuck: And we use the border for different locations (left VS bottom)

src/panels/Menu/TopMenuBar/Menus/WindowsMenu.tsx
```diff
@@ -17,20 +17,36 @@ export function WindowsMenu() {
...
+                  title: item.title,
+                  position: item.preferredPosition,
+                  id: item.componentID,
+                  floatPosition: item.floatPosition
+                })
+              }
+              style={{
+                borderLeft: itemConfig.isOpen
+                  ? 'var(--openspace-border-active)'
+                  : '4px solid transparent',
```
WeirdRubberDuck, line 34: Maybe this is overkill, but I can imagine that there will easily be a mismatch between this hardcoded 4 pixels and the active border size. 

How about adding either:
1. The closed style in a variable `--openspace-border-active-placeholder` 
or 
2. A variable for just the width, so that this looks like this 
```
                borderLeft: itemConfig.isOpen
                  ? 'var(--openspace-border-active)'
                  : 'var(--openspace-border-active-width) solid transparent',
```
and the theme like this: 
```
export const cssVariablesResolver: CSSVariablesResolver = () => ({
  variables: {
    '--openspace-border-active-width': '4px',
    '--openspace-border-active':
      'var(--openspace-border-active-width) solid var(--mantine-primary-color-filled)'
  },
```

?  Maybe 1. is preferable? Cleaner, and I can see us wanting to use that style in more places

WeirdRubberDuck (approved):
Good stuff 👍

PR #106, opened 2025-03-26
Add time frame info for Scene Graph Nodes

WeirdRubberDuck:
To test, start for example the James Webb profile. The JWST renderables have a time frame attached to them

In timeframe vs out of time frame (advanced user): 

Example of how it looks in the list when searching for things (wanted this to be pretty subtle and not distract too much)

OBS In 'User' Mode, the TimeFrame is not shown. What do we think about this text here then? I'm not quite sure what to do about it . Other suggestions for this view? what information is needed?  There's a tooltip on hover, but do we maybe need an infobox somewhere?  Or add one to the TimeFrame property owner?
image


src/panels/Scene/SceneGraphNode/SceneGraphNodeHeader.tsx
```diff
@@ -51,12 +55,28 @@ export function SceneGraphNodeHeader({ uri, onClick, label }: Props) {
...
+  // TODO: Make sure that the timeframe information is accessible
+  const visibilityControl = hasRenderable && (
+    <Group gap={'xs'}>
+      <PropertyOwnerVisibilityCheckbox uri={renderableUri} />
+      {timeFrame && !isInTimeFrame && (
+        <Tooltip
+          label={`This node is currently hidden since the time is outside its specified
+            time frame. It will not be visible even if enabled.`}
+          w={300}
+          multiline
```
WeirdRubberDuck, line 67:
```suggestion
```
Added this to the theme

src/util/keys.ts
```diff
@@ -40,6 +40,3 @@ export const RecordingsFolderKey = '${RECORDINGS}/';
 export const WindowsKey = 'Windows';
 export const ScreenSpaceKey = 'ScreenSpace';
 export const RenderableKey = 'Renderable';
-export const ScaleKey = 'Scale';
```
WeirdRubberDuck, line 43: OBS Leaning in to a reduced use of keys. It does make the code easier to read, since this is how we've started doing it more and more

src/panels/Scene/SceneGraphNode/SceneGraphNodeHeader.tsx
```diff
@@ -51,12 +55,28 @@ export function SceneGraphNodeHeader({ uri, onClick, label }: Props) {
...
   );
 
+  // TODO: Make sure that the timeframe information is accessible
+  const visibilityControl = hasRenderable && (
+    <Group gap={'xs'}>
+      <PropertyOwnerVisibilityCheckbox uri={renderableUri} />
+      {timeFrame && !isInTimeFrame && (
+        <Tooltip
+          label={`This node is currently hidden since the time is outside its specified
+            time frame. It will not be visible even if enabled.`}
```
engbergandreas, line 65:
```suggestion
          label={`This node is hidden because the current time is outside its specified
            time frame. It will not be visible even if enabled.`}
```
?

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
```diff
@@ -99,6 +129,44 @@ export function SceneGraphNodeView({ uri }: Props) {
...
+                  >
+                    {isInTimeFrame ? 'Active' : 'Inactive'}
+                  </Badge>
+                </Tooltip>
+              </Group>
+              <Text mt={'xs'} size={'sm'} c={'dimmed'}>
+                The object will not be visible outside the specified time range.
+              </Text>
+            </Box>
+            <PropertyOwner uri={timeFrame.uri} />
```
engbergandreas, line 166: I guess this is information coming from OpenSpace rather than frontend. But can we get which unit and timeframe (seconds since J2000 ?) in the tooltip for start and end input fields. Hard to know what the start and end times represent right now

engbergandreas (approved):

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
WeirdRubberDuck: Thats an update that needs to be done in OpenSpace, so not related to this PR.

A discussed solution is to add a viewoption for numbers that represent time (which are currently always seconds since J2000). The time frame property owners is only ever showed for advanced users so I don't think it's very urgent to fix. 

It would be nice to also get a string representation of the dates where the time frame is active, but that's also something that we need to build in OpenSpace since there are different time frame types - not only a simple interval like the one used for JWST

ylvaselling:
I think it looks great! 🎉 

Some thoughts:
* In the tooltip we mentioned the name "Time Frame", as well as in the info text. If we are at user level, as you mentioned, there is no way to find what this Time Frame is. I think 
    1. Either we hide the time frame at user level and we leave out that name from the tooltip. We could write someting like "This object is not visible right now". However, I think that might be frustrating for a user since they are not able to see when the object will be visible.
    2. We don't hide the time frame at user level. We can then work on improving the look of the time frame so that it is easy to understand. Conceptually the time frame doesn't seem like an "advanced" feature to me. Anyone should be able to understand that an object is visible between two points in time I think 🤔

src/panels/Scene/hooks.ts
```diff
@@ -57,3 +60,13 @@ export function useOpenCurrentSceneNodeWindow() {
     closeCurrentNodeWindow
   };
 }
+
+export function useTimeFrame(uri: Uri) {
```
ylvaselling, line 64: Nice!

src/theme/mantineTheme.ts
```diff
@@ -82,7 +90,9 @@ export const theme = createTheme({
       defaultProps: {
         withArrow: true,
         transitionProps: { duration: 400, enterDelay: 400 },
-        position: 'top'
+        position: 'top',
+        maw: 300,
+        multiline: true
```
ylvaselling, line 95: 👍

ylvaselling (approved):

WeirdRubberDuck:
> I think it looks great! 🎉
> 
> Some thoughts:
> 
> * In the tooltip we mentioned the name "Time Frame", as well as in the info text. If we are at user level, as you mentioned, there is no way to find what this Time Frame is. I think
>   
>   1. Either we hide the time frame at user level and we leave out that name from the tooltip. We could write someting like "This object is not visible right now". However, I think that might be frustrating for a user since they are not able to see when the object will be visible.
>   2. We don't hide the time frame at user level. We can then work on improving the look of the time frame so that it is easy to understand. Conceptually the time frame doesn't seem like an "advanced" feature to me. Anyone should be able to understand that an object is visible between two points in time I think 🤔
> * I'm wondering whether we want to show the time frame symbol in the scene graph node header even when the time frame is active. I found myself looking for it. It could be a nice indicator it has a time frame. We could potentially also make it have the same color as in the tabs, for extra clarity.

1. Good points here!  My current thought would be to keep the "time frame" in the name and tooltip, and instead make sure that we implement some good human readable way that tells exactly WHEN the object is active. This needs to be done in the engine, however, so it's a TODO. I personally think this information is very important, and some people at the office seemed to agree. I can make an issue after this PR. :) 
    - You're not really supposed to change the TimFrame during runtime, which is why it's an advanced feature. But we need to communicate which times the object is available
2. I guess we could show an icon that on-hover says something along the lines of "this object has a time frame, and will hence be hidden at certain times". However, some points against it: 
    -  IMO, it's not really relevant to the user at the point of enabling/disabling objects, which is what this part of the menu is for. If the object is not disabled due to it's time frame, there's no reason to inform about it
    - If the idea is to show which nodes have a time frame, the user already gets that information if opening the scene graph node
    - Reduce clutter: If we add the icon also for when it's active, because the icons are so similar, we need to clearly visually distinguish between the two in some way, e.g., using color (which I'm not a fan of😬) => visual clutter 
    - In the future, we might want to add other symbols that signify if an object is unavailable for some specific reason

ylvaselling:
1. Sounds good 👍 If the time frame is not supposed to be editable, doesn't that suggest it is a read only feature rather than an advanced feature? You can always bypass our restrictions with the console anyway, right? 
2. 👍

WeirdRubberDuck:
> 1. Sounds good 👍 If the time frame is not supposed to be editable, doesn't that suggest it is a read only feature rather than an advanced feature? You can always bypass our restrictions with the console anyway, right?
> 2. 👍

1. Unsure - that's better asked to someone who has worked with it more. I can imagine that we would like to be able to edit it while deciding on what a good time frame for something is, for example 🤷

PR #107, opened 2025-03-26
Feature/make scene panel not split emmasuggestion

WeirdRubberDuck:
Small suggestion to merge into the branch for this PR: https://github.com/OpenSpace/OpenSpace-WebGui/pull/100

Addresses some things that I did not know how to comment on: 
1. `SceneGraphNodeView` should not have to know about the currently selected node - IMO.  Open to discussion here, if adding back the "closable" variable is preferable to you
2. Tighten up the UI a bit - i.e. remove some padding
3. Make a separate component for the button that opens a SGN in another window
4. Do not show this button in a popped out window

ONly for @ylvaselling. @engbergandreas you can ignore this one :)

### DIFF
--- a/src/panels/Scene/SceneGraphNode/OpenInNewWindowButton.tsx
+++ b/src/panels/Scene/SceneGraphNode/OpenInNewWindowButton.tsx
@@ -1,0 +1,38 @@
+import { ActionIcon } from '@mantine/core';
+
+import { usePropertyOwner } from '@/hooks/propertyOwner';
+import { OpenInNewIcon } from '@/icons/icons';
+import { Uri } from '@/types/types';
+import { displayName } from '@/util/propertyTreeHelpers';
+import { useWindowLayoutProvider } from '@/windowmanagement/WindowLayout/hooks';
+
+import { SceneGraphNodeView } from './SceneGraphNodeView';
+
+interface Props {
+  uri: Uri;
+}
+
+export function OpenInNewWindowButton({ uri }: Props) {
+  const propertyOwner = usePropertyOwner(uri);
+  const { addWindow } = useWindowLayoutProvider();
+
+  if (!propertyOwner) {
+    return <></>;
+  }
+
+  const name = displayName(propertyOwner);
+
+  function openInNewWindow() {
+    addWindow(<SceneGraphNodeView uri={uri} showOpenInNewWindow={false} />, {
+      id: 'sgn-' + uri,
+      title: name,
+      position: 'float'
+    });
+  }
+
+  return (
+    <ActionIcon onClick={openInNewWindow} size={'sm'} flex={0}>
+      <OpenInNewIcon />
+    </ActionIcon>
+  );
+}
--- a/src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
+++ b/src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
@@ -61,7 +61,14 @@ export function SceneGraphNodeView({
   return (
     <Layout>
       <Layout.FixedSection>
-        <Group gap={'xs'} py={'xs'} pl={'xs'} grow preventGrowOverflow={false}>
+        <Group
+          gap={'xs'}
+          py={'xs'}
+          pl={'xs'}
+          grow
+          preventGrowOverflow={false}
+          wrap={'nowrap'}
+        >
           {/* The key here is necessary to make sure that the component is re-rendered
           when the anchor node changes, to trigger the animation */}
           <Box key={uri} className={styles.highlight}>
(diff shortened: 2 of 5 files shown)


src/panels/Scene/SceneGraphNode/OpenInNewWindowButton.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { displayName } from '@/util/propertyTreeHelpers';
+import { useWindowLayoutProvider } from '@/windowmanagement/WindowLayout/hooks';
+
+import { SceneGraphNodeView } from './SceneGraphNodeView';
+
+interface Props {
+  uri: Uri;
+}
+
+export function OpenInNewWindowButton({ uri }: Props) {
```
ylvaselling, line 15:
```suggestion
export function OpenSgnInNewWindowButton({ uri }: Props) {
```
To make it clear it is a scene graph node:)

ylvaselling: Make we want to make this button more general. We already have the same button in the skybrowser panel, for popping out the wwt window. If we pass the data passed to addWindow as a prop, I think that could potentially be a useful component to have 🤔

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
```diff
@@ -17,15 +15,20 @@ import styles from './SceneGraphNodeView.module.css';
 
 interface Props {
   uri: Uri;
-  closeable?: boolean;
+  // Extra content to be shown at the top right of the view
+  extraTopControls?: React.ReactNode;
```
ylvaselling, line 19: I can't really see a use case for this prop to be used for anything else than the close button 🤔 Especially where it is placed, so this seems overly general to me. 

I can see that it is conceptually nice to have the closing code in the scene panel, but then I would prefer to call this prop `closeButton` or something similar unless we can come up with another use case for it

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
```diff
@@ -17,15 +15,20 @@ import styles from './SceneGraphNodeView.module.css';
...
+  showOpenInNewWindow?: boolean;
 }
 
-export function SceneGraphNodeView({ uri, closeable = false }: Props) {
+export function SceneGraphNodeView({
+  uri,
+  extraTopControls,
+  showOpenInNewWindow
```
ylvaselling, line 26: To mimic the behavior in the SceneGraphNodeHeader, set this to true by default?

src/panels/Scene/ScenePanel.tsx
```diff
@@ -21,7 +26,16 @@ export function ScenePanel() {
...
-          <SceneGraphNodeView uri={currentlySelectedNode} closeable />
+          <SceneGraphNodeView
+            uri={currentlySelectedNode}
+            extraTopControls={
+              <CloseButton
+                flex={0}
+                onClick={() => dispatch(setSceneTreeSelectedNode(null))}
+              />
+            }
+            showOpenInNewWindow
```
ylvaselling, line 37: If it is default to true, we don't need this here

ylvaselling (approved):
🎉 Looks good overall!

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
WeirdRubberDuck: Myeah my problem was that to me it didnät feel like the "close button" belonged in the scene graph node view component.  In the suggestion I sent an image to you (where the close button was above the SGN header), I had moved it out of this view completely, but realized that this led to way to much screen space being wasted

Ideally, I would not want to have this in here at all

src/panels/Scene/SceneGraphNode/OpenInNewWindowButton.tsx
WeirdRubberDuck: Good point!  Maybe we could do a refactor after the PRs are in? 

Would be nice to have all the "open in new window" buttons look and work the same :)

ylvaselling: That one is already in master, but sure we can do a refactor later

src/panels/Scene/SceneGraphNode/SceneGraphNodeView.tsx
ylvaselling: I agree to both accounts :) maybe we can split the scene graph node view component into two parts. one with the tabs and one with the header. That way, we can make a tiny component for the SceneGraphNodeView that is used in windows standalone, and just use the tab component in the scene menu

WeirdRubberDuck: Yeah I think that would be better :)  I've been thinking that we maybe want to use a different header component for the SGN View anyway. Using the same one as in the list feels pretty limiting as of now

PR #108, opened 2025-03-28
Fix getting started guide buttons that broke in recent rewrite

WeirdRubberDuck:
Unfortunately, the solution we had for showing the taskbar buttons broke in the recent rewrite. This fixes them, and also makes the button unclickable

I made what I found to be the easiest solution, which is to just send in an empty string for the `id`. This felt more intentional than e.g. not specifying the `id` at all and led to minimal code changes. 

Let me know what you think. This should get in before the user test

### DIFF
--- a/src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
+++ b/src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
@@ -14,7 +14,21 @@ export function TaskBarMenuButton({ id, children, ...props }: Props) {
   );
   const { addWindow, closeWindow } = useWindowLayoutProvider();
 
+  if (id === '') {
+    // This is a special case used to show the buttons in the getting started guide.
+    // These buttons should not be clickable
+    return (
+      <Button px={'sm'} size={'lg'} {...props} style={{ pointerEvents: 'none' }}>
+        {children}
+      </Button>
+    );
+  }
+
   const item = menuItemsData[id];
+
+  if (!item) {
+    throw new Error(`No menu item found for id: ${id}`);
+  }
 
   function onClick(event: React.MouseEvent<HTMLButtonElement, MouseEvent>) {
     if (event.shiftKey) {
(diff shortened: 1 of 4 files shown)


src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
```diff
@@ -41,7 +55,7 @@ export function TaskBarMenuButton({ id, children, ...props }: Props) {
       onContextMenu={onRightClick}
       size={'xl'}
       variant={'menubar'}
-      aria-label={item.title}
+      aria-label={item?.title}
```
WeirdRubberDuck, line 58:
```suggestion
      aria-label={item.title}
```

src/panels/Menu/TaskBar/TaskBarMenuButton.tsx
```diff
@@ -14,8 +14,22 @@ export function TaskBarMenuButton({ id, children, ...props }: Props) {
...
+      <Button px={'sm'} size={'lg'} {...props} style={{ pointerEvents: 'none' }}>
+        {children}
+      </Button>
+    );
+  }
+
   const item = menuItemsData[id];
 
+  if (!item) {
+    throw new Error(`No menu item found for id: ${id}`);
```
engbergandreas, line 30:
```suggestion
    throw new Error(`No menu item found for id: '${id}'`);
```

engbergandreas:
I think this solution is okay-ish, I'm not a huge fan of having to add a bunch of logic that deals with the getting started tour. I think that adds unnecessary coupling. But I can't also think of a better solution that doesn't have other issues, like adding guards in the `addWindow` and `closeWindow` functions.. so I'm ok with it

WeirdRubberDuck:
> I think this solution is okay-ish. I'm not a huge fan of adding a bunch of logic that deals with the getting started tour. I think that adds unnecessary coupling. But I can't think of a better solution that doesn't have other issues, like adding guards in the `addWindow` and `closeWindow` functions. so I'm ok with it

I agree with you, but didn't really know what else to do about it and it had to be fixed

ylvaselling:
Here is an alternative solution: https://github.com/OpenSpace/OpenSpace-WebGui/pull/111

PR #110, opened 2025-03-28
Only show focus style on keyboard navigation

WeirdRubberDuck:
I don't know about you, but I found it a bit annoying that the focus style is visible even on mouse click (it clutters the view unnecessarily, IMO). This changes so it is only visible when using keyboard navigation. 

However, if you prefer to always have it on, this PR can be closed. Wanted to suggest it at least

Source: 
https://mantine.dev/theming/theme-object/#focusclassname

### DIFF
--- a/src/theme/mantineTheme.module.css
+++ b/src/theme/mantineTheme.module.css
@@ -1,4 +1,4 @@
-.focus:focus {
+.focus:focus-visible {
   outline-offset: -2px;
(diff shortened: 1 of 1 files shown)


engbergandreas:
Yeah agree, found it weird as well.

engbergandreas (approved):

ylvaselling (approved):
👍

PR #111, opened 2025-03-31
Feature/fix getting started guide buttons ylvasuggestion

ylvaselling:
Alternative solution to `fix-getting-started-guide`


src/panels/OriginPanel/OriginPanelMenuButton.tsx
```diff
@@ -37,17 +30,13 @@ export function OriginPanelMenuButton({ id }: Props) {
     );
   }
 
-  return hasDistinctAim ? (
+  if (!isReady) {
+    return <Skeleton>Anchor</Skeleton>;
+  }
```
engbergandreas, line 35: This needs to be wrapped in a button or the size will be weird

src/panels/OriginPanel/OriginPanelMenuButton.tsx
```diff
@@ -34,14 +31,16 @@ export function OriginPanelMenuButton({ id }: Props) {
...
-      id={id}
-    />
-  ) : (
-    <FocusButton anchorName={anchorNode?.name} isOpenSpaceReady={isReady} id={id} />
+  if (!isReady) {
+    return (
+      <TaskBarMenuButton id={id} disabled>
+        <LoadingBlocks n={1} w={84} />
+      </TaskBarMenuButton>
+    );
```
WeirdRubberDuck, line 39: While I like the simplicty of this, I think we might want the focus icon to still be there?  At least if, as we've discussed, the longer-term plan is to not disable the menu button while loading. Then, the user needs a way to know which menu this is. 

To clarify, I suggest to make this button look the way it did before
image
at least including the icon, but personally I would prefer to add the focus part as well, or a loading block that represents that text.  The blocks makes the most sense if they visually represent the thing that is loading directly, at least IMO

src/panels/GettingStartedPanel/Steps/NavigationSteps.tsx
```diff
@@ -76,7 +76,9 @@ export const NavigationSteps: React.JSX.Element[] = [
     <Text>
       Open the navigation menu and click on the "Focus" button to focus on the Moon. The
       navigation menu can be found in the bottom bar:
-      <OriginPanelMenuButton onClick={() => {}} />
+      <Box mt={'md'} p={'md'}>
+        <OriginPanelMenuButtonContent />
+      </Box>
```
WeirdRubberDuck, line 81: I think it would be good with some border around these buttons, to make it clearer that it is one entity. How about:
```suggestion
      <Paper mt={'md'} p={'xs'} w={'fit-content'} withBorder bg={'transparent'}>
        <OriginPanelMenuButtonContent />
      </Paper>
```
image

or 
```suggestion
      <Paper mt={'md'} py={'xs'} px={'md'} w={'fit-content'}>
        <OriginPanelMenuButtonContent />
      </Paper>
```
image

? 

And maybe make this wrapper a helper component that we can apply to the other buttons we use here as well? 

I already did something similar for the `SceneGraphNodeHeader` in COntentSteps.tsx, for example. That could also use the wrapper component

WeirdRubberDuck (commented):
The structure makes sense to me 👍 See some comments though

src/panels/OriginPanel/OriginPanelMenuButton.tsx
ylvaselling: Technically we don't know if this will show the anchor / aim or just the focus button 🤔 I guess we could always make the placeholder the focus anyway 🤷 but if we want to remove this functionality altogether it doesn't make sense to change it imo

WeirdRubberDuck: Yeah making the placeholder the focus makes sense - or at least give it the focus icon since that's the once we use for navigation. 

Also, we wouldn't know if we're in anchor/aim mode, and the general assumption can be that we probably aren't since Focus is default :)

src/panels/GettingStartedPanel/Steps/ButtonWrapper.tsx
```diff
@@ -0,0 +1,21 @@
+import { PropsWithChildren } from 'react';
+import { MantineStyleProps, Paper } from '@mantine/core';
+
+interface Props extends PropsWithChildren, MantineStyleProps {
+  withBorder?: boolean;
+}
+
+export function ButtonWrapper({ withBorder = false, children, ...styleProps }: Props) {
```
WeirdRubberDuck, line 8: Since this is a component with a very specific use case and a general name, maybe add a short comment above it explaining the intended use case?

WeirdRubberDuck (approved):
Looks great!

src/panels/GettingStartedPanel/Steps/ButtonWrapper.tsx
```diff
@@ -0,0 +1,24 @@
+import { PropsWithChildren } from 'react';
+import { MantineStyleProps, Paper } from '@mantine/core';
+
+interface Props extends PropsWithChildren, MantineStyleProps {
+  withBorder?: boolean;
+}
+
+// This component is designed to be used as a wrapper for buttons in the Getting Started Panel
+// for buttons that we only want to displat "as an image", i.e. show what they look like
```
engbergandreas, line 9:
```suggestion
// for buttons that we only want to display "as an image", i.e. show what they look like
```

PR #112, opened 2025-04-01
Add package with types for the experimental file I/O features

WeirdRubberDuck:
Had to be fixed to create a `build` version of the page. 

Uses [this package](https://www.npmjs.com/package/@types/wicg-file-system-access), which seems to be the recommended way to deal with this.  Sources: 
https://issues.chromium.org/issues/40244598#comment2
https://stackoverflow.com/questions/71309058/property-showsavefilepicker-does-not-exist-on-type-window-typeof-globalthis

Note that this PR is based on master, which as time of creation has a couple of other issues that prevent building until #111 is merged. Tested it in the `feature/user-test-2` branch

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -33,7 +33,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "strict": true,
-    "typeRoots": ["./src/types/*.d.ts"]
+    "typeRoots": ["./src/types/*.d.ts", "./node_modules/@types"],
   },
   "references": [{ "path": "./tsconfig.app.json" }, { "path": "./tsconfig.node.json" }]
 }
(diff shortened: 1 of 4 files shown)


tsconfig.json
```diff
@@ -33,7 +33,7 @@
     "noUnusedLocals": true,
     "noUnusedParameters": true,
     "strict": true,
-    "typeRoots": ["./src/types/*.d.ts"]
+    "typeRoots": ["./src/types/*.d.ts", "./node_modules/@types"],
```
WeirdRubberDuck, line 36: I found it quite interesting that we did not already get the types from the modules. 

Weird that we haven't gotten issues before, but maybe this is the first time we explicitly need the type from a module.

ylvaselling (approved):

engbergandreas (approved):

PR #73, opened 2025-02-27
add readme

engbergandreas:
Started typing something for the readme, I kindly ask for some help with text and what else we might want in here

PR #115, opened 2025-04-04
Use markedNodes data from profile topic to populate featured nodes lists

WeirdRubberDuck:
The order in both the Scene panel and the navigation menu now match the one specified in the profile. e.g. Default profile: 

image

https://github.com/OpenSpace/OpenSpace/issues/3454

### DIFF
--- a/src/panels/OriginPanel/OriginPanel.tsx
+++ b/src/panels/OriginPanel/OriginPanel.tsx
@@ -6,9 +6,10 @@ import { Layout } from '@/components/Layout/Layout';
 import { AnchorIcon, FocusIcon, TelescopeIcon } from '@/icons/icons';
 import { useAppSelector } from '@/redux/hooks';
 import { EngineMode, IconSize } from '@/types/enums';
-import { Uri } from '@/types/types';
+import {  Uri } from '@/types/types';
 import { NavigationAimKey, NavigationAnchorKey } from '@/util/keys';
-import { hasInterestingTag, isPropertyOwnerHidden } from '@/util/propertyTreeHelpers';
+import { isPropertyOwnerHidden } from '@/util/propertyTreeHelpers';
+import { useFeaturedNodes } from '@/util/propertyTreeHooks';
 
 import { AnchorAimView } from './AnchorAimView/AnchorAimView';
 import { FocusView } from './FocusView/FocusView';
@@ -24,6 +25,8 @@ export function OriginPanel() {
   const properties = useAppSelector((state) => state.properties.properties);
   const engineMode = useAppSelector((state) => state.engineMode.mode);
 
+  const featuredNodes = useFeaturedNodes();
+
   const shouldStartInAnchorAim = useAppSelector((state) => {
     const aimProp = state.properties.properties[NavigationAimKey];
     const anchorProp = state.properties.properties[NavigationAnchorKey];
@@ -33,17 +36,6 @@ export function OriginPanel() {
     shouldStartInAnchorAim ? NavigationMode.AnchorAim : NavigationMode.Focus
   );
 
-  const sortedDefaultList = useMemo(() => {
-    const uris: Uri[] = propertyOwners.Scene?.subowners ?? [];
-    const markedInterestingNodeUris = uris.filter((uri) =>
-      hasInterestingTag(uri, propertyOwners)
-    );
-    const favorites = markedInterestingNodeUris
-      .map((uri) => propertyOwners[uri])
-      .filter((po) => po !== undefined);
-    return favorites.slice().sort((a, b) => a.name.localeCompare(b.name));
-  }, [propertyOwners]);
-
   // @TODO (2025-02-24, emmbr): Remove dependency on properties object
   const sortedSearchableNodes = useMemo(() => {
     const uris: Uri[] = propertyOwners.Scene?.subowners ?? [];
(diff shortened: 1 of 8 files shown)


src/redux/profile/profileSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+    setMarkedNodes: (state, action: PayloadAction<Identifier[]>) => {
+      state.markNodes = action.payload;
+      return state;
+    },
+  }
+});
+
+// Action creators are generated for each case reducer function, replaces the `Actions/index.js`
+export const { setMarkedNodes } = profileSlice.actions;
+export const profileReducer = profileSlice.reducer;
```
WeirdRubberDuck, line 32:
```suggestion
export const profileReducer = profileSlice.reducer;

```

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -6,9 +6,10 @@ import { Layout } from '@/components/Layout/Layout';
 import { AnchorIcon, FocusIcon, TelescopeIcon } from '@/icons/icons';
 import { useAppSelector } from '@/redux/hooks';
 import { EngineMode, IconSize } from '@/types/enums';
-import { Uri } from '@/types/types';
+import {  Uri } from '@/types/types';
```
WeirdRubberDuck, line 9:
```suggestion
import { Uri } from '@/types/types';
```

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -24,6 +25,8 @@ export function OriginPanel() {
   const properties = useAppSelector((state) => state.properties.properties);
   const engineMode = useAppSelector((state) => state.engineMode.mode);
 
+  const featuredNodes = useFeaturedNodes();
+
   const shouldStartInAnchorAim = useAppSelector((state) => {
```
ylvaselling, line 30: Hook order: 
   - useAppSelector
   - useState
   - useContext
   - useRef
   - customHooks (useOpenSpaceApi, etc)
   - derived state
   - useAppDispatch
   - useEffect
   - useMemo, useCallback
   - functions

src/redux/profile/profileSlice.ts
```diff
@@ -0,0 +1,32 @@
...
+    // else will be ignored
+    [key: string]: boolean;
+  };
+  markNodes: Identifier[];
+}
+
+const initialState: ProfileState = {
+  uiPanelVisibility: {},
+  markNodes: []
+};
```
ylvaselling, line 17: If we store the uipanel visibility here, shouldn't we set it also? If we don't store it I think there is no need to have a state for it?

src/util/keys.ts
```diff
@@ -17,7 +17,9 @@ export const JumpToFadeDurationKey = 'NavigationHandler.JumpToFadeDuration';
 // To get any scene graph node you need ScenePrefix+NodeIdentifier
 export const ScenePrefixKey = 'Scene.';
 
-export const InterestingTagKey = 'GUI.Interesting';
+// @TODO (2024-04-07, emmbr): Remove this key here and from engine - it is no longer used
```
ylvaselling, line 20: Why is it no longer used? Is it replaced by the marked nodes?

src/util/propertyTreeHooks.ts
```diff
@@ -2,26 +2,24 @@ import { useMemo } from 'react';
...
-import { Uri } from '@/types/types';
+import { PropertyOwner } from '@/types/types';
 
 import { NavigationAimKey, NavigationAnchorKey } from './keys';
-import { hasInterestingTag, sgnUri } from './propertyTreeHelpers';
+import { sgnUri } from './propertyTreeHelpers';
 
-export function useInterestingTagOwners() {
+/**
+ * Get all the nodes marked in the profile, as a memoized list of property owners.
```
ylvaselling, line 11:
```suggestion
 * Get all the nodes marked in the profile, as a list of property owners.
```
Don't think we need to worry about the memoization outside of this function

ylvaselling (changes requested):

src/util/keys.ts
WeirdRubberDuck: Yes, this tag was previously used to identify the marked nodes

src/panels/OriginPanel/OriginPanel.tsx
WeirdRubberDuck: Thanks for the reminder. Maybe we could merge this PR sometime soon...? 

https://github.com/OpenSpace/OpenSpace-WebGui/pull/73/files

src/redux/profile/profileSlice.ts
WeirdRubberDuck: Good question. I just copied over the state data structure from the middleware file... 

But looks like the top object rather describes the data structure that we get from the profile topic?  So, we should split it up into something like: 

```
export interface ProfileData {
  uiPanelVisibility: {
    // The key here should match the id of the menu item,
    // else will be ignored
    [key: string]: boolean;
  };
  markNodes: Identifier[];
}

export interface ProfileState {
  markNodes: Identifier[];
}


const initialState: ProfileState = {
  markNodes: []
};
```
but ofc move the `ProfileData` interface to the middleware file.  Do I understand your implementation correctly then, @ylvaselling ?

ylvaselling (approved):

PR #116, opened 2025-04-07
Handle "Focusable" setting for scene graph nodes

WeirdRubberDuck:
Do not show non-focusable nodes in origin/navigation menu list, and do not show a focus button in scene menu.

Update 2025-04-09: 
- Now it also adds a setting in the Navigation menu for whether we want to include non-focusable nodes or not. And a filter in the scene menu

To test it, either:
1. Set the visibility level to "everything" and set the "Focusable" settings for a Scene Graph Node under "Other"
2. Or set the `Focusable` setting under the GUI parameters in an asset file, like this:
image

This is how it will look in the UI as of now: 
image

### DIFF
--- a/src/hooks/sceneGraphNodes/util.ts
+++ b/src/hooks/sceneGraphNodes/util.ts
@@ -44,10 +44,6 @@ export function isSgnFocusable(uri: Uri, properties: Properties): boolean {
 }
 
 export function hasActiveFilters(filters: SceneGraphNodesFilters): boolean {
-  return (
-    filters.showHiddenNodes ||
-    filters.onlyFocusable ||
-    (filters.tags && filters.tags.length > 0) ||
-    false
-  );
+  const tagFilterIsActive = (filters.tags && filters.tags.length > 0) || false;
+  return filters.includeGuiHiddenNodes || filters.onlyFocusable || tagFilterIsActive;
 }
--- a/src/panels/OriginPanel/OriginPanel.tsx
+++ b/src/panels/OriginPanel/OriginPanel.tsx
@@ -3,7 +3,7 @@ import { Center, Group, SegmentedControl, Text, VisuallyHidden } from '@mantine/
 
 import { generateMatcherFunctionByKeys } from '@/components/FilterList/util';
 import { Layout } from '@/components/Layout/Layout';
-import { useSceneGraphNodes } from '@/hooks/sceneGraphNodes/sceneGraphNodes';
+import { useSceneGraphNodes } from '@/hooks/sceneGraphNodes/hooks';
 import { AnchorIcon, FocusIcon, TelescopeIcon } from '@/icons/icons';
 import { useAppSelector } from '@/redux/hooks';
 import { EngineMode, IconSize } from '@/types/enums';
@@ -28,8 +28,8 @@ export function OriginPanel() {
     return aimProp?.value !== anchorProp?.value && aimProp?.value !== '';
   });
 
-  const showNonFocusable = useAppSelector(
-    (state) => state.local.menus.navigation.showNonFocusable
+  const showOnlyFocusable = useAppSelector(
+    (state) => state.local.menus.navigation.onlyFocusable
   );
 
   const [navigationMode, setNavigationMode] = useState(
@@ -40,8 +40,8 @@ export function OriginPanel() {
 
   // @TODO (2024-04-08, emmbr): Expose these filters to the user? Could also include tags
   const searchableNodes = useSceneGraphNodes({
-    showHiddenNodes: false,
-    onlyFocusable: !showNonFocusable
+    includeGuiHiddenNodes: false,
+    onlyFocusable: showOnlyFocusable
   });
 
   const sortedSearchableNodes = useMemo(
--- a/src/panels/Scene/SceneTree/hooks.ts
+++ b/src/panels/Scene/SceneTree/hooks.ts
@@ -1,6 +1,6 @@
 import { useMemo } from 'react';
 
-import { useSceneGraphNodes } from '@/hooks/sceneGraphNodes/sceneGraphNodes';
+import { useSceneGraphNodes } from '@/hooks/sceneGraphNodes/hooks';
 import { sgnGuiOrderingNumber } from '@/hooks/sceneGraphNodes/util';
 import { useAppSelector } from '@/redux/hooks';
 import {
@@ -29,7 +29,7 @@ export function useSceneTreeData(filter: SceneTreeFilterSettings) {
 
   // First, filter the scene graph nodes based on the GUI filter settings
   const filteredSceneGraphNodes = useSceneGraphNodes({
-    showHiddenNodes: filter.showHiddenNodes,
+    includeGuiHiddenNodes: filter.includeGuiHiddenNodes,
     onlyFocusable: filter.onlyFocusable,
     tags: filter.tags
   });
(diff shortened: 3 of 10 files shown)


ylvaselling (changes requested):

WeirdRubberDuck:
Rewrote parts of the scene menu creation code to use the `useSceneGraphNodes` I created for this feature, so this PR became a bit bigger and maybe more confusing than intended.  Now, the Scene tree and the focus menu use the same code for their filtering. 

This means that we can apply filtering based on tags, and expose the settings for whether to show only focusable and/or hidden nodes even in the navigation menu, if we want to.  Do we?  And maybe it should then be a separate PR?

src/hooks/propertyOwner.ts
```diff
@@ -82,3 +84,47 @@ export function usePropertyOwnerVisibility(uri: Uri) {
...
+        .map((uri) => propertyOwners[uri])
+        .filter((node) => node !== undefined)
+        .filter((node) => {
+          const isHidden = isSgnHidden(node.uri, properties);
+          const isFocusable = isSgnFocusable(node.uri, properties);
+
+          if (!includeHidden && isHidden) {
+            return false;
+          }
+          if (!includeNonFocusable && !isFocusable) {
```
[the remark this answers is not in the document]
WeirdRubberDuck: Combined the hooks so the logic for this filtering is now only in one place :)

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
```diff
@@ -0,0 +1,86 @@
+import { useMemo } from 'react';
```
ylvaselling, line 1: I think maybe this file should be called "hooks" if we want to match the other pattern with "types" and "util"?

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
```diff
@@ -0,0 +1,86 @@
...
+} from './util';
+
+export function useIsSgnFocusable(uri: Uri): boolean {
+  return (
+    useAppSelector((state) => isSgnFocusable(uri, state.properties.properties)) || false
+  );
+}
+
+export function useSceneGraphNodes({
+  showHiddenNodes = false,
```
ylvaselling, line 21: Is it possible here to use the default settings that are stored in `src/panels/Scene/SceneTree`? Is that something we want?

src/hooks/sceneGraphNodes/util.ts
```diff
@@ -0,0 +1,54 @@
...
+  const isFocusable = properties[`${uri}.Focusable`]?.value as boolean | undefined;
+  return isFocusable || false;
+}
+
+export function hasActiveFilters(filters: SceneGraphNodesFilters): boolean {
+  return (
+    filters.showHiddenNodes ||
+    filters.onlyFocusable ||
+    (filters.tags && filters.tags.length > 0) ||
+    false
```
ylvaselling, line 52: Why the `|| false` ? It should not do anything here right?

src/panels/OriginPanel/OriginPanel.tsx
```diff
@@ -31,26 +28,26 @@ export function OriginPanel() {
     return aimProp?.value !== anchorProp?.value && aimProp?.value !== '';
   });
 
+  const showNonFocusable = useAppSelector(
+    (state) => state.local.menus.navigation.showNonFocusable
```
ylvaselling, line 32: Could we match this name to the other name? `onlyFocusable`?

src/panels/Scene/SceneTree/hooks.ts
```diff
@@ -109,15 +123,20 @@ function sceneTreeDataFromGroups(
...
+      treeData.push(groupData);
+    }
   });
 
   // Add the nodes without any group to the top level
   const nodesWithoutGroup = groups['/']?.propertyOwners || [];
   nodesWithoutGroup.forEach((uri) => {
     const owner = propertyOwners[uri];
-    if (owner === undefined) {
+    if (!owner || filteredPropertyOwners.indexOf(owner) === -1) {
```
ylvaselling, line 139:
```suggestion
    const ownerExists = filteredPropertyOwners.indexOf(owner) !== -1;
    if (!owner || !ownerExists) {
```

ylvaselling (approved):
Overall I think it looks great!! It's really nice to have the scenegraphnode utils not look so overwhelming! 🎉 Some minor comments.

Made me however think more about the file structure... Alternatives:

1. Util, types and hooks in one "hooks" folder like this. I think I would expect this kind of folder to be "encapsulated" in the hooks file, like the exports from types and util are only used in the hooks file. That is a bit difficult to maintain however since all exports are global... say that someone else will want to import an export from one of them in the future. I thought maybe it will be confusing to get a type from `hooks/scenegraphnodes/types`.
2.  An alternative is to call them the same name `sceneGraphNode.ts`, and place them at their top level respective folder, like `hooks`, `util`, and `types`. That might remove the encapsulating effect, but it has the benefit that the import paths will make sense and we also can add other scenegraphnode related things to the util for example. 

Now that i thought about this a bit more im leaning more towards 2... (i know i suggested 1 before sorry hehe). lmk what you think :)

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
WeirdRubberDuck: Hm, perhaps. However, it is also already located in the hooks folder, and the other files with hooks in this folder have a name that explains their usage. That's why I named it like this when I created it. 

But I guess now that this file is moved into a subfolder, it might make sense to name it hooks instead. Might need some discussion. What does @engbergandreas think about it?

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
WeirdRubberDuck: No, I don't think so. The deafult setting for the filtering could be different for different use cases. 

For example, in the scene menu we show non-focusable nodes per default, but in the navigation menu we don't

engbergandreas (approved):
Looking good 👍

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
```diff
@@ -0,0 +1,86 @@
...
+  onlyFocusable = false,
```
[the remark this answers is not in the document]
WeirdRubberDuck: I would personally prefer not to have the "show" in there at all in any of the variables.  This hook has nothing to do with "showing" things one its own  - it just filters the list

src/hooks/sceneGraphNodes/util.ts
WeirdRubberDuck: It was needed for the final line (the one checking the tags) to always return a boolean.  I rewrote it with a variable to make it clearer: 

```
  const tagFilterIsActive = (filters.tags && filters.tags.length > 0) || false;
  return filters.includeGuiHiddenNodes || filters.onlyFocusable || tagFilterIsActive;
```

src/hooks/sceneGraphNodes/sceneGraphNodes.ts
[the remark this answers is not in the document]
WeirdRubberDuck: Renamed `showHiddenNodes` to `includeHiddenNodes`

PR #121, opened 2025-04-16
Feature/profile updates

ylvaselling:
Changes:

* Sends and show profile name in top bar
* Creates a default taskbar function
* Adds reset button to view menu
* Also realized view menu wasn't showing while loading so fixed that too

Needs the latest commit in the engine ui branch

### DIFF
--- a/src/components/Input/BoolInput.tsx
+++ b/src/components/Input/BoolInput.tsx
@@ -1,0 +1,33 @@
+import { JSX } from 'react';
+import { Checkbox, Group } from '@mantine/core';
+
+import { Label } from '../Label/Label';
+
+interface Props {
+  name: string;
+  description: string | JSX.Element;
+  value: boolean;
+  setValue: (value: boolean) => void;
+  readOnly?: boolean;
+}
+
+export function BoolInput({
+  value,
+  setValue,
+  name,
+  description,
+  readOnly = false
+}: Props) {
+  return (
+    <Group gap={'xs'} wrap={'nowrap'}>
+      <Checkbox
+        checked={value}
+        onChange={(event) => setValue(event.currentTarget.checked)}
+        onKeyDown={(event) => event.key === 'Enter' && setValue(!value)}
+        disabled={readOnly}
+        aria-label={`Toggle ${name}`}
+      />
+      <Label name={name} description={description} readOnly={readOnly} />
+    </Group>
+  );
+}
--- a/src/components/Input/BoolInput.tsx
+++ b/src/components/Input/BoolInput.tsx
@@ -25,7 +25,7 @@ export function BoolInput({
         onChange={(event) => setValue(event.currentTarget.checked)}
         onKeyDown={(event) => event.key === 'Enter' && setValue(!value)}
         disabled={readOnly}
-        aria-label={`Toggle ${name}`}
+        aria-label={name}
       />
       <Label name={name} description={description} readOnly={readOnly} />
     </Group>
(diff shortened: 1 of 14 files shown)


engbergandreas (approved):
Looks good, nice catch with the view menu not working I was thinking about it some day ago and thought it was weird

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -29,11 +36,18 @@ export function ViewMenu() {
...
+    Object.entries(defaultTaskbar).forEach(([key, value]) => {
+      dispatch(setMenuItemVisible({ id: key, visible: value }));
+    });
   }
-  const { Options: userLevelOptions } =
-    description.additionalData as AdditionalDataOptions;
+  const additionalData = description
+    ? (description.additionalData as AdditionalDataOptions)
+    : undefined;
+  const userLevelOptions = additionalData?.Options;
```
WeirdRubberDuck, line 50:
```suggestion

  const additionalData = description?.additionalData as AdditionalDataOptions | undefined;
  const userLevelOptions = additionalData?.Options;
```

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -50,7 +64,14 @@ export function ViewMenu() {
         withinPortal={false}
         closeOnItemClick={false}
       >
-        <Menu.Label>Toggle Task Bar Items</Menu.Label>
+        <Menu.Label pr={0}>
+          <Group justify={'space-between'}>
+            Toggle Task Bar Items{' '}
+            <Button size={"xs"} onClick={() => resetTaskbar()}>
```
WeirdRubberDuck, line 70:
```suggestion
            <Button size={"xs"} onClick={resetTaskbar}>
```

src/panels/Menu/TopMenuBar/TopMenuBar.tsx
```diff
@@ -1,35 +1,42 @@
...
         bg={alpha('var(--mantine-color-dark-9)', 0.9)}
         justify={'space-between'}
-        pb={'md'}
+        align={'center'}
         style={{
           whiteSpace: 'nowrap'
         }}
       >
-        <Group flex={'0 0 auto'}>
+        <Group flex={flex}>
```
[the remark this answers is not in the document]
WeirdRubberDuck: Unfortunately, it breaks when made small 
image

WeirdRubberDuck (changes requested):
Looks great, but we need to figure out what to do about this situation when the width of the window changes: 
image


Love the placement of the reset button for the taskbar!

ylvaselling:
Resolved the issue of the friction controls and top menu wrapping.

While I was at it I saw that the flight control panel had different styled input than our properties, so I made a component for bool input and another component for our label that doesn't need a uri and a property to make it easier to have a unified look in the ui.

WeirdRubberDuck (approved):

src/components/Input/BoolInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+  readOnly = false
+}: Props) {
+  return (
+    <Group gap={'xs'} wrap={'nowrap'}>
+      <Checkbox
+        checked={value}
+        onChange={(event) => setValue(event.currentTarget.checked)}
+        onKeyDown={(event) => event.key === 'Enter' && setValue(!value)}
+        disabled={readOnly}
+        aria-label={`Toggle ${name}`}
```
WeirdRubberDuck, line 28: I think we might want to put this on each individual component instead, to be able to make sure that it always makes sense. 

Now there will be some case where the aria label reads "Toggle toggle ..."  due to how the bool inputs are named

We could make it a required property of this component? :)

ylvaselling: Unfortunately, the aria-label can not be passed as a prop unless we rename it to ariaLabel, which would be a new way to pass an aria-label in this repo and I'm not a huge fan of that tbh... Making the aria-label the name for now :) I think that should be good enough

engbergandreas (approved):

PR #122, opened 2025-04-16
Add within cef flag that is set from C++

ylvaselling:
Checks if webpage is in browser or cef and sets the flightcontroller panel visibility accordingly

### DIFF
--- a/src/cef-globals.d.ts
+++ b/src/cef-globals.d.ts
@@ -1,0 +1,13 @@
+// Export an empty object to ensure this file is treated as a module.
+// Required when using declare global {} blocks so that TypeScript treats the file correctly.
+export {};
+
+// Declare global types to extend the Window interface
+declare global {
+  interface Window {
+    // Add a custom property to the Window object to indicate if it's within CEF.
+    // This is set within the CEF environment so will be true within CEF and undefined otherwise.
+    // This is useful for checking if the app is running in a CEF context or not.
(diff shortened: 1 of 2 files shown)


engbergandreas (approved):
sweet!

src/cef-globals.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// Required when using declare global {} blocks so that TypeScript treats the file correctly.
+export {};
+
+// Declare global types to extend the Window interface
+declare global {
+  interface Window {
+    // Add a custom property to the Window object to indicate if it's within CEF.
+    // This is set within the CEF environment so will be true within CEF and undefined otherwise.
+    // This is useful for checking if the app is running in a CEF context or not.
+    IsWithinCEF: boolean | undefined;
```
WeirdRubberDuck, line 11: Copilot already commented on this, but how about using camelcase for the variable name?  Especially since we've started to remove cases where variables start with a capital letter, because we don't want it ;)

WeirdRubberDuck: That is
```suggestion
    isWithinCEF: boolean | undefined;
```

WeirdRubberDuck (approved):
Very nice!  

We'll have to see if we get any backlash from only providing the "remote UI", i.e. with the flight controller, in the browser . But I don't think so :) 

It's fewer things to keep track of, which is always nice.

PR #123, opened 2025-04-17
Feature/add checkboxes for search

ylvaselling:
Adds a filterlist component for checking which keys should be searched in the object.

Added it to all places we have the `generateMatcherFunction`


src/panels/Scene/SceneTree/SceneTree.tsx
```diff
@@ -33,6 +34,11 @@ export function SceneTree() {
     sceneTreeFilterDefaults
   );
 
+  const { allowedKeys, toggleKey, selectedKeys } = useKeySettings<SceneTreeNodeData>({
+    guiPath: false,
+    label: true
```
WeirdRubberDuck, line 39: Reorder so the default enabled one is always first? 

Also, here I think we should make it so that the UI says "Name" instead of Label. That will be a lot less confusing to the user.

src/panels/Scene/SceneTree/SceneTree.tsx
```diff
@@ -71,6 +77,7 @@ export function SceneTree() {
       <Group justify={'space-between'} gap={'xs'} mr={'xs'}>
         <FilterList.InputField placeHolderSearchText={'Search for a node...'} flex={1} />
         <SceneTreeFilters setFilter={setFilter} filter={filter} />
+        <FilterList.SearchSettingsMenu keys={allowedKeys} setKey={toggleKey} />
```
WeirdRubberDuck, line 80: I would probably put this before the filter, since this is more closely related to the search field that the filter settings are

src/panels/Scene/SceneTree/SceneTree.tsx
```diff
@@ -102,7 +109,8 @@ export function SceneTree() {
         renderElement={(node: SceneTreeNodeData) => (
           <SceneTreeNodeContent key={node.value} node={node} expanded={false} />
         )}
-        matcherFunc={generateMatcherFunctionByKeys(['label', 'guiPath'])} // For now we just use the name
+        // For now we just use the name
```
WeirdRubberDuck, line 112:
```suggestion
```

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -48,50 +49,53 @@ export function SceneTreeFilters({ setFilter, filter }: Props) {
           </ActionIcon>
         </Menu.Target>
         <Menu.Dropdown maw={'300px'}>
```
WeirdRubberDuck, line 51:
```suggestion
        <Menu.Dropdown maw={'330px'}>
```
to fix that info box jumps down to next level

src/panels/SkyBrowserPanel/ImageList/ImageList.tsx
```diff
@@ -17,14 +19,18 @@ interface Props {
 // Memoizing this as it is very expensive
 export const ImageList = memo(function ImageList({ imageList, noImagesDisplay }: Props) {
   const { width } = useWindowSize();
+  const { allowedKeys, toggleKey, selectedKeys } = useKeySettings<SkyBrowserImage>({
+    collection: true,
+    name: true
```
WeirdRubberDuck, line 24: I would opt for always just including the name per default? 

Or at least, always put the name as the first key in the list:
```suggestion
    name: true,
    collection: true
```

src/components/FilterList/SearchSettingsMenu/FilterListSearchSettingsMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <Menu position={'right-start'} withArrow closeOnItemClick={false}>
+      <Menu.Target>
+        <ActionIcon flex={'0 0 auto'}>
+          <SettingsIcon />
+        </ActionIcon>
+      </Menu.Target>
+      <Menu.Dropdown maw={'300px'}>
+        <Menu.Label>Search these keys</Menu.Label>
+        <Stack p={'xs'}>
+          {Object.entries(keys).map(([key, enabled]) => (
```
WeirdRubberDuck, line 22: Now it's possible to uncheck all the checkboxes => there will be no result in the search and it's hard to know why. 

Should we maybe 
1. Prevent unchecking all checkboxes? 
2. or, at least show some type of warning if no checkbox is checked? The settings icon could change to a warning, for example, and some warning text could be shown in the menu

src/components/FilterList/SearchSettingsMenu/FilterListSearchSettingsMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface Props<T> {
+  keys: Record<keyof T, boolean>;
+  setKey: (key: keyof T, enabled: boolean) => void;
+}
+
+export function FilterListSearchSettingsMenu<T>({ keys, setKey }: Props<T>) {
+  return (
+    <Menu position={'right-start'} withArrow closeOnItemClick={false}>
+      <Menu.Target>
+        <ActionIcon flex={'0 0 auto'}>
```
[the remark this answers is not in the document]
ylvaselling: It is needed because we don't want this to expand on flex. For example, if this action icon is placed in a flex box, it will default to grow to the same size as the other elements in that flexbox. However, we always want action icons to only take up the space that the icon actually requires. That is why the flex is needed

src/components/FilterList/SearchSettingsMenu/hook.ts
```diff
@@ -0,0 +1,28 @@
...
+export function useKeySettings<T>(keys: Partial<Record<keyof T, boolean>>) {
+  const [enabledKeys, setEnabledKeys] = useState<Record<keyof T, boolean>>(
+    Object.entries(keys).reduce(
+      (acc, [key, value]) => {
+        acc[key as keyof T] = value as boolean; // Use provided value, otherwise default to true
+        return acc;
+      },
+      {} as Record<keyof T, boolean>
+    )
+  );
```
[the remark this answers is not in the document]
ylvaselling: Here is an article explaining when to use Record:
https://medium.com/@dhrumitpatell/understanding-the-difference-between-record-and-object-in-typescript-a-beginners-guide-1e491a9f8182

TLDR; use record when the types are the same for key and for value in an object

src/components/FilterList/SearchSettingsMenu/hook.ts
```diff
@@ -0,0 +1,28 @@
...
+      }
+    },
+    [keys]
+  );
+
+  const selectedKeys = Object.entries(enabledKeys)
+    .filter(([, value]) => value)
+    .map(([key]) => key) as Array<keyof T>;
+
+  return { allowedKeys: enabledKeys, toggleKey, selectedKeys } as const;
```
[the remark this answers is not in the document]
ylvaselling: It sets the object as readonly which means you shouldn't be allowed to mutate it. I'm not sure if it is strictly necessary here but why not🤷

src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/EarthPanel.tsx
```diff
@@ -194,7 +200,10 @@ export function EarthPanel({ currentAnchor }: Props) {
           {places.length > 0 ? (
             <ResizeableContent defaultHeight={250}>
               <FilterList>
-                <FilterList.InputField placeHolderSearchText={'Filter search'} />
+                <Group>
+                  <FilterList.InputField placeHolderSearchText={'Filter search'} />
+                  <FilterList.SearchSettingsMenu keys={allowedKeys} setKey={toggleKey} />
```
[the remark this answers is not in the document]
ylvaselling: That may be true, but that is outside the scope of this PR :) Make a ticket in notion or an issue on github?

src/components/FilterList/SearchSettingsMenu/hook.ts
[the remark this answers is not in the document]
ylvaselling: As for the default state, this was forgotten code by me and we could just set the default state as keys

src/panels/SkyBrowserPanel/ImageList/ImageList.tsx
ylvaselling: I ordered them alphabetically which makes sense to me. We can not assume that the objects will contain a key called name so we can't make that generalization

src/panels/Scene/SceneTree/SceneTree.tsx
ylvaselling: I ordered them alphabetically which makes sense to me. I think both ways are fine in terms of ordering but we should probably stay consistent and I think alphabetically is fine since it is less complex than ordering first 1) true / false 2) alphabetically. Makes it easier to change too and I don't really see that it is an issue

engbergandreas (approved):

src/components/FilterList/SearchSettingsMenu/FilterListSearchSettingsMenu.tsx
```diff
@@ -0,0 +1,35 @@
...
+      </Menu.Target>
+      <Menu.Dropdown maw={'300px'}>
+        <Menu.Label>Search these keys</Menu.Label>
+        <Stack p={'xs'}>
+          {Object.entries(keys).map(([key, enabled]) => (
+            <Group key={key}>
+              <Checkbox
+                label={camelCaseToRegularText(key)}
+                checked={enabled as boolean}
+                onChange={(event) => setKey(key as keyof T, event.currentTarget.checked)}
```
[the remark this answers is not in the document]
WeirdRubberDuck: I see. 

Maybe this is worth having a short discussion about (IRL). I fail to understand why we need `keyof T` instead of using just `string`. do you think we need the flexibility of specifying keys of different types?  

Using just `string` would possibly make the code more straightforward?

src/components/FilterList/SearchSettingsMenu/hook.ts
```diff
@@ -0,0 +1,55 @@
...
+      if (key in keys) {
+        setAllowedSearchKeys((prev) => ({ ...prev, [key]: enabled }));
+      }
+    },
+    [keys]
+  );
+
+  const selectedSearchKeys = Object.entries(allowedSearchKeys)
+    .filter(([, value]) => value)
+    .map(([key]) => key) as Array<keyof T>;
```
WeirdRubberDuck, line 48: Related to the previous comment. Could we actually just use `string` instead of `keyof T` ?

Also, this code seem to rely on that `T` is an object type, but now it is technically allowed to be anything. 

I think one way to make sure that it's an object is to specify it as `T extends object` ,i.e., `function useSearchKeySettings<T extends object>`. However, maybe even better would be to specify the expected object more clearly (maybe that it's supposed to have `string` keys for example?

Reducing the "anything is allowed"-ness of the situation her would make the component both easier to use and understand, I think

src/components/FilterList/SearchSettingsMenu/hook.ts
```diff
@@ -0,0 +1,55 @@
+import { useCallback, useState } from 'react';
+
+/**
+ * A custom React hook that manages a set of keys and their enabled/disabled states.
+ *
+ * @template T - The type of the object whose keys are being managed.
+ *
+ * @param keys - An object containing a partial mapping of keys of type `T` to boolean values,
+ *               indicating whether each key is enabled or disabled by default.
```
WeirdRubberDuck, line 9: "keys of type T" here sounds like the keys are supposed to be of the same type as the object, which is not quite correct. 

This docs would also be simpler if we limit the keys to being strings, and restrict `T` to only be a valid object type using typescript. YThen, those type details wouldn't be needed in the documentation here

WeirdRubberDuck (commented):
Essentially an approve, but I had some comments on the template and `keyof T` situation. Let's discuss

WeirdRubberDuck (approved):

PR #124, opened 2025-04-17
Feature/scriptlog

engbergandreas:
Add the script log as a panel

For now, it's fetching all the data through the Lua API, which will clog the log with some extra entries. In the future, we should write a topic for this so that we pass the data through that, but that also requires a subscription to changes in files, so it's a bit of a bigger rewrite. 

Also, for future changes: Scripts that are multiple lines e.g., enabling/disabling sgn should be grouped as one entry


engbergandreas:
If you have any idea on why the text wrapping/overflow works in the browser but not in cef I would gladly take pointers
Browser: 
image
OpenSpace
image

ylvaselling:
> If you have any idea on why the text truncate/wrapping/overflow works in the browser but not in cef I would gladly take pointers Browser: obs, the `setPropertyValue` has been expanded here but the first two entries you can see are correctly truncated image OpenSpace image

It might be a firefox vs chromium issue? Have you tried using Chrome?

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+import { ScriptLogEntry } from './ScriptLogEntry';
+
+// TODO (anden88 2025-04-17): Remove when lists from OpenSpace don't have the
+// ['1']: value, ['2']: value format
+type FileLines = Record<string, string> | undefined;
+
+export function ScriptLogPanel() {
+  const luaApi = useOpenSpaceApi();
+  const [scriptLogEntries, setScriptLogEntries] = useState<string[]>([]);
```
ylvaselling, line 17: state goes above custom hooks

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+    setScriptLogEntries(
+      Object.values(data)
+        .filter((script) => script)
+        .reverse()
+    );
+  }, [luaApi]);
+
+  useEffect(() => {
+    fetchScriptLogEntries();
+  }, [fetchScriptLogEntries]);
```
ylvaselling, line 41: in other places where we fetch we have placed the function within the hook itself. might be nicer for consistency 🤷

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,76 @@
...
+  function preprocessScriptEntry(entry: string): string {
+    // @TODO (anden88, 2025-04-17): do we want to rewrite this using regex instead?
+    // return entry.replace(/[.,()""]/g, ' ');
+    return entry
+      .replaceAll('.', ' ')
+      .replaceAll(',', ' ')
+      .replaceAll('"', ' ')
+      .replaceAll('(', ' ')
+      .replaceAll(')', ' ');
+  }
```
ylvaselling, line 52: Make this a more generic util function where we pass an array of characters `[".", ",", ")"]` etc and we can loop through it and return the finished string? I think other components might find that useful too

src/panels/LogPanel/ScriptLogPanel.tsx
ylvaselling, line 17: Maybe we want default state to be `undefined` instead of `[]` since I guess it is theoretically possible (but probably will never happen, at least not now) with an empty scriptlog? Just seeing it could potentially show loading blocks when the log is just empty

ylvaselling (changes requested):
🎉 

* Style the scripts as code?
https://mantine.dev/core/code/
* Somehow the refresh symbol looks like a "undo" arrow. I think refresh goes clockwise? 
image

src/panels/LogPanel/ScriptLogPanel.tsx
engbergandreas: As discussed irl, we are rewriting regex with less power at that point. Opting for a regex command instead since I've also added additional characters to improve the search further

src/panels/LogPanel/ScriptLogEntry.tsx
```diff
@@ -0,0 +1,46 @@
...
+            onClick={runScript}
+            variant={'light'}
+            color={'green'}
+            aria-label={'Rerun script'}
+          >
+            <RerunScriptIcon />
+          </ActionIcon>
+        </Tooltip>
+      </Group>
+    </Group>
```
WeirdRubberDuck, line 44: The suggestion below addresses: 
- Reduce text size and run button size (very large on my 1080p screen)
- Remove one redundant group, and reduce gap between text and buttons
- Fix awkward newlines when scripts are expanded

```suggestion
    <Group wrap={'nowrap'} justify={'space-between'} align={'start'} gap={'xs'}>
      <Code color={'dark.7'} w={'100%'}>
        <Text
          truncate={expanded ? undefined : 'end'}
          size={'xs'}
          onClick={() => setExpanded(!expanded)}
          style={{ cursor: 'pointer', wordBreak: 'break-all' }}
        >
          {script}
        </Text>
      </Code>
      <CopyToClipboardButton value={script} />
      <Tooltip label={'Rerun script'}>
        <ActionIcon
          onClick={runScript}
          variant={'light'}
          size={'sm'}
          color={'green'}
          aria-label={'Rerun script'}
        >
          <RerunScriptIcon />
        </ActionIcon>
      </Tooltip>
    </Group>
```

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const luaApi = useOpenSpaceApi();
+
+  const fetchScriptLogEntries = useCallback(async () => {
+    // eslint-disable-next-line no-template-curly-in-string
+    const fileName = await luaApi?.absPath('${LOGS}/ScriptLog.txt');
+
+    if (!fileName) {
+      return;
+    }
+    const data = (await luaApi?.readFileLines(fileName)) as FileLines;
```
WeirdRubberDuck, line 28: Since you will know exactly what these lines look like, how about filtering them away from the output for now?

WeirdRubberDuck: I know we want to make a topic for it eventually, but who knows when that will happen so we might just as well do this simple check now

src/panels/LogPanel/ScriptLogEntry.tsx
```diff
@@ -0,0 +1,46 @@
...
+        </Text>
+      </Code>
+      <Group wrap={'nowrap'} gap={'xs'}>
+        <CopyToClipboardButton value={script} />
+        <Tooltip label={'Rerun script'}>
+          <ActionIcon
+            onClick={runScript}
+            variant={'light'}
+            color={'green'}
+            aria-label={'Rerun script'}
```
WeirdRubberDuck, line 38: Nice with the aria label. However, do we also need some type of "visually hidden" text that says what the script actually is? Now I don't think a screen reader will get this information

WeirdRubberDuck: It might also be confusing for screen reader users now since the button for each line says exactly the same thing in the aria label. 

Worth some consideration

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  useEffect(() => {
+    fetchScriptLogEntries();
+  }, [fetchScriptLogEntries]);
+
+  function preprocessScriptEntry(entry: string): string {
+    // Replace special characters with spaces. The '+' ensures repeated punctuation is
+    // replaced by a single space instead of many.
+    return entry.replace(/[.,()[\]{}"'\\/]+/g, ' ');
+  }
```
WeirdRubberDuck, line 49: "preprocess" does not seem like the right word here, IMO. Also, it does not say anything about what the function does, which is to "sanitize" the string

How about using "sanitize" or "clean" instead? Also, here, the `entry` is actually just a string for the script, so make that clearer in the name: 

```suggestion
  function sanitizeScript(script: string): string {
    // Replace special characters with spaces. The '+' ensures repeated punctuation is
    // replaced by a single space instead of many
    return script.replace(/[.,()[\]{}"'\\/]+/g, ' ');
  }
```

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,73 @@
...
+        </Tooltip>
+        <FilterList.InputField placeHolderSearchText={'Search for a script'} flex={1} />
+      </Group>
+      <FilterList.SearchResults
+        data={scriptLogEntries ?? []}
+        renderElement={(entry) => <ScriptLogEntry script={entry} />}
+        matcherFunc={(data, searchString) => {
+          const preprocessedData = preprocessScriptEntry(data);
+          return wordBeginningSubString(preprocessedData, searchString);
+        }}
```
WeirdRubberDuck, line 67: With the change suggested above, this could be: 
```suggestion
        matcherFunc={(script, searchString) =>
          wordBeginningSubString(sanitizeScript(script), searchString)
        }
```
which seems more readable to me (I don't have to go and check what the preprocessing step is in the function)

src/panels/LogPanel/ScriptLogPanel.tsx
WeirdRubberDuck: With this new more descriptive name, it could also be moved to a utility file, if we want

WeirdRubberDuck (approved):
Very nice!  

One thought I had now when testing it (that might involve a redesign, however) is that I was missing the ability to select multiple scripts and run/copy all the selected ones at once.  I'm imagining something similar to the profile editor's additional scrips menu: 
image

If we want to be able to select multiple lines there, I imagine we want to be able to do it here too. Now it's a bit weird that the way it works is different

Don't know how hard it would be, and the existing panel is already very useful. So, maybe it can be skipped. Redesigning it to look more like the view in the profile editor could make it less cluttered, however, because if we can select a line to run, only one copy and run button would really be needed. This would also help with the comments I had related to screen readers.

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -31,46 +37,133 @@ export function ScriptLogPanel() {
...
+
+  function runSelectedScripts() {
+    if (scriptLogEntries === undefined || !luaApi) {
+      return;
+    }
+    // TODO (anden88 2025-04-23): In what order do we want to run the scripts? The scripts
+    // are ordered latest first (top-bottom) but they were actually executed bottom up...
+    const scriptsToRun = Array.from(selectedScripts)
+      .sort((a, b) => a - b)
+      .map((index) => scriptLogEntries[index]);
```
ylvaselling, line 93: Yeah, this is probably a bit confusing for a user. One way to get around the issue is to have a tooltip on the rerun scripts button?

I think we definitely want to run the scripts bottom-up.

Another thing to consider is that now, with the checkboxes, it takes three clicks to re-run two scripts whereas re-running two scripts with the regular button takes two clicks. Just a thought

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -31,46 +37,133 @@ export function ScriptLogPanel() {
...
+            renderElement={(entry, index) => (
+              <ScriptLogEntry
+                script={entry}
+                index={index}
+                isSelected={selectedScripts.has(index)}
+                onToggleSelection={handleToggleSelection}
+              />
+            )}
+            matcherFunc={(script, searchString) => {
+              return wordBeginningSubString(sanitizeScript(script), searchString);
```
ylvaselling, line 160: nice function name :)

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -31,46 +37,133 @@ export function ScriptLogPanel() {
...
+        <Group wrap={'nowrap'} justify={'space-between'} align={'start'} gap={'xs'}>
+          <Group gap={'xs'}>
+            <Tooltip label={'Run all the selected scripts in sequential order'}>
+              <Button
+                onClick={runSelectedScripts}
+                rightSection={<RerunScriptIcon />}
+                disabled={selectedScripts.size === 0}
+              >
+                Run Scripts
+              </Button>
```
ylvaselling, line 112: Can we make this styled the same as the other run scripts buttons? I.e. green and subtle 🙏

ylvaselling (approved):

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+  const luaApi = useOpenSpaceApi();
+
+  const fetchScriptLogEntries = useCallback(async () => {
+    // eslint-disable-next-line no-template-curly-in-string
+    const fileName = await luaApi?.absPath('${LOGS}/ScriptLog.txt');
+
+    if (!fileName) {
+      return;
+    }
+    const data = (await luaApi?.readFileLines(fileName)) as FileLines;
```
WeirdRubberDuck, line 33: Nitpicky
```suggestion

    const data = (await luaApi?.readFileLines(fileName)) as FileLines;
```

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -0,0 +1,198 @@
...
+              This script log has been cleaned for ease of use. Some entries may have been
+              filtered out unintentionally.
+              <Text>
+                Entries are listed from latest to oldest (top to bottom). Running multiple
+                scripts at once will run them in the order bottom to top.
+              </Text>
+            </Text>
+            <Text>
+              To view the complete log or previous runs, check the{' '}
+              <Text fs={'italic'} fw={500} display={'inline-block'}>
```
WeirdRubberDuck, line 154:
```suggestion
              <Text fs={'italic'} fw={500} span>
```

WeirdRubberDuck (approved):
LOVE it! :D 

One thought - do we want the "Run" button to also reload the script log?

PR #125, opened 2025-04-22
Feature/improve vectors

ylvaselling:
Some improvements to the vectors:

* Changes Slider to components
* Adds sliders to vector properties
* Adds labels to vectors
* Better line breaking of the input fields
* Adds max columns optional property to dynamic grid

image


src/components/Property/Types/VectorProperty/ViewOptions/VectorDefaultView.tsx
```diff
@@ -10,19 +12,29 @@ interface Props {
...
 
   const { MinimumValue: min, MaximumValue: max, SteppingValue: step } = additionalData;
+  // @TODO (2025-03-03, emmbr) This should be handled a better way... This is a bit of a
+  // hack and the max value is just arbitrarily chosen
+  const maxAllowedExtentForSlider = 1e12;
+  const shouldShowSlider = max.some((value, i) => {
+    // When no min/max is set, the marks for the slider cannot be nicely computed
+    const extent = value - min[i];
+    return isFinite(value - min[i]) && extent < maxAllowedExtentForSlider;
+  });
```
WeirdRubberDuck, line 37: Looks like this is copied from the numeric property

Should we put this in some type of utility function instead? So that, when we address it, we will have to deal with both places at once?

WeirdRubberDuck (commented):
I like this conceptually, and in cases with up to three values it looks ok (maybe the sliders look a bit cluttery with that many ticks but that might just be me)

However, they do not work great for colors IMO, especially not colors with 4 values:
image
 
There was already limited horizontal space in this case, and with the labels the space is reduced even more

Never thought I'd miss the old sliders but I kind of do 😓 Don't know what a good redesign would be, but potentially we could redesign the view for colors entirely to make it a bit cleaner?  Maybe we should get rid of the number inputs entirely and just use one of the Mantine color input components?

WeirdRubberDuck (changes requested):
Let's discuss the color view when given a chance! 

I really like replacing the view with the input, but now we've lost the feature of specifying colors in HEX , HSL or 0-1 RGB formats. We should tweak it a little to bring back that flexibility

That is, the functionality that lives in the `ColorPicker` and `ColorEdit` components (which are now no longer used and could be retired, i.e. removed, once we've redesigned the input for colors)

Also:
- setting the alpha value of vec4 colors does not work currently. 
- when discussing the solution above, we need to make sure that inputting invalid color values is handled in a meaningful way

WeirdRubberDuck:
It is currently also possible to accidentally break the color properties

Steps to reproduce: 
* Open `Settings -> Lua Console`
* Mark the entire color string (as if trying to delete it and rewrite it) of e.g. `Background Color`
image

* Press backspace (as if to delete the test)
* Houston, we have  problem

image

src/components/Input/NumericInput/NumericSlider/util.ts
```diff
@@ -0,0 +1,28 @@
...
+ * ```typescript
+ * const isValid = validSliderExtent(0, 100); // true
+ * const isInvalid = validSliderExtent(0, Infinity); // false
+ * ```
+ */
+export function validSliderExtent(min: number, max: number) {
+  const extent = max - min;
+  // When no min/max is set, the marks for the slider cannot be nicely computed
+  return isFinite(extent) && extent < maxAllowedExtentForSlider;
+}
```
WeirdRubberDuck, line 28: This comment seems too elaborate compared to the function it documents. Not sure we want to maintain oxygen comments written to this extent? Especially since we don't have them this elaborated even in the main code base. 

How about simplifying it a bit to match what we do in OpenSpace: 

```suggestion
/**
 * Determines whether the given slider extent is valid based on the provided minimum and maximum values. 
 * If no `min` or `max` values are set, the slider marks cannot be computed properly.
 *
 * @param min - The minimum value of the slider.
 * @param max - The maximum value of the slider.
 *
 * @returns `true` if the extent (difference between max and min) is finite and less than the allowed maximum extent for the slider; otherwise, returns `false`.
 */
export function validSliderExtent(min: number, max: number) {
  const extent = max - min;
  // When no min/max is set, the marks for the slider cannot be nicely computed
  return isFinite(extent) && extent < maxAllowedExtentForSlider;
}
```

WeirdRubberDuck: Please also make sur ethat the comment adheres to outer max line width limitation.

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
```diff
@@ -1,9 +1,28 @@
...
 import { AdditionalDataVectorMatrix } from '@/types/Property/propertyTypes';
+import { openspaceColorToRgba, toOpenspaceColor } from '@/util/colorHelper';
 
-import { VectorDefaultView } from './VectorDefaultView';
+function validateInput(
+  value: number[],
+  additionalData: AdditionalDataVectorMatrix,
+  isInt: boolean
+) {
+  if ((value.length !== 3 && value.length !== 4) || isInt) {
```
WeirdRubberDuck, line 16:
```suggestion
  if (value.length !== 3 || value.length !== 4 || isInt) {
```
?

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
```diff
@@ -20,38 +39,71 @@ export function ColorView({
...
+
+  validateInput(value, additionalData, isInt);
+
+  function isValidColorString(colorString: string): boolean {
+    const newValues = colorString.split(',').map((v) => parseFloat(v));
+    const result =
+      newValues.length === value.length &&
+      newValues.every((v) => !isNaN(v) && isNumber(v));
+
+    return result;
```
WeirdRubberDuck, line 59:
```suggestion
    return (
      newValues.length === value.length &&
      newValues.every((v) => !isNaN(v) && isNumber(v))
    );
```
`result` is redundant here and to see what the function returns one has to go up to the variable anyways

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
```diff
@@ -20,38 +39,71 @@ export function ColorView({
...
+  }
+
+  function onColorPickerChange(color: RGBA) {
+    const newValue = toOpenspaceColor(color, hasAlpha);
+    setPropertyValue(newValue);
+
+    setIsError(!isValidColorString(newValue.join(', ')));
```
WeirdRubberDuck, line 66: How about moving the `.join(', ')` part into the `isvalidColor` function? 

Or, better, remove it completely? Seems like the first thing we do in the function is to split based on the `,` again :)

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
```diff
@@ -20,38 +39,71 @@ export function ColorView({
...
+      onBlur={handleBlur}
+      readOnly={readOnly}
+      rightSection={
+        isOutsideRange && (
+          <WarningIcon
+            tooltipText={`Value outside range [${additionalData.min}, ${additionalData.max}]`}
+          />
+        )
+      }
+      placeholder={'Example: 1, 0.5, 0.2'}
```
WeirdRubberDuck, line 105: How about inclusing in the placeholder whether the input string should include alpha or not? 
```suggestion
      placeholder={`RGB${hasAlpha ? 'A' : ''}: 1, 0.5, 0.2${hasAlpha ? ', 0.5' : ''}`}
```
image
image

src/components/Property/Types/VectorProperty/ViewOptions/VectorDefaultView.tsx
```diff
@@ -32,20 +37,48 @@ export function VectorDefaultView({
...
-          min={min[i]}
-          max={max[i]}
-          step={step[i]}
-          allowDecimal={!isInt}
-          onEnter={(newValue) => setValue(i, newValue)}
-        />
+        >
+          <Text c={'dimmed'} mt={5} flex={0} size={'sm'}>
+            {labels[i]}
+          </Text>
```
WeirdRubberDuck, line 57: IIRC we decided to comment out the code for the labels for now?

WeirdRubberDuck (changes requested):
Love it! Some things to fix: 

1. It is now possible to set an empty color.  The value should reset on blur
2. It should also reset the value on ESC :) 
3.  Not sure if part of this PR - but the color input is not accessible using keyboard. Make an issue?  IIRC we discussed using "focusTrap" for this popup (it also has to be closed on click and cannot be closed on escape, after changing a color)
4.  After entering an invalid color in text (e.g. `0.8, 0.2, -132123, 1`) and then setting a valid color using the color edit, the string value is not changed

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
ylvaselling: Will always evaluate to true so not correct in this case

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
ylvaselling: The function is used in four places, where two of them are strings like this so I thinks it's better to keep it the way it is

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
ylvaselling: Maybe this is a bit confusing? You might think you need to write "RGBA". But i agree that the placeholder should have the fourth number

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
WeirdRubberDuck: Correct!  sorry for the noise :)

src/components/Property/Types/VectorProperty/ViewOptions/ColorView.tsx
WeirdRubberDuck: I get your point, and maybe it's obvious to the user what each value is

ylvaselling:
Fixed issues with focus and on blur behavior

WeirdRubberDuck (approved):
Nice! There are still some weirdness when entering invalid colors, but that's expected. Imortantly, I couldn't get into a situation that wasn't recoverable, which is great!  :)

engbergandreas (approved):
Have not had time to test it but did a quick look through and it looks good. Also since many of these components will now have translated texts, just make sure they are correctly merged with master and add any missing ones

PR #126, opened 2025-04-25
Feature/unify task bar

ylvaselling:
* Removes `enabled` property for menu items
* Makes all task menu bar buttons act the same (never disabled - shows loading blocks if needed)
* Adds loading blocks to panels that need it


WeirdRubberDuck (changes requested):
It was convenient to try the loading states, since the UI in this branch does not load correctly when used with the later `2025-ui-rewrite` branch 😉 The loading states look great

Some TODOs
1. We should also make sure to remove the "enabled" settings from the taskbar settings in the engine right away, so that that is not forgotten. 
2. Let's add loading blocks to the Settings menu and Actions as well? 
3. I don't think the loading of the `default_ui_panels.json` work anymore?  Or the problem that this UI branch is not up to date with master?  The reason I'm asking is that the Exoplanets and SkyBrowser should be visible per default according to that file, as well as some other panels

Regarding 3 - let me know if you observe the same behavior (so that it's not just something with my setup that breaks it)

src/panels/Menu/hooks.ts
```diff
@@ -1,52 +1,14 @@
...
-    (id: string, value: boolean) => {
-      dispatch(setMenuItemEnabled({ id: id, enabled: value }));
-      dispatch(setMenuItemVisibleRedux({ id: id, visible: value }));
-    },
-    [dispatch]
-  );
-
-  useEffect(() => {
-    enablePanel('mission', hasMission);
-  }, [hasMission, enablePanel]);
```
WeirdRubberDuck, line 40: Just realised that this is a TODO for the release!  The mission profiles should ce changed so that they load the mission panel per default. 

Let's make an issue?

src/windowmanagement/data/MenuItems.tsx
```diff
@@ -138,7 +142,7 @@ export const menuItemsData: Record<string, MenuItem> = {
     content: <ExoplanetsPanel />,
     renderIcon: (size) => <ExoplanetIcon size={size} />,
     preferredPosition: 'right',
-    defaultVisible: true
+    defaultVisible: false
```
WeirdRubberDuck, line 145: I'm confused. It seems like this is still the setting we listen to, not the setting in the `default_ui_panels.json` file?

WeirdRubberDuck: Conclusion form discussion - change this back to true
```suggestion
    defaultVisible: true
```

src/windowmanagement/data/MenuItems.tsx
```diff
@@ -162,7 +166,7 @@ export const menuItemsData: Record<string, MenuItem> = {
     content: <SkyBrowserPanel />,
     renderIcon: (size) => <TelescopeIcon size={size} />,
     preferredPosition: 'right',
-    defaultVisible: true
+    defaultVisible: false
```
WeirdRubberDuck, line 169:
```suggestion
    defaultVisible: true
```

WeirdRubberDuck (approved):

PR #127, opened 2025-04-25
Feature/notifications system

engbergandreas:
Adds the notification system, I've updated all places where we previously did a `console.log` to use the notification system where it makes sense. 

to be tested and used with: https://github.com/OpenSpace/OpenSpace/pull/3639

also:
Updates Mantine to 7.17.4


src/panels/Menu/hooks.ts
```diff
@@ -60,12 +62,16 @@ export function useStoredLayout() {
...
-      console.error('File is empty');
+      showNotification('Error loading JSON file', 'File is empty', LogLevel.Error);
       return;
     }
     const newLayout = Object.values(content) as TaskbarItemConfig[];
     if (newLayout.length !== menuItems.length) {
-      console.error('Invalid layout file. Length does not match');
+      showNotification(
+        'Error loadding JSON file',
+        'Invali layoutfile, length does not match',
```
ylvaselling, line 72:
```suggestion
        'Invalid layoutfile, length does not match',
```

src/panels/Menu/hooks.ts
```diff
@@ -74,7 +80,11 @@ export function useStoredLayout() {
       menuItems.find((existingItem) => existingItem.id === newItem.id)
     );
     if (!isValid) {
-      console.error("Invalid layout file. All id's must match");
+      showNotification(
+        'Error loading JSON file',
+        "Invalid layout file, all id's must match",
```
ylvaselling, line 85:
```suggestion
        "Invalid layout file, all IDs must match panel IDs",
```

src/redux/local/localSlice.ts
```diff
@@ -65,10 +67,10 @@ export const localSlice = createSlice({
       if (item) {
         item.visible = action.payload.visible;
       } else {
-        // @TODO (ylvse 2025-03-31): handle this error with the notification system?
-        console.error(
-          'Tried to set visibility of non-existent menu item',
-          action.payload.id
+        showNotification(
+          'Error missing menu item',
+          `Tried to set visibility of non-existent enu item: '${action.payload.id}'`,
```
ylvaselling, line 72:
```suggestion
          `Tried to set visibility of non-existent menu item: '${action.payload.id}'`,
```

src/util/logging.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function showNotification(
+  title: string,
+  message: React.ReactNode,
+  level: LogLevel
+) {
+  const color = {
+    [LogLevel.Info]: 'white',
+    [LogLevel.Warning]: 'yellow',
+    [LogLevel.Error]: 'red'
```
ylvaselling, line 13: Place the colors in enums?

Personally i would prefer to defined the object first and then accessing with the key. For some reason it took me a while to realize that was what was happening :) Same below for log

src/util/logging.ts
```diff
@@ -0,0 +1,32 @@
...
+    [LogLevel.Error]: console.error
+  }[level];
+
+  notifications.show({
+    title: title,
+    message: message,
+    color: color
+  });
+
+  log(message);
```
ylvaselling, line 31:
```suggestion
  const colors = {
    [LogLevel.Info]: 'white',
    [LogLevel.Warning]: 'yellow',
    [LogLevel.Error]: 'red'
  };

  const log = {
    // eslint-disable-next-line no-console
    [LogLevel.Info]: console.log,
    // eslint-disable-next-line no-console
    [LogLevel.Warning]: console.warn,
    // eslint-disable-next-line no-console
    [LogLevel.Error]: console.error
  };

  notifications.show({
    title: title,
    message: message,
    color: color[level]
  });

  log[level](message);
```

ylvaselling (approved):
looks good! I think a nice addition would be to make this a function:

```
    if (isReactRenderable(error)) {
      showNotification('Error fetching data', error, LogLevel.Error);
    } else {
      console.error('Error fetching data:', error);
    }
```

Maybe called `handleError` or something? I think it would be nice to make it a function because then we can deliberately hide the console logs from the linter and we won't have any printouts when running the linter :)

src/util/logging.ts
engbergandreas: Yeah agree with the accessing, fixing that. Not sure if it's necessary to put the colors in enums, though?

ylvaselling: Up to you 🤷

src/api/LuaApiProvider.tsx
```diff
@@ -30,7 +32,11 @@ export function LuaApiProvider({ children }: PropsWithChildren) {
         const res = await api.singleReturnLibrary();
         setLuaApi(res);
       } catch (e) {
-        console.error(e);
+        if (isReactRenderable(e)) {
```
WeirdRubberDuck, line 35: How about renaming the function
```suggestion
        if (isReactNode(e)) {
```
?  "Renderable" immediately made me think of OpenSpace, and I was a bit confused. Also, seems like we are actually testing if it is a react node, and not a "renderable" (whatever that would mean in react context)

src/api/LuaApiProvider.tsx
```diff
@@ -30,7 +32,11 @@ export function LuaApiProvider({ children }: PropsWithChildren) {
...
+          showNotification('Error fetching OpenSpace API', e, LogLevel.Error);
```
WeirdRubberDuck, line 36: Wait, what? is the thing we're checking here actually just if we got an error?  Could we change the check to make that more clear? I.e. maybe flip the if statement somehow

src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/EarthPanel.tsx
```diff
@@ -79,7 +82,11 @@ export function EarthPanel({ currentAnchor }: Props) {
 
       setPlaces(uniquePlaces);
     } catch (error) {
-      console.error('Error fetching data:', error);
+      if (isReactRenderable(error)) {
+        showNotification('Error fetching data', error, LogLevel.Error);
```
WeirdRubberDuck, line 86: Same comment here as above.  This check should read "if we got an error" rather than "the error is a react node", and the code should be updated accordingly

Do you see the difference?

src/redux/logging/loggingMiddleware.ts
```diff
@@ -0,0 +1,109 @@
...
+import { showNotification } from '@/util/logging';
+
+import { onCloseConnection, onOpenConnection } from '../connection/connectionSlice';
+import { AppStartListening } from '../listenerMiddleware';
+
+let topic: Topic | null = null;
+
+/** This `LogLevel` must match OpenSpace/Ghoul LogLevel in
+ * https://github.com/OpenSpace/Ghoul/blob/f02810ad2f77166711f4503060e20745f0d808c6/include/ghoul/logging/loglevel.h#L41
+ */
```
WeirdRubberDuck, line 15: Maybe nitpicky, but comment style is important
```suggestion
/** 
 * This `LogLevel` must match OpenSpace/Ghoul LogLevel in
 * https://github.com/OpenSpace/Ghoul/blob/f02810ad2f77166711f4503060e20745f0d808c6/include/ghoul/logging/loglevel.h#L41
 */
```

src/redux/logging/loggingMiddleware.ts
```diff
@@ -0,0 +1,109 @@
...
+      settings: {
+        logLevel: 'Warning'
+      }
+    });
+
+    (async () => {
+      for await (const message of topic.iterator() as AsyncIterable<LogMessage>) {
+        logNotificationMessage(message);
+      }
+    })();
```
WeirdRubberDuck, line 49: Would there be a meaning of making this a named function to call, instead of doing it like this? Even for just readability?

Also, do we need the async function here? Considering that it's already part of an async function? I.e. could this be just: 

```suggestion
    for await (const message of topic.iterator() as AsyncIterable<LogMessage>) {
      logNotificationMessage(message);
    }
```

src/redux/logging/loggingMiddleware.ts
```diff
@@ -0,0 +1,109 @@
...
+      case OpenSpaceLogLevel.Warning:
+        return LogLevel.Warning;
+
+      case OpenSpaceLogLevel.Error:
+      case OpenSpaceLogLevel.Fatal:
+        return LogLevel.Error;
+
+      default:
+        return LogLevel.Info;
+    }
```
WeirdRubberDuck, line 83: maybe this is nitpicky, but i found myself really idliking the extra spaces here
```suggestion
    switch (level) {
      case OpenSpaceLogLevel.Warning:
        return LogLevel.Warning;
      case OpenSpaceLogLevel.Error:
      case OpenSpaceLogLevel.Fatal:
        return LogLevel.Error;
      default:
        return LogLevel.Info;
    }
```

src/util/fileIOhooks.ts
```diff
@@ -18,7 +23,11 @@ function useLoadJsonFile(handlePickedFile: (content: JSON) => void): () => void
       handlePickedFile(json);
     } catch (e) {
       // TODO: do we want to throw here?
-      console.error('Error parsing file', e);
+      if (isReactRenderable(e)) {
+        showNotification('Error parsing file', e, LogLevel.Error);
```
WeirdRubberDuck, line 27: Same comment as above

src/util/fileIOhooks.ts
```diff
@@ -60,8 +69,11 @@ async function openSaveFileDialog(contents: JSON) {
       // Close the file and write the contents to disk
       await writable.close();
     } catch (e) {
-      // @TODO ylvse 2025-03-31: handle this error with the notification system?
-      console.error(e);
+      if (isReactRenderable(e)) {
+        showNotification('Error parsing file', e, LogLevel.Error);
```
WeirdRubberDuck, line 73: See above

src/util/logging.ts
WeirdRubberDuck: Agreeing with the comment on accessing here. This syntax is quite confusing

src/util/reactHelpers.ts
```diff
@@ -0,0 +1,16 @@
+import { isValidElement, ReactNode } from 'react';
+
+/**
+ * This utility function checks whether an unkown value is of type ReactNode, useful when
+ * sending notifications from a try/catch
+ */
+export function isReactRenderable(value: unknown): value is ReactNode {
```
WeirdRubberDuck, line 7: See above
```suggestion
export function isReactNode(value: unknown): value is ReactNode {
```

Also, what's going on with the return value for the function here? Why not `boolean`?

Seems like `is` should be used for type testing? I.e. maybe in the function?

WeirdRubberDuck (changes requested):
Getting linking issues when trying to run the OS branch, so haven't been able to try it :(  But here are some code comments for now

src/redux/logging/loggingMiddleware.ts
```diff
@@ -0,0 +1,109 @@
...
+  dateStamp: string;
+};
+
+export const setupSubscription = createAsyncThunk(
+  'logging/setupSubscription',
+  async () => {
+    topic = api.startTopic('errorLog', {
+      event: 'start_subscription',
+      settings: {
+        logLevel: 'Warning'
```
WeirdRubberDuck, line 41: Add a comment as to what this means. I.e. that we subscribe to all log messages that are "Warning" or worse. 

And maybe a comment as to why we are not including Info level log messages as of now? These are otherwise the type of messages that are intended for the user, but with the new notification system they would be too intrusive

WeirdRubberDuck (commented):
Tested it now!  It's beautiful! :) 

I guess it's intentional that we currently do not show info level messages as notifications? That would be a lot more intrusive than we probably want it to be.

src/api/LuaApiProvider.tsx
engbergandreas: The check is if we can show the error as a notification, since `e`is unkown from the catch we must verify its type as a React.Node before using `showNotification` otherwise we will just print it to the console

ylvaselling:
Tried it again, and I got lots of notifications so I was thinking maybe we need a way to clear them or to have an option to not show them? Could be useful when making a recording for example

Also I noticed that the logmanager also logs how many logs there were:
image

Maybe that's a bit meta? But also not a huge deal

ylvaselling (approved):

src/util/reactHelpers.ts
engbergandreas: This is to tell TypeScript that the value we passed in is of the type ReactNode
Otherwise we would get an error here: 
```javascript
        if (isReactRenderable(e)) {
          showNotification('Error fetching OpenSpace API', e, LogLevel.Error);  //<--- argument of type 'unknown' is not assignable to parameter of type 'ReactNode'.ts(2345)
        }
```

alexanderbock: Dumb Typescript question and not a suggestion. How is this different from:
`export function isReactNode(value: ReactNode)`
?

engbergandreas: You wouldn't be able to call it when the value is unknown which is the case in try/catch blocks. So the following wouldn't work: 

```javascript
try {
}
catch(e) {
 isReactNode(e) // <-- Argument of type 'unknown' is not assignable to parameter of type 'ReactNode'.ts(2345)
} 
```

engbergandreas: To clarify my previous comment, when using the `value is ReactNode`, TypeScript knows that in the scope of the function where it's called, after that function has finished and returned `true`, the value can be treated as a `ReactNode`

src/redux/profile/profileMiddleware.ts
```diff
@@ -30,7 +31,20 @@ export const addProfileListener = (startListening: AppStartListening) => {
...
+          listenerApi.dispatch(
+            handleNotificationLogging(
+              'Error missing menu item',
+              `Tried to set visibility of non-existent menu item: '${key}'`,
+              LogLevel.Error
+            )
+          );
+        } else {
+          listenerApi.dispatch(setMenuItemVisible({ id: key, visible: value }));
+        }
```
ylvaselling, line 47: flip this if statement? so the "good state" is in the if

ylvaselling (approved):
image
Works well. The menu is starting to look a bit busy so maybe some gap above headings could be good

WeirdRubberDuck:
> image Works well. The menu is starting to look a bit busy so maybe some gap above headings could be good

Update to mantine 8 and use the submenus feature for notifications?  I think yet another submenu would be nice here

WeirdRubberDuck:
Awaiting any final changes before reviewing this one. Seemed like there were some more changes that were intended and don't wan tot have to review twice if not necessary

Let me know about your plans @engbergandreas

WeirdRubberDuck:
Also, the GUI setting menu got too busy, so now the notification settings are in their own menu: 
image

PR #128, opened 2025-04-25
Custom GUI scale slider in View Menu 

WeirdRubberDuck:
1. Add a custom slider for the GUI scale in the view menu, that sets the value onChangeEnd
   * Note that we decided to not use the min/max/step of the property in this case, and instead set min and max differently.  
2. Move visibility level setting into the new "GUI Settings" menu created for the scale setting
3. Re-fix the view menu not showing at startup. It was fixed in #121, but the change was seemingly lost as part of a merge conflict resolve

### DIFF
--- a/src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
+++ b/src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
@@ -1,4 +1,12 @@
-import { Button, CheckboxIndicator, Group, Menu, Radio, Stack } from '@mantine/core';
+import {
+  Button,
+  CheckboxIndicator,
+  Group,
+  Menu,
+  Radio,
+  Slider,
+  Stack
+} from '@mantine/core';
 
 import { DragReorderList } from '@/components/DragReorderList/DragReorderList';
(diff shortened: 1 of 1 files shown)


src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
```diff
@@ -128,28 +142,58 @@ export function ViewMenu() {
...
+                ))
+              ) : (
+                <LoadingBlocks />
+              )}
+            </Stack>
+          </Radio.Group>
+        </Container>
+
+        <Menu.Label>
+          <Group justify={'space-between'}>
```
ylvaselling, line 171: For consistency with other info boxes, maybe we should place this info box next to the title?

ylvaselling (approved):

src/panels/Menu/TopMenuBar/Menus/ViewMenu.tsx
WeirdRubberDuck: I agree, and think we should do this everywhere TBH.  

Here I just copied the code for the already existing visibility label, but I'll remove the space between for that one as well

engbergandreas (approved):

PR #130, opened 2025-04-28
Fix/broken properties

WeirdRubberDuck:
Brings back the label for `BoolProperty` and fixes `TriggerProperty`, which led to an OpenSpace error on trigger due to the wrong type being used in the setValue function. 

Would have merged this to `master` directly, but would like a thumbs up from @ylvaselling just to see that the changes I made in `property.ts` makes sense.  

If I understand the problem correctly, using the `& AdditionalData<T> &
    ViewOptions<T>` for MetaData made the meta data object to somehow get the properties from the `EmptyObject` type, and somehow this made it so that we wouldn't get a type error when we wrote `meta.name` instead of `meta.guiName`. Not quite sure why the issue occurred, but this rewrite fixes it and still infers the correct type for the additional data and viewoptions

### DIFF
--- a/src/types/Property/property.ts
+++ b/src/types/Property/property.ts
@@ -15,26 +15,26 @@ type EmptyObject = Record<string, never>;
  * Extracts the `additionalData` property from a specific property type in `PropertyTypes`.
  *
  * This utility type checks if the specified property type in `PropertyTypes` contains
- * an `additionalData` field. If it does, it creates a new type with the `additionalData`
- * field. Otherwise, it defaults to an empty object type.
+ * an `additionalData` field. If it does, it uses the inferred type for the additional
+ * data. Otherwise, it defaults to an empty object type.
  */
 type AdditionalData<T extends keyof PropertyTypes> = PropertyTypes[T] extends {
   additionalData: infer A;
 }
-  ? { additionalData: A }
+  ? A
   : EmptyObject;
 
 /**
  * Extracts the `viewOptions` property from a specific property type in `PropertyTypes`.
  *
  * This utility type checks if the specified property type in `PropertyTypes` contains
- * an `viewOptions` field. If it does, it creates a new type with the `viewOptions`
- * field. Otherwise, it defaults to an empty object type.
+ * an `viewOptions` field. If it does, it uses the inferred type for the view options.
+ * Otherwise, it defaults to an empty object type.
  */
 type ViewOptions<T extends keyof PropertyTypes> = PropertyTypes[T] extends {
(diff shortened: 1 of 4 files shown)


src/types/Property/property.ts
```diff
@@ -15,26 +15,26 @@ type EmptyObject = Record<string, never>;
  * Extracts the `additionalData` property from a specific property type in `PropertyTypes`.
  *
  * This utility type checks if the specified property type in `PropertyTypes` contains
- * an `additionalData` field. If it does, it creates a new type with the `additionalData`
```
ylvaselling, line 18: This creates the viewOptions and additionalData keys for types that shouldn't have them, for example BoolProperty (although the objects will be empty). Don't really see what this is intended to fix?

src/types/Property/property.ts
```diff
@@ -15,26 +15,26 @@ type EmptyObject = Record<string, never>;
...
- * field. Otherwise, it defaults to an empty object type.
+ * an `additionalData` field. If it does, it uses the inferred type for the additional
+ * data. Otherwise, it defaults to an empty object type.
  */
 type AdditionalData<T extends keyof PropertyTypes> = PropertyTypes[T] extends {
   additionalData: infer A;
 }
-  ? { additionalData: A }
+  ? A
   : EmptyObject;
```
ylvaselling, line 25: I was testing some stuff and a way to make typescript give an error when using a key that doesn't exist is to use {} instead of Record<string, never> for the EmptyObject

ylvaselling (changes requested):
Good catch!

src/types/Property/property.ts
WeirdRubberDuck: I see!  Before I realised what you meant by your comment, I pushed an alternative solution with `undefined` and optional keys instead. 

First let me know what you think about that one. The type for the `BoolProperty` will e.g. look like this then: 
image. 


I could also change it to your proposed solution of using `{}`. However, this would mean that we would have to disable an eslint rule, like this: 
```
type ViewOptions<T extends keyof PropertyTypes> = PropertyTypes[T] extends {
  viewOptions: infer A;
}
  ? { viewOptions: A }
  : // eslint-disable-next-line @typescript-eslint/no-empty-object-type
    {};
```
since we otherwise get this error as `{}` apparently allows more things than just empty object : 
image


Would be good to discuss what solution we prefer.

WeirdRubberDuck: I agree. And that solution seems to work well :) 

Awaiting approve and then I'll merge this PR. Thanks for the input

PR #131, opened 2025-04-28
Fix missing copy URI in property labels and split label component up into one specific for properties

WeirdRubberDuck:
As part of a recent PR, the code for property labels had been rewritten to use a common Label component that can be used for inputs that are not properties. 

This PR fixes some issues with that rewrite: 

1. The copy URI button had disappeared for all property types, and it would not have been possible to add it to a `BoolProperty`. Now, the `BoolInput` component can take a custom label component, so that we can use the `PropertyLabel` for it for bool properties. 
2.  The generic `Label` component now knows nothing about properties. It was confusing to me that it had a bunch of property-specific text when it was seemingly intended to be a generic label that can be used for anything for which we want an infobox description. Instead, there is a specific `PropertyLabel` component, that uses the `Label` component so that the visual style is the same

Some questions/TODOs:
- We should use the `BoolInput` component everywhere we are currently using a `Checkbox` component. Do you think I should do this as part of this PR @ylvaselling @engbergandreas ?
- What's the intended use of the `Label` component @ylvaselling?  Is it only for inputs, or for anything?  If it's only for inputs, maybe we should rename it to `InputLabel`

### DIFF
--- a/src/components/Label/Label.tsx
+++ b/src/components/Label/Label.tsx
@@ -1,42 +1,22 @@
 import { JSX } from 'react';
-import { Group, InputLabel, Text, Tooltip } from '@mantine/core';
+import { Group, InputLabel, Text } from '@mantine/core';
 
-import CopyUriButton from '@/components/CopyUriButton/CopyUriButton';
 import { InfoBox } from '@/components/InfoBox/InfoBox';
-import { Uri } from '@/types/types';
 
 interface Props {
-  name: string;
+  name: string | JSX.Element;
   description: string | JSX.Element;
-  uri?: Uri;
-  readOnly?: boolean;
 }
 
-export function Label({ name, description, uri, readOnly = false }: Props) {
+export function Label({ name, description }: Props) {
   return (
     <Group wrap={'nowrap'}>
       <InputLabel fw={'normal'}>
         <Text span size={'sm'}>
           {name}
         </Text>
-        {readOnly && (
-          <Tooltip
-            maw={200}
-            multiline
-            label={`This property is read-only, meaning that it's not intended to be changed.`}
-          >
-            <Text span ml={'xs'} size={'xs'} c={'dimmed'}>
-              (Read-only)
-            </Text>
-          </Tooltip>
-        )}
       </InputLabel>
-      {description && (
-        <InfoBox>
-          {description}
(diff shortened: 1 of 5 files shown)


ylvaselling (approved):

src/components/Label/Label.tsx
```diff
@@ -1,42 +1,22 @@
...
-import { Group, InputLabel, Text, Tooltip } from '@mantine/core';
+import { Group, InputLabel, Text } from '@mantine/core';
 
-import CopyUriButton from '@/components/CopyUriButton/CopyUriButton';
 import { InfoBox } from '@/components/InfoBox/InfoBox';
-import { Uri } from '@/types/types';
 
 interface Props {
-  name: string;
+  name: string | JSX.Element;
```
engbergandreas, line 7:
```suggestion
  name: JSX.Element;
```
String is part of JSX.Element

src/components/Property/PropertyLabel.tsx
```diff
@@ -0,0 +1,43 @@
+import { JSX } from 'react';
+import { Text, Tooltip } from '@mantine/core';
+
+import CopyUriButton from '@/components/CopyUriButton/CopyUriButton';
+import { Uri } from '@/types/types';
+
+import { Label } from '../Label/Label';
```
engbergandreas, line 7: import using @

src/components/Property/PropertyLabel.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+interface Props {
+  name: string;
+  description: string | JSX.Element;
```
engbergandreas, line 11:
```suggestion
  description: JSX.Element;
```

engbergandreas (approved):

engbergandreas:
Like Ylva said, if you want to make the change with checkboxes, go ahead!

src/components/Property/PropertyLabel.tsx
WeirdRubberDuck: Nope: 
image

But this should be `string | React.JSX.Element;`

WeirdRubberDuck:
I'll leave the checkboxes to another Pr, and left the name of the Label component as is

PR #134, opened 2025-05-02
Language/components

engbergandreas:
Added localization for all components that had some form of text. However, I did not do it for thrown errors, which should only occur during development, e.g., `throw Error('useSearchResultProvider must be used within a SearchResultProvider');`
But then I ended up not doing it for any thrown errors, do we want it in those places anyway?

Do check the JSON file structure and see what you think of the nesting. In some places, e.g., properties are grouped under `property`, while others, for example, `color-edit` and `color-picker` are two separate entries. We should discuss when and if we should group them. 

The order of the entries is alphabetical at the top-level, i.e., `about`, `collapsable-header` etc. The internal structure is either top to bottom or grouped by some internal hierarchy for that specific component. 

Did not do anything for `Pills.tsx` since that one has an aria-label using `name` that is deprecated? Something we want to look at. 

I also noticed that `FilterlistGrid.tsx` is an empty file, and should be deleted?

### DIFF
--- a/public/locales/en/components.json
+++ b/public/locales/en/components.json
@@ -2,11 +2,134 @@
   "about": {
     "modal-title": "About OpenSpace",
     "img-alt-text": "OpenSpace logo",
-    "about-openspace": "OpenSpace is open-source interactive data visualization software designed to visualize the entire known universe and portray our ongoing efforts to investigate the cosmos.",
+    "about-openspace-description": "OpenSpace is open-source interactive data visualization software designed to visualize the entire known universe and portray our ongoing efforts to investigate the cosmos.",
     "fetching-version-number": "Fetching OpenSpace version...",
     "custom-version-number": "Custom"
   },
   "collapsable-header": {
-    "chevron-aria-label": "{{variable}} ? open : close "
+    "chevron-aria-label": {
+      "open": "Open",
+      "close": "Close"
+    }
+  },
+  "color-edit": {
+    "error-invalid-color": "Invalid color"
+  },
+  "color-picker": {
+    "color-swatch-aria-label": "Open color edit"
+  },
+  "copy-to-clipboard-button": {
+    "tooltip": {
+      "copy": "Copy",
+      "copied": "Copied"
+    },
+    "label": "Copy"
+  },
+  "copy-uri-button": {
+    "label": "Copy URI:"
+  },
+  "drag-reorder-list": {
+    "drag-handle-aria-label": "Move item up or down (keyboard control not supported)"
+  },
+  "error-fallback": {
+    "alert-title": "🚨 Houston, we have a problem...🚨",
+    "alert-text": "An error with the following message was thrown:",
+    "reload-button-label": "Reload page"
+  },
+  "filter-list": {
+    "search-results": {
+      "empty-results": "No results found. Try another search!"
+    },
+    "search-settings-menu": {
+      "no-selection-tooltip": "Nothing is selected. Search will be empty.",
+      "dropdown-menu-label": "Search in"
+    },
+    "filter-list-input-field": {
+      "cancel-button-aria-label": "Clear search input",
+      "show-more-button": {
+        "less": "Less",
+        "more": "More"
+      }
+    }
+  },
+  "friction-controls": {
+    "rotation-label": "Rotation",
+    "zoom-label": "Zoom",
+    "roll-label": "Roll"
+  },
+  "friction-controls-info": {
+    "heading": "Friction controls",
+    "description": "Enable or disable rotation, zoom, and roll friction. If checked, applies friction to camera movement."
+  },
+  "info-box": {
+    "aria-label": "More information"
+  },
+  "input": {
+    "numeric-input": {
+      "error-label": "Value outside range [{{min}}, {{max}}]"
+    },
+    "step-control-button": {
+      "aria-label": {
+        "increment": "Increment value",
+        "decrement": "Decrement value"
+      }
+    }
+  },
+  "node-navigation-button": {
+    "jump": {
+      "title": "Jump to",
+      "tooltip": "Teleport to object using a fade transition"
+    },
+    "jump-geo": {
+      "title": "Jump to Geo",
+      "tooltip": "Teleport to position using a fade transition"
+    },
+    "focus": {
+      "title": "Focus",
+      "tooltip": "Focus the object (set as anchor). Hold <0>Shift</0> when clicking to focus without retargeting"
+    },
+    "fly": {
+      "title": "Fly to",
+      "tooltip": "Trigger a flight to the object. Hold <0>Shift</0> when clicking to instantly teleport/jump"
+    },
+    "fly-geo": {
+      "title": "Fly to Geo",
+      "tooltip": "Trigger a flight to the position. Hold <0>Shift</0> when clicking to instantly teleport/jump"
+    },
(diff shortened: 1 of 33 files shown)


public/locales/en/components.json
```diff
@@ -0,0 +1,135 @@
...
+  },
+  "friction-controls": {
+    "rotation-label": "Rotation",
+    "zoom-label": "Zoom",
+    "roll-label": "Roll"
+  },
+  "friction-controls-info": {
+    "heading": "Friction controls",
+    "description": "Enable or disable rotation, zoom, and roll friction. If checked, applies friction to camera movement."
+  },
```
ylvaselling, line 63: nest the info in `friction-controls` ? So that its 
```
"friction-controls: {
 ...
 "info": {
....
}
}
"
```

public/locales/en/components.json
```diff
@@ -0,0 +1,135 @@
...
+  "color-picker": {
+    "color-swatch-aria-label": "Open color edit"
+  },
+  "copy-to-clipboard-button": {
+    "tooltip": {
+      "copy": "Copy",
+      "copied": "Copied"
+    },
+    "label": "Copy"
```
WeirdRubberDuck, line 26: This could potentially just be 
```suggestion
    "copy": "Copy",
    "copied": "Copied"
```
and use `"copy"` for the label as well

WeirdRubberDuck: Alternatively, I think you should be able to reuse values from other keys, to avoid duplication

```suggestion
    "copy": "Copy",
    "copied": "Copied",
    "label": "$t(components.copy-to-clipboard-button.copy)"
```

Not sure what the preferred way to go is, though. Maybe it's easier to have it split up as you do now, even if it requires translating the same word multiple times

public/locales/en/components.json
```diff
@@ -0,0 +1,135 @@
...
+      "placeholder-empty-field": "No selection"
+    },
+    "string-property": {
+      "aria-label": "{{guiName}} input"
+    },
+    "property-label": {
+      "tooltip-label": "This property is read-only, meaning that it's not intended to be changed.",
+      "label": "(Read-only)"
+    }
+  },
```
WeirdRubberDuck, line 128: How about removing "property" from the subkeys here?
```suggestion
  "property": {
    "list": {
      "double-list-placeholder-text": "number1, number2, ...",
      "int-list-placeholder-text": "integer1, integer2, ...",
      "string-list-placeholder-text": "item1, item2, ..."
    },
    "option": {
      "aria-label": "{{guiName}} option input",
      "placeholder": {
        "options": "Choose an option",
        "no-options": "No options were loaded"
      }
    },
    "selection": {
      "aria-label": "{{guiName}} multi-select",
      "placeholder-empty-field": "No selection"
    },
    "string": {
      "aria-label": "{{guiName}} input"
    },
    "label": {
      "tooltip-label": "This property is read-only, meaning that it's not intended to be changed.",
      "label": "(Read-only)"
    }
  },
```

public/locales/en/components.json
```diff
@@ -0,0 +1,135 @@
...
+    "selection-property": {
+      "aria-label": "{{guiName}} multi-select",
+      "placeholder-empty-field": "No selection"
+    },
+    "string-property": {
+      "aria-label": "{{guiName}} input"
+    },
+    "property-label": {
+      "tooltip-label": "This property is read-only, meaning that it's not intended to be changed.",
+      "label": "(Read-only)"
```
WeirdRubberDuck, line 126:
```suggestion
      "read-only-tooltip": "This property is read-only, meaning that it's not intended to be changed.",
      "read-only": "Read-only"
```
? Not clear that this key only applies for the read-only parts until you start reading the text

Also note that my suggestion includes moving the parentheses out of the "(Read-only)" text. Please apply these to the component instead :) 

These should be the same for all languages and are not really part of the text, but the formatting, and we don't want translators to accidentally remove them

public/locales/en/components.json
```diff
@@ -0,0 +1,135 @@
+{
+  "about": {
+    "modal-title": "About OpenSpace",
+    "img-alt-text": "OpenSpace logo",
```
WeirdRubberDuck, line 4:
```suggestion
    "logo-alt-text": "OpenSpace logo",
```
`img` is too vague here, and we might add another image in the future

src/components/NodeNavigationButton/NodeNavigationButton.tsx
```diff
@@ -150,59 +152,41 @@ export function NodeNavigationButton({
...
-      content.title = 'Fly to';
+      content.title = t('fly.title');
       content.icon = <AirplaneIcon />;
-      content.tooltip = (
-        <span>
-          Trigger a flight to the object. Hold <Kbd>Shift</Kbd> when clicking to instantly
-          teleport/jump
-        </span>
-      );
+      content.tooltip = <Trans t={t} i18nKey={'fly.tooltip'} components={[<Kbd />]} />;
```
WeirdRubberDuck, line 175: Interesting

src/components/Property/PropertyLabel.tsx
```diff
@@ -22,10 +25,10 @@ export function PropertyLabel({ name, description, uri, readOnly = false }: Prop
             <Tooltip
               maw={200}
               multiline
-              label={`This property is read-only, meaning that it's not intended to be changed.`}
+              label={t('property.property-label.tooltip-label')}
             >
               <Text span ml={'xs'} size={'xs'} c={'dimmed'}>
-                (Read-only)
+                {t('property.property-label.label')}
```
WeirdRubberDuck, line 31: See comment in json file :) The keys here need to be revisited

src/localization/resources.ts
```diff
@@ -0,0 +1,13 @@
+import components from '@/public/locales/en/components.json';
+
+/**
+ * This object in combination with `i18next.d.ts` enables intellisense and syntax
+ * highlighting when accessing the translation units, e.g., useTranslation('components');
+ */
+export const resources = {
+  en: {
+    components: components
+  }
```
WeirdRubberDuck, line 10: should this be `en-US` ? remembering Alex's comment

src/types/i18next.d.ts
```diff
@@ -0,0 +1,9 @@
+import 'i18next';
+
+import type { Resources } from '@/localization/resources';
+
+declare module 'i18next' {
+  interface CustomTypeOptions {
+    resources: Resources['en'];
```
WeirdRubberDuck, line 7: same here. `en-US`?

WeirdRubberDuck (commented):
Nice! 

Agreeing with Ylvas comments. But in general, the structure makes sense :)

src/localization/resources.ts
alexanderbock: Yeees.. `en-US` please 🙏      ISO 639-1 + 3166-1 ftw

public/locales/en/components.json
engbergandreas: Yeah, I think it would be nice to have some kind of common keys if they are recurring. However, having this: `"label": "$t(components.copy-to-clipboard-button.copy)"` is going to be a lot of mental overhead to put on the users/community to understand. So then having to translate the same word twice is faster and easier

WeirdRubberDuck: Definitely, especially since it's in the same place here

Something to think about in future places for common words though

PR #135, opened 2025-05-07
Use common `BoolInput` component for all checkboxes

WeirdRubberDuck:
Replaces all checkboxes with the new `BoolInput` component, wherever possible

Fixes so all checkboxes: 
* Have aria-labels
* Are settable using 'Enter'
* Can be set by clicking the title
* Have consistent formatting, including infobox


src/components/FilterList/SearchSettingsMenu/FilterListSearchSettingsMenu.tsx
```diff
@@ -49,10 +42,10 @@ export function FilterListSearchSettingsMenu<T extends object>({
           {/* When using Object.entries a new object is created, and we cant infer the type from that */}
           {Object.entries(keys).map(([key, enabled]) => (
             <Group key={key}>
-              <Checkbox
-                label={labels ? labels[key as keyof T] : camelCaseToRegularText(key)}
-                checked={enabled as boolean}
-                onChange={(event) => setKey(key as keyof T, event.currentTarget.checked)}
+              <BoolInput
+                name={labels ? labels[key as keyof T] || '' : camelCaseToRegularText(key)}
```
ylvaselling, line 46: I think in the case that we have a labels object but that we don't have that particular key in the object, we want to fall back on the camelCaseToRegularText, not an empty string.

Example: 
```
keys = {
action: true,
guiPath: false
}

labels = {
guiPath: "User Interface Path"
}
```
In this case, action will become an empty string in this line.
Alternatively, we can enforce in the types that all the keys in `keys` must also be defined in the `labels` prop, if `labels`is passed

src/components/Input/BoolInput.tsx
```diff
@@ -1,35 +1,45 @@
 import { JSX } from 'react';
-import { Checkbox, Group } from '@mantine/core';
+import { Checkbox, Group, MantineStyleProps } from '@mantine/core';
 
-import { Label } from '../Label/Label';
+import { InfoBox } from '../InfoBox/InfoBox';
```
ylvaselling, line 4: @ notation

src/components/Input/BoolInput.tsx
```diff
@@ -1,35 +1,45 @@
...
-    <Group gap={'xs'} wrap={'nowrap'}>
+    <Group gap={'xs'} wrap={'nowrap'} {...styleProps}>
       <Checkbox
         checked={value}
         onChange={(event) => setValue(event.currentTarget.checked)}
         onKeyDown={(event) => event.key === 'Enter' && setValue(!value)}
         disabled={disabled}
+        defaultChecked={defaultChecked}
+        label={showLabel ? name : undefined}
         aria-label={name}
```
ylvaselling, line 40: since we use name for aria label, should throw an error if name is set to an empty string? It seems to me that we always expect here that name is set to something

src/components/Property/PropertyLabel.tsx
```diff
@@ -31,11 +31,10 @@ export function PropertyLabel({ name, description, uri, readOnly = false }: Prop
...
         </>
       }
-      description={
-        <>
-          {description}
-          {uri && <CopyUriButton uri={uri} />}
-        </>
+      info={
+        description ? (
+          <PropertyDescription uri={uri} description={description} />
```
ylvaselling, line 36: Maybe here we want to display the info box just to show the uri even if there is no description?

src/panels/ExoplanetsPanel/ExoplanetsSettings.tsx
```diff
@@ -64,47 +64,29 @@ export function ExoplanetsSettings({ hasAddedExoplanets }: Props) {
...
+          name={'1 AU Size Ring'}
+          value={show1AuRing}
+          setValue={toggleShow1AuRing}
+          info={`If true, show a ring with the radius 1 AU around the host star of each system,
             to use for size comparison. Setting the value automatically updates the
-            visibility for all added exoplanet systems.
-          </InfoBox>
-        </Group>
+            visibility for all added exoplanet systems.`}
+        />
```
ylvaselling, line 89: Since these are properties, how about we update the description in the engine and just use the `Property` component? I think these properties are only used here anyway, and in the module settings?

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -53,43 +52,27 @@ export function SceneTreeFilters({ setFilter, filter }: Props) {
...
+              setValue={(value) => setFilter({ onlyFocusable: value })}
+              info={`Hide scene graph nodes that are not markes as focusable, meaning that they
+                cannot be directly set as the focus node in the scene.`}
+            />
+            <BoolInput
+              name={'Show objects with GUI hidden flag'}
+              value={filter.includeGuiHiddenNodes}
+              setValue={(value) => setFilter({ includeGuiHiddenNodes: value })}
+              info={`Show scene graph nodes that are marked as hidden in the GUI part of the
+                asset. These are otherwise hidden in the interface..`}
```
ylvaselling, line 74:
```suggestion
                asset. These are otherwise hidden in the interface.`}
```

src/panels/Scene/SceneTree/SceneTreeFilters.tsx
```diff
@@ -53,43 +52,27 @@ export function SceneTreeFilters({ setFilter, filter }: Props) {
...
-                Show scene graph nodes that are marked as hidden in the GUI part of the
-                asset. These are otherwise hidden in the interface.
-              </InfoBox>
-            </Group>
+          <Stack px={'xs'} gap={'xs'}>
+            <BoolInput
+              name={'Show only visible'}
+              value={filter.showOnlyVisible}
+              setValue={(value) => setFilter({ showOnlyVisible: value })}
+              info={'Visible = Enabled and not faded out'}
```
ylvaselling, line 60:
```suggestion
              info={'Visible = Enabled and not faded out.'}
```

ylvaselling: Maybe this description should be written in the same style as the other two?

src/components/Input/BoolInput.tsx
```diff
@@ -1,35 +1,45 @@
...
-  description: string | JSX.Element;
-  value: boolean;
+  value?: boolean; // This is optional to allow for defaultChecked to work
   setValue: (value: boolean) => void;
-  label?: JSX.Element;
+  showLabel?: boolean;
+  defaultChecked?: boolean;
+  info?: string | JSX.Element;
   disabled?: boolean;
 }
```
ylvaselling, line 14: Personally i find it a little confusing that `name` is used as the label, which you can set with the `showLabel` prop.

Alternative suggestion: 
* Rename `name` to `label` and make it optional, and remove showLabel.
* Pass in a second optional prop for the `ariaLabel`. 
* Use the label as the aria label in case aria-label doesnt exist. If both don't exist, throw an error. Also throw if they are empty strings.

What do you think?

ylvaselling (changes requested):
Looks great! You're doing god's work

Had some minor comments

src/components/Input/BoolInput.tsx
WeirdRubberDuck: Makes sense!

src/components/Input/BoolInput.tsx
WeirdRubberDuck: I like it!   Was also thinking that `label` would make more sense than `Name`,  and having the option to customize the aria-label would probably be very useful.

src/components/FilterList/SearchSettingsMenu/FilterListSearchSettingsMenu.tsx
WeirdRubberDuck: Let me see if I understand this correctly. This should work and address what you're mentioning here, right?
```
                label={(labels && labels[key as keyof T]) || camelCaseToRegularText(key)}
```

ylvaselling: Correct, you can also write it like this

```
labels?.[key as keyof T] || camelCaseToRegularText(key)
```

ylvaselling (approved):
Some ideas about types for checking the required props, otherwise looks good!

src/components/Input/BoolInput.tsx
WeirdRubberDuck: Good point. I ended up doing it like this, since we should not provide boty a `defaultChecked` and a `value`. Thoughts? 

```
interface BaseProps extends MantineStyleProps {
  setValue: (value: boolean) => void;
  info?: string | JSX.Element;
  disabled?: boolean;
}

// Either the label or the ariaLabel must be provided, or both.
interface Label {
  label?: string;
  ariaLabel?: string;
}
type Labels = RequireAtLeastOne<Label, 'label' | 'ariaLabel'>;

interface PropsWithValue extends BaseProps {
  value: boolean;
  defaultChecked?: never;
}

interface PropsWithDefault extends BaseProps {
  defaultChecked: boolean;
  value?: never;
}

type Props = (PropsWithValue | PropsWithDefault) & Labels;
``` 

The error messages we get from both of these checks (`RequireAtLeastOne` and ` (PropsWithValue | PropsWithDefault) `) are quite confusing, but better than no error at all

WeirdRubberDuck: Edit: `defaultChecked` should be optional in `PropsWithDefault`, since that leads to the default value being not checked if nothing is specified. we had a pace where this was the case.

src/components/Input/BoolInput.tsx
```diff
@@ -1,35 +1,76 @@
...
+  onChange?: (value: boolean) => void;
+  info?: string | JSX.Element;
   disabled?: boolean;
 }
 
+// Either the label or the ariaLabel must be provided, or both
+interface Label {
+  label?: string;
+  ariaLabel?: string;
+}
```
engbergandreas, line 17:
```suggestion
}

```

src/components/Input/BoolInput.tsx
```diff
@@ -1,35 +1,76 @@
...
+
+// If defaultChecked is provided, value must not be provided, but it's also possible to
+// not specify either, in which case defaultChecked will be considered false by default
+interface PropsWithDefault extends BaseProps {
+  defaultChecked?: boolean;
+  value?: never;
+}
+
+type Props = (PropsWithValue | PropsWithDefault) & Labels;
```
engbergandreas, line 33: What's the reason for not extending CheckboxProps? I'm not sure I understand the reasoning behind handling PropsWithDefault and PropsWithValue like we do above here..?

engbergandreas: iirc mantine allows both value and defaultchecked?

src/components/Label/Label.tsx
```diff
@@ -5,18 +5,18 @@ import { InfoBox } from '@/components/InfoBox/InfoBox';
 
 interface Props {
   name: string | JSX.Element;
-  description: string | JSX.Element;
+  info?: string | JSX.Element;
```
engbergandreas, line 8: Should this be React.ReactNode instead of just JSX.Element?

engbergandreas (approved):

src/components/Input/BoolInput.tsx
ylvaselling: Nice looks good!

src/components/Input/BoolInput.tsx
WeirdRubberDuck: Hm, I goruped them intentionally to show that they are connected

src/components/Label/Label.tsx
WeirdRubberDuck: Probably. IIRC, we should always use ReactNode rather than JSX.Element, right?  

We have been a bit sloppy in that case, and should fix this throughout the code base. Would prefer to leave this to after this PR is merged

src/components/Input/BoolInput.tsx
engbergandreas: Sure a todo is fine

PR #136, opened 2025-05-08
First attempt in tightening up the mission panel a bit

WeirdRubberDuck:
More work is probably needed, but this is at least a little better than it was before

Before VS after


alexanderbock (commented):
No comment on the code, but 👍 on the looks

ylvaselling (approved):

PR #137, opened 2025-05-08
Refactor keybinds panel

WeirdRubberDuck:
1. Changes title to "Keybinds" instead of "Keybindings Layout", since it now also includes a list of all keybinds
2. Add the identifier of the bound action, and a copy button 
3. Change tab style to be consistent with tabs in other places
4. Increase width to fit entire keyboard
5. General tighten up, and redesign both layouts to work better with the wide format that the panel has, due to the keyboard

After: 

Before:

Ley me know what you think about the redesign, especially the list view. If we don't like the change of moving the selected keybind to the right instead of the bottom, I can change it back to how it was before

### DIFF
--- a/src/panels/KeybindsPanel/KeybindsPanel.tsx
+++ b/src/panels/KeybindsPanel/KeybindsPanel.tsx
@@ -1,13 +1,19 @@
 import { Box, Tabs } from '@mantine/core';
+import { useElementSize } from '@mantine/hooks';
+
+import { useWindowSize } from '@/windowmanagement/Window/hooks';
 
 import { KeyboardLayout } from './KeyboardLayout';
 import { ListLayout } from './ListLayout';
 
 export function KeybindsPanel() {
+  const { height: windowHeight } = useWindowSize();
+  const { ref, height: tabsHeight } = useElementSize();
+
   return (
     <Box>
       <Tabs radius={'md'} defaultValue={'keyboardLayout'}>
-        <Tabs.List>
+        <Tabs.List ref={ref}>
           <Tabs.Tab value={'keyboardLayout'}>Keyboard View</Tabs.Tab>
           <Tabs.Tab value={'listLayout'}>List View</Tabs.Tab>
         </Tabs.List>
@@ -16,7 +22,7 @@ export function KeybindsPanel() {
           <KeyboardLayout />
         </Tabs.Panel>
         <Tabs.Panel value={'listLayout'}>
-          <ListLayout />
+          <ListLayout height={windowHeight - tabsHeight} />
         </Tabs.Panel>
       </Tabs>
     </Box>
--- a/src/panels/KeybindsPanel/KeybindsPanel.tsx
+++ b/src/panels/KeybindsPanel/KeybindsPanel.tsx
@@ -1,19 +1,13 @@
 import { Box, Tabs } from '@mantine/core';
-import { useElementSize } from '@mantine/hooks';
-
-import { useWindowSize } from '@/windowmanagement/Window/hooks';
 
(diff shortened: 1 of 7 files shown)


src/panels/KeybindsPanel/ListEntry.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+import { KeybindButtons } from './KeybindButtons';
+
+interface Props {
+  keybind: KeybindInfoType;
+  onClick: () => void;
+  isSelected: boolean;
+}
+
+export function ListEntry({ keybind, onClick, isSelected }: Props) {
```
ylvaselling, line 13: Does this need a more specific name?

src/panels/KeybindsPanel/ListLayout.tsx
```diff
@@ -8,10 +8,15 @@ import { Layout } from '@/components/Layout/Layout';
...
 import { KeybindInfo } from './KeybindInfo';
+import { ListEntry } from './ListEntry';
 
-export function ListLayout() {
+interface Props {
+  // The height of this view, used to get the scrolling behavior right
+  height: number;
+}
+
+export function ListLayout({ height }: Props) {
```
ylvaselling, line 19: This component doesn't need to know the height. It is sufficient if the parent div has the height and the outermost div in this component is "100%".

src/panels/KeybindsPanel/ListLayout.tsx
```diff
@@ -53,63 +58,68 @@ export function ListLayout() {
...
-                variant={
-                  node.identifier === selectedAction?.identifier ? 'filled' : 'light'
-                }
-                fullWidth
-                rightSection={
-                  <KeybindButtons modifiers={node.modifiers} selectedKey={node.key} />
-                }
-                justify={'space-between'}
+    <Group align={'top'} px={'xs'}>
+      <Box h={height} pt={'xs'} flex={2}>
```
ylvaselling, line 62: Remove the box and use the Layout component instead?

```
<Layout pt={'xs'} flex={2}>
```

src/panels/KeybindsPanel/ListLayout.tsx
```diff
@@ -53,63 +58,68 @@ export function ListLayout() {
...
-                size={'md'}
-                variant={
-                  node.identifier === selectedAction?.identifier ? 'filled' : 'light'
-                }
-                fullWidth
-                rightSection={
-                  <KeybindButtons modifiers={node.modifiers} selectedKey={node.key} />
-                }
-                justify={'space-between'}
+    <Group align={'top'} px={'xs'}>
```
ylvaselling, line 61: This one can be set to h={"100%"}

src/panels/KeybindsPanel/KeybindsPanel.tsx
```diff
@@ -21,10 +21,10 @@ export function KeybindsPanel() {
         <Tabs.Panel value={'keyboardLayout'}>
           <KeyboardLayout />
         </Tabs.Panel>
-        <Tabs.Panel value={'listLayout'} h={windowHeight - tabsHeight}>
-          <ListLayout />
+        <Tabs.Panel value={'listLayout'}>
+          <ListLayout height={windowHeight - tabsHeight} />
```
ylvaselling, line 25: You can keep the original version here, list layout doesn't need to know the height

WeirdRubberDuck:
Thanks for the help with the `Layout` component @ylvaselling ! I was struggling a bit with making it work :) 

Waiting for an approve from you and then I'll merge (considering Alex's thumbs up an approve)

ylvaselling (approved):

PR #140, opened 2025-05-14
Language/localize panels

engbergandreas:
started localizing panels, this one includes:
actions 
exoplanets
geolocation
flightcontroller


src/localization/resources.ts
```diff
@@ -7,7 +11,11 @@ import notifications from '@/public/locales/en/notifications.json';
  */
 export const resources = {
   en: {
+    actionpanel: actionPanel,
     components: components,
+    exoplanetspanel: exoplanetsPanel,
+    flightcontrolpanel: flightcontrolPanel,
+    geolocationpanel: geolocationPanel,
     notifications: notifications
```
alexanderbock, line 19: No comment, just a question about the coding style.   How come it is `actionPanel` here , but `ActionIcon` in the ActionsBreadcrumbs.tsx?

engbergandreas: My thinking is that each of these JSON files will be for an entire panel, and then the file itself will have some internal structure that highlights the different sub-components. But since the actionPanel does not have much text, I figured nesting things under sub-names would be unnecessary in this case

src/panels/ActionsPanel/ActionsBreadcrumbs.tsx
```diff
@@ -8,6 +9,7 @@ import { calculateLevelDepth, createPath, getFolders } from './util';
 
 export function ActionsBreadcrumbs() {
   const navigationPath = useAppSelector((state) => state.actions.navigationPath);
+  const { t } = useTranslation('actionpanel');
```
WeirdRubberDuck, line 12: With the change suggested in resources.ts above, this would be 
```suggestion
  const { t } = useTranslation('panels.actions');
```
which I personally really prefer

src/panels/GeoLocationPanel/AnchorPanels/EarthPanel/EarthPanel.tsx
```diff
@@ -145,36 +147,36 @@ export function EarthPanel({ currentAnchor }: Props) {
...
             onKeyDown={(event) => {
               if (event.key === 'Enter') {
                 getPlaces();
               }
             }}
             onChange={(event) => setInputValue(event.target.value)}
-            rightSection={<Button onClick={() => getPlaces()}>Search</Button>}
+            rightSection={
+              <Button onClick={() => getPlaces()}>{t('search.button-label')}</Button>
+            }
```
WeirdRubberDuck, line 165:
```suggestion
            rightSection={
              <Button onClick={getPlaces}>{t('search.button-label')}</Button>
            }
```

src/localization/resources.ts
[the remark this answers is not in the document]
WeirdRubberDuck: Note that the filenames should then be changed to: 

panels/actions.json
panels/exoplanets.json
 ...
 and so on

WeirdRubberDuck (changes requested):

src/localization/resources.ts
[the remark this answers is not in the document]
WeirdRubberDuck: Nvm, looks like having subfolders is not possible without extra tools. 

But at least rename so that "panel" is first? To have them all in one place? With dashes, for readability?

E.g: 
```suggestion
    components: components,
    notifications: notifications,
    panel-actions: actionPanel,
    panel-exoplanets: exoplanetsPanel,
    panel-flightcontrol: flightcontrolPanel,
    panel-geolocation: geolocationPanel,
```
and do the same with the file names. This is how we did it for the Stort installation and it's definitely easier to overview:

image

public/locales/en/flightcontrolpanel.json
```diff
@@ -0,0 +1,30 @@
...
+        }
+      }
+    }
+  },
+  "settings-title": "Settings",
+  "friction-label": "Friction control",
+  "sensitivity": {
+    "label": "Input sensitivity",
+    "description": "Controls how sensitive the touch and mouse inputs are"
+  }
```
WeirdRubberDuck, line 29: Here the naming sceme is also a bit weird. Why do some things live under the "controller" namespace?

public/locales/en/geolocationpanel.json
```diff
@@ -0,0 +1,41 @@
+{
+  "geolocation": {
+    "title": "Geo Location",
+    "select-anchor-label": "Select an anchor",
+    "no-anchor-panel": "Currently there is no data for locations on {{anchor}}"
+  },
```
WeirdRubberDuck, line 6: same comment here. What naming scheme should we use for the actual panel component?   Maybe
```suggestion
  "panel": {
    "title": "Geo Location",
    "select-anchor-label": "Select an anchor",
    "no-anchor-panel": "Currently there is no data for locations on {{anchor}}"
  },
```
or just remove the namespace completely?
```suggestion
  "title": "Geo Location",
  "select-anchor-label": "Select an anchor",
  "no-anchor-panel": "Currently there is no data for locations on {{anchor}}",
```

public/locales/en/geolocationpanel.json
```diff
@@ -0,0 +1,41 @@
...
+  "earth-panel": {
+    "tab-search": "Search Place",
+    "tab-custom-coordinates": "Custom Coordinates",
+
```
WeirdRubberDuck, line 10:
```suggestion
```

public/locales/en/actionpanel.json
```diff
@@ -0,0 +1,5 @@
+{
+  "breadcrumbs-aria-label": "Back",
```
WeirdRubberDuck, line 2: should we maybe also add the title of each planet to the top of each respective json file?  Now that one is not being translated

src/util/fileIOhooks.ts
```diff
@@ -98,7 +98,7 @@ async function openSaveFileDialogInternal(
 export function useSaveLoadJsonFiles(handlePickedFile: (content: JSON) => void) {
   const dispatch = useAppDispatch();
   const openLoadFileDialog = useLoadJsonFile(handlePickedFile);
-  const { t } = useTranslation('notifications', {keyPrefix: 'error'})
+  const { t } = useTranslation('notifications', { keyPrefix: 'error' });
```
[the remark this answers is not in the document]
engbergandreas: Didn't notice, but here appDispatch should be after all hooks. I think so far I've just put the translation as the last useHook

ylvaselling (approved):

public/locales/en/actionpanel.json
```diff
@@ -0,0 +1,5 @@
...
+  "action-button-tooltip": "Local action",
```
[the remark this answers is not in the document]
engbergandreas: I think it make sense to do that in some places but I feel like its a bad idea to do it everywhere. I have a feeling it will be difficult to maintain and keeping the TSX and JSON synced, especially once we start getting more than one language.

public/locales/en/panel-flightcontrol.json
```diff
@@ -0,0 +1,25 @@
+{
+  "toggle-flight-label": "Toggle flight control",
+  "description": {
+    "intro": "Interact with the highlighted area to control the camera.",
+    "mouse-title": "Mouse controls:",
+    "mouse-body": {
+      "text": "Click and drag to rotate. Hold",
+      "keybinds": ["Shift to zoom (y-axis) or roll (x-axis)", "Ctrl to pan"]
```
WeirdRubberDuck, line 8: I think we should try to always linebreak arrays, to make life easier for the person editing the file next time
```suggestion
      "keybinds": [
        "Shift to zoom (y-axis) or roll (x-axis)",
        "Ctrl to pan"
      ]
```

src/localization/resources.ts
```diff
@@ -1,5 +1,9 @@
 import components from '@/public/locales/en/components.json';
 import notifications from '@/public/locales/en/notifications.json';
+import actionPanel from '@/public/locales/en/panel-actions.json';
+import exoplanetsPanel from '@/public/locales/en/panel-exoplanets.json';
+import flightControlPanel from '@/public/locales/en/panel-flightcontrol.json';
+import geolocationPanel from '@/public/locales/en/panel-geolocation.json';
```
WeirdRubberDuck, line 6:
```suggestion
import panelActions from '@/public/locales/en/panel-actions.json';
import panelActions from '@/public/locales/en/panel-exoplanets.json';
import panelFflightControl from '@/public/locales/en/panel-flightcontrol.json';
import panelGeolocation from '@/public/locales/en/panel-geolocation.json';
```
? Makes sure we group the panels in the same order here as above

I you don't do this, let me know because then I should update in my PR :)

src/panels/ExoplanetsPanel/ExoplanetEntry.tsx
```diff
@@ -9,21 +10,24 @@ interface ExoplanetProps {
...
+      <ActionIcon
+        size={'lg'} // Fixed size, adjust as needed
+        key={`${name}button`}
+        variant={isAdded ? 'outline' : 'filled'}
+        style={{ flexGrow: 0 }}
+        onClick={onClick}
+        color={isAdded ? 'red' : 'blue'}
+        aria-label={`${
+          isAdded ? t('remove-system-aria-label') : t('add-system-aria-label')
+        }: ${name}`}
```
WeirdRubberDuck, line 26: Should this use the name as a variable in the string? 

Or, based on our discussion, this should probably be avoided in general when possible. But that means removing the anchor node at the end of this stinrg in the geolocation panel as well: 

```
"no-anchor-panel": "Currently there is no data for locations on {{anchor}}",
```

WeirdRubberDuck (approved):
Looks good - some small comments that need to be addressed to ensure consistency in this PR and between ours. 

Address those and then go ahead and merge :)

src/localization/resources.ts
engbergandreas: I find it easier to read it as actionPanel rather than panelActions?

src/panels/ExoplanetsPanel/ExoplanetEntry.tsx
engbergandreas: Yeah I'm not sure which one I prefer anymore.. if I was to translate the "no-anchor-panel" and it only said `"Currently there is no data for locations on"` I would be confused, since the translation might be different depending on what comes after "on"? 

But in this case it was to simplify the translation and to not have the semicolon be part of the translation and accidently be missed between languages. Though since its an aria label it doesn't matter as much?

public/locales/en/panel-flightcontrol.json
WeirdRubberDuck: Ah ok, ignoer then

src/localization/resources.ts
WeirdRubberDuck: I'm ok with either - so go wit the way you have it then and I'll update on my side :)

src/panels/ExoplanetsPanel/ExoplanetEntry.tsx
WeirdRubberDuck: Ah, good points. Feel free to ignore this comment!

Thanks for thorough answers :)

PR #142, opened 2025-05-15
Language/more panels

engbergandreas:
more panels, this one includes 
scriptlog
sessionrecording
menu


public/locales/en/log.json
```diff
@@ -0,0 +1,22 @@
...
+    "clear-scripts-button-label": "Clear",
+    "refresh-scripts-tooltip": "Refresh script log",
+    "search-script-placeholder": "Search for a script",
+    "more-info": {
+      "first": "This script log has been cleaned for ease of use. Some entries may have been filtered out unintentionally.",
+      "second": "Entries are listed from latest to oldest (top to bottom). Running multiple scripts at once will run them in the order bottom to top.",
+      "third": "To view the complete log or previous runs, check the ",
+      "forth": "OpenSpace/logs ",
+      "fith": "folder."
+    }
```
ylvaselling, line 16: Use [this kind of tag](https://react.i18next.com/latest/trans-component) instead? 

```
<Trans i18nKey="fullTextWithTag" values={{ showExtra }}>
    Here is the text. <span>{{ extra }}</span>
</Trans>
```

public/locales/en/common.json
```diff
@@ -1,3 +1,7 @@
 {
-  "cancel": "Cancel"
```
ylvaselling, line 2: Not sure I understand the common-ness of these. Aren't these just session recording, and the cancel is also in the file menu? Imo it's probably better to not try and generalize too much.

public/locales/en/sessionrecordingpanel.json
```diff
@@ -0,0 +1,29 @@
+{
+  "stop-playback": "Stop Playback",
+  "stop-recording": "Stop Recording",
+  "play-session": {
+    "loop-playback-label": "Loop playback",
+
```
ylvaselling, line 6:
```suggestion
```

ylvaselling (approved):

public/locales/en/log.json
WeirdRubberDuck: Agreeing with this.. number 3, 4 and 5, needs to me merged into one sentence

WeirdRubberDuck: We also need a standard way of dealing with multiple paragraphs. Had a similar case in this PR: https://github.com/OpenSpace/OpenSpace-WebGui/pull/144

I would personally prefer 
```
    "more-info": {
      "part-1": "This script log has been cleaned for ease of use. Some entries may have been filtered out unintentionally.",
      "part-2": "Entries are listed from latest to oldest (top to bottom). Running multiple scripts at once will run them in the order bottom to top.",
      "part-3": "To view the complete log or previous runs, check the ", .....
    }
```
or maybe even just the numbers if that's possible: 
```
    "more-info": {
      "1": "This script log has been cleaned for ease of use. Some entries may have been filtered out unintentionally.",
      "2": "Entries are listed from latest to oldest (top to bottom). Running multiple scripts at once will run them in the order bottom to top.",
      "3": "To view the complete log or previous runs, check the ", .....
    }
```

but typing the numbers in text form like you've done it here gets difficult to overview, IMO.  

Thoughts?

public/locales/en/log.json
```diff
@@ -0,0 +1,22 @@
+{
+  "scriptlog": {
+    "run-scripts-button-label": "Run",
+    "run-scripts-tooltip": "Run all the selected scripts in sequential order, bottom to top",
+    "copy-to-clipboard-error": "No script selected",
+    "clear-scripts-tooltip": "Clears all selected scripts",
+    "clear-scripts-button-label": "Clear",
```
WeirdRubberDuck, line 7: This could be a common label

src/panels/LogPanel/ScriptLogPanel.tsx
```diff
@@ -124,55 +123,49 @@ export function ScriptLogPanel() {
...
             <Text>
-              To view the complete log or previous runs, check the{' '}
+              {t('more-info.third')}
               <Text fs={'italic'} span>
-                OpenSpace/logs
-              </Text>{' '}
-              folder.
+                {t('more-info.forth')}
+              </Text>
+              {t('more-info.fith')}
```
WeirdRubberDuck, line 151: As suggested above, but how about using `<Trans />` here? Suggestion:
 ```
<Trans t={t} i18nKey={'more-info.part-3'} components={{ italics: <Text fs={'italic'} span /> }} />
```

File: public/locales/en/log.json (hunk 1 above)
Comment (ylvaselling) [reply public/locales/en/log.json:16]: I think the <Trans> tag can deal with multiple paragraphs. That way you can have one text but you place the text in multiple <Text> tags

Comment (WeirdRubberDuck) [reply public/locales/en/log.json:16]: Now completely sure what you're suggesting, but there seems to be a couple of ways to do it.  My first thought was to maybe use arrays: https://www.i18next.com/translation-function/objects-and-arrays

This is what ChatGPT suggests as the first option as well, and it had an interesting suggestion for how to create components based on the items in the array: 
image

In addition, it also suggested to use `\n\n` or `<br /><br />` in the actual strings. Or to make the strings themselves HTML (e.g. `"<p>This is the first paragraph.</p><p>This is the second one.</p>"`). However, since we expect users to potentially edit/write these strings, I don't personally think that this is a good idea. It's so easy to forget one of the tags, and would be better if the string ar ejust the text and includes no formatting logic

So, maybe we should go with arrays?  Here, this would be: 
```
    "more-info": [
      "This script log has been cleaned for ease of use. Some entries may have been filtered out unintentionally.",
      "Entries are listed from latest to oldest (top to bottom). Running multiple scripts at once will run them in the order bottom to top.",
      "To view the complete log or previous runs, check the ", .....
    ]
```
and then we can create one text object per item in the array?

File: public/locales/en/common.json (hunk 2 above)
Comment (WeirdRubberDuck) [reply public/locales/en/common.json:2]: This needs some discussion, then. 

Personally, I'm for having strings that are commonly used for certain buttons and components in the `common` file, to make sure that they are consistent across the user interface. It also prevents us/the translator from translate the same word over and over again which could be tedious. But to me, consistency is the main reason. 

Here are some other examples of strings I added to the common.json file as part of PR #144. Let's discuss?
image

File: public/locales/en/log.json (hunk 1 above)
Comment (alexanderbock) [reply public/locales/en/log.json:16]: Arrays seem to be better in my mind as well, if that works.   Getting layouting control into the strings will quickly lead to the problem that the spacing between icons might be different if there are `\n\n` in one language but `\n\n\n` in another, for example

Comment (ylvaselling) [reply public/locales/en/log.json:16]: I'm suggesting this:

```suggestion
    "more-info": "<1>This script log has been cleaned for ease of use. Some entries may have been filtered out unintentionally. </1><1> Entries are listed from latest to oldest (top to bottom). Running multiple scripts at once will run them in the order bottom to top. </1><1> To view the complete log or previous runs, check the <2>OpenSpace/logs</2> folder.<1/>"
```

```
  <Trans i18nextKey={'more-info'} t={t}>
              <Text mb={'xs'}>
                This script log has been cleaned for ease of use. Some entries may have
                been filtered out unintentionally.
              </Text>
              <Text mb={'xs'}>
                Entries are listed from latest to oldest (top to bottom). Running multiple
                scripts at once will run them in the order bottom to top.
              </Text>
              <Text mb={'xs'}>
                To view the complete log or previous runs, check the{' '}
                <Text fs={'italic'} span>
                  OpenSpace/logs
                </Text>{' '}
                folder.
              </Text>
</Trans>
```

public/locales/en/log.json
```diff
@@ -0,0 +1,22 @@
+{
+  "scriptlog": {
+    "run-scripts-button-label": "Run",
+    "run-scripts-tooltip": "Run all the selected scripts in sequential order, bottom to top",
```
alexanderbock, line 4:
```suggestion
    "run-scripts-tooltip": "Run all the selected scripts in sequential order from bottom to top",
```

public/locales/en/menu.json
```diff
@@ -0,0 +1,72 @@
...
+  },
+  "windows-menu": {
+    "title": "Windows",
+    "add-new-window": "Add New Window"
+  },
+  "error": {
+    "title": "Error loading JSON file",
+    "message": {
+      "empty-file": "File is empty",
+      "invalid-length": "Invalid layout file, length does not match",
```
alexanderbock, line 68:
```suggestion
      "invalid-length": "Invalid layout file: Length does not match",
```

public/locales/en/menu.json
```diff
@@ -0,0 +1,72 @@
...
+      "invalid-ids": "Invalid layout file, all IDs must match panel IDs"
```
alexanderbock, line 69:
```suggestion
      "invalid-ids": "Invalid layout file: All IDs must match panel IDs"
```

public/locales/en/sessionrecordingpanel.json
```diff
@@ -0,0 +1,29 @@
...
+    "output-frames": {
+      "label": "Ouput frames",
```
alexanderbock, line 8:
```suggestion
      "label": "Output frames",
```

public/locales/en/sessionrecordingpanel.json
```diff
@@ -0,0 +1,29 @@
...
+      "info": "If checked, the specified number of frames will be recorded as screenshots and saved to disk. Per default, they are saved in the user/screenshots folder. This feature can not be used together with 'loop playback.",
```
alexanderbock, line 9:
```suggestion
      "info": "If checked, the specified number of frames will be recorded as screenshots and saved to disk. Per default, they are saved in the user/screenshots folder. This feature can not be used together with 'loop playback'.",
```

public/locales/en/common.json
WeirdRubberDuck: After a seconds look a tthese, I agree that many of these are not that common. Play, pause, record, resume should be moved to the session recording panel json

engbergandreas:
Closing in favor of https://github.com/OpenSpace/OpenSpace-WebGui/pull/167  https://github.com/OpenSpace/OpenSpace-WebGui/pull/166 https://github.com/OpenSpace/OpenSpace-WebGui/pull/165

PR #146, opened 2025-05-21
Update copy URI component to include URI, and fix URI for actions

WeirdRubberDuck:
I think we can consider this a bug fix 😇 

## Before: 
image
image

## After: 
image
image

### DIFF
--- a/src/components/CopyUriButton/CopyUriButton.tsx
+++ b/src/components/CopyUriButton/CopyUriButton.tsx
@@ -1,5 +1,4 @@
-import { useTranslation } from 'react-i18next';
-import { Code, Group, Stack, Text } from '@mantine/core';
+import { Code, Group } from '@mantine/core';
 
 import { CopyToClipboardButton } from '../CopyToClipboardButton/CopyToClipboardButton';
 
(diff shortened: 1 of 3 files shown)


src/components/CopyUriButton/CopyUriButton.tsx
```diff
@@ -10,15 +10,18 @@ interface Props {
...
-        </Group>
-      )}
-    </>
+    <Stack pt={'sm'}>
+      <Group gap={'xs'} wrap={'nowrap'}>
+        <Text size={'xs'}>{t('label')}:</Text>
+        <Code>{uri}</Code>
+        <CopyToClipboardButton value={uri} />
+      </Group>
+    </Stack>
```
ylvaselling, line 24:
```suggestion
      <Group pt={'sm'} gap={'xs'} wrap={'nowrap'}>
        <Text size={'xs'}>{t('label')}:</Text>
        <Code>{uri}</Code>
        <CopyToClipboardButton value={uri} />
      </Group>
```

ylvaselling (approved):
I think it's ok to remove the "URI" title. 

It would also be nice to style the scrollbar so its a bit smaller :)

image

src/components/CopyUriButton/CopyUriButton.tsx
WeirdRubberDuck: Indeed - it's a remnant from when I was experimenting with a different layout... Good catch :)

engbergandreas (approved):

PR #147, opened 2025-05-23
Issue/3672 scrollable menus

WeirdRubberDuck:
Make long topbar menus scrollable.  Did this by adding a wrapper that can be added to dropdown menu components. Let me know what you think about it  and if you have any better idea - I'm not 100% happy since it seems a bit "hacky" .  

Also tried applying the setting to the all menus using e.g. the theme, but this did not work. so this solution gives us explicit control of each menu dropdown. 

image
image

⚠️ **Also updated to Mantine 8.0.1 and utilized the new sub-menu feature as part of this!** ⚠️ This solves some other issues with navigating the top bar menu using keyboard, and looks a bit cleaner. So I was hoping that it falls under "bug fixing" territory. 

Only found one place where the code had to be updated to work with the version update


src/panels/Menu/TopMenuBar/MenuDropdownWrapper.tsx
```diff
@@ -0,0 +1,47 @@
+import { PropsWithChildren } from 'react';
+import { Menu } from '@mantine/core';
+
+interface Props extends PropsWithChildren {
+  /**
+   * If true, limit the height of the dropdown to the window height minus an optoinal
```
alexanderbock, line 6:
```suggestion
   * If true, limit the height of the dropdown to the window height minus an optional
```

ylvaselling (approved):

engbergandreas (approved):

PR #153, opened 2025-05-27
accounting for pin size in position on night sky map, also overflow

micahnyc:
Closes https://github.com/OpenSpace/OpenSpace/issues/3673

Added the size of the image into the calculation for the marker position.

Added overflow hidden to parent container for when the marker is on the edge.

New: (0,0) and (0,180)
image
image

Old (0,0) and (0,180)
image
image


engbergandreas (approved):
Don't forget to run the linter, if you want to not have to think about it you can automate it on save in vscode for example

PR #144, opened 2025-05-16
Localize panels: Navigation, Keybinds, Scene

WeirdRubberDuck:
* navigation
* keybinds
* scene

Please help check for copy-paste errors

PR #154, opened 2025-06-03
Language/time

engbergandreas:
Adds localization to the time panel

Clarifying the change in `src/panels/TimePanel/types.ts` now computes a month milliseconds as 365/12 * 60 * 60 * 24 so that inputting 1 year in the delta time and then changing it to display months will actually display 12 months / s instead of 11.7 something


public/locales/en/panel-time.json
```diff
@@ -0,0 +1,68 @@
...
+    "Hours_one": "Hour",
+    "Hours_other": "Hours",
+    "Days_one": "Day",
+    "Days_other": "Days",
+    "Weeks_one": "Week",
+    "Weeks_other": "Weeks",
+    "Months_one": "Month",
+    "Months_other": "Months",
+    "Years_one": "Year",
+    "Years_other": "Years"
```
WeirdRubberDuck, line 25: these should not start with a capitla letter

src/panels/TimePanel/TimeInput/TimeInput.tsx
```diff
@@ -354,10 +351,10 @@ export function TimeInput() {
               }}
               variant={'filled'}
             >
-              Set
+              {t('lock-time.button-labels.set')}
             </Button>
             <Button variant={'default'} onClick={() => setUseLock(false)}>
-              Cancel
+              {t('common:cancel')}
```
WeirdRubberDuck, line 357: For some reason, this string is no longer capitalized. It should be

image

public/locales/en/panel-time.json
```diff
@@ -0,0 +1,68 @@
...
+    "error": {
+      "invalid-time": {
+        "title": "Invalid time",
+        "message": "Error trying to set time to an invalid date"
+      },
+      "invalid-date": "New date is outside range (April 20, 271821 BC to Sep 13, 275760 AD)",
+      "invalid-year": "Year '{{value}}' is outside allowed year range (April 20, 271821 BC to Sep 13, 275760 AD)"
+    },
+    "lock-time": {
+      "tooltip": "Lock time updates to prevent automatic changes. Use 'Interpolate' or 'Set' to manually apply the selected date and time.",
```
WeirdRubberDuck, line 53: Seems like it would potentially be a place where we want to refer to the `button-labels` keys instead of repeating the words? To prevent the risk of a mismatch?

src/panels/TimePanel/TimeInput/TimeInput.tsx
WeirdRubberDuck: Actually, the issue is that the key isn't found
image

The issue is the `keyPrefix` being `time-input`, and it seems like this prefix is applied to both the included namespaces. 

One that I used to include the common namespace in #144 is like this: 
```
  const { t: tCommon } = useTranslation('common');
  const { t } = useTranslation('panel-keybinds', { keyPrefix: 'keybind-info' });
```
maybe doing it like that would be enough?  Or do we prefer to do a namespaced solution?

public/locales/en/panel-time.json
engbergandreas: for example: 
```javascript
t(`time-parts.${TimePart.Seconds}`, {count: 1}) 
```

src/panels/TimePanel/SimulationIncrement.tsx
```diff
@@ -41,14 +48,16 @@ export function SimulationIncrement() {
...
-          label={'Display Unit'}
+          label={t('simulation-increment.select-unit-label')}
           value={stepSize}
           data={selectableData}
           allowDeselect={false}
           onChange={(value) => setStepSize(value! as TimePart)}
         />
         <NumericInput
-          label={`${stepSize} / second`}
+          label={`${t(`time-parts.${stepSize}`, {
```
WeirdRubberDuck, line 58: This is a little magic - the fact that the key has to match enum string. It took me a while to actually find where the conversion was happening. 

How about making a function or something in util.ts that maps from the enum string to the correct i18next key? It should also throw if a provided key does not exist (important!)

And, as mentioned above, then make the keys lowercase. This would be a lot more explicit, and hence less magic and more understandable. As well as reduce the risk for errors in the case where we for example decide to change the keys for the enum or add another time unit in the future

WeirdRubberDuck (changes requested):

public/locales/en/panel-time.json
```diff
@@ -0,0 +1,68 @@
...
+    "Years_other": "Years"
+  },
+  "simulation-increment": {
+    "select-unit-label": "Display Unit"
+  },
+  "quick-adjust": {
+    "title": "Quick Adjust"
+  },
+  "delta-time-step-control": {
+    "aria-label": {
```
WeirdRubberDuck, line 34: In another PR you used the plural `aria-labels` when grouping multiple labels like this. I preferred that - please do it here too
```suggestion
    "aria-labels": {
```

public/locales/en/panel-time.json
```diff
@@ -0,0 +1,69 @@
...
+      }
+    }
+  },
+  "time-input": {
+    "error": {
+      "invalid-time": {
+        "title": "Invalid time",
+        "message": "Error trying to set time to an invalid date"
+      },
+      "invalid-date": "New date is outside range ($t(error.valid-time-range))",
```
WeirdRubberDuck, line 50:
```suggestion
      "invalid-date": "New date is outside range $t(error.valid-time-range)",
```

src/panels/TimePanel/DeltaTimeStepControl.tsx
```diff
@@ -62,7 +68,7 @@ export function DeltaTimeStepsControl() {
           disabled={!hasPrevDeltaTimeStep}
           size={'lg'}
           w={'100%'}
-          aria-label={'Set previous delta time step'}
+          aria-label={t('delta-time-step-control.aria-labels.previous-step')}
```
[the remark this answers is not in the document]
engbergandreas: Can't, this one uses `time-part` as well

PR #160, opened 2025-06-09
Disable text selection in the UI

engbergandreas:
Could have sworn I had seen an issue for this, but can't seem to find it now. However, it's possible that we only had it in our notion. idk

### DIFF
--- a/src/components/InfoBox/InfoBox.tsx
+++ b/src/components/InfoBox/InfoBox.tsx
@@ -32,7 +32,9 @@ export function InfoBox({ children, w = 320 }: Props & PropsWithChildren) {
         </ActionIcon>
       </Popover.Target>
 
-      <Popover.Dropdown maw={w}>{children}</Popover.Dropdown>
+      <Popover.Dropdown maw={w} style={{ userSelect: 'text' }}>
+        {children}
+      </Popover.Dropdown>
     </Popover>
   );
 }
--- a/src/App.css
+++ b/src/App.css
@@ -1,0 +1,3 @@
+* {
+  user-select: none;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,6 +12,7 @@ import { RoutesPage } from './pages/RoutesPage';
 import { store } from './redux/store';
 import { cssVariablesResolver, theme } from './theme/mantineTheme';
 
+import './App.css';
 import 'rc-dock/dist/rc-dock-dark.css';
 import '@mantine/dates/styles.css';
 import '@mantine/notifications/styles.css';


alexanderbock:
I believe I just mentioned it in a meeting once

WeirdRubberDuck (changes requested):
Hm, I haven't heard anything anout that we wanted to disable the selection, but maybe that's something you discussed while I was away? 🤔 

I can see how it's annoying in some cases, sure. However, I found myself missing the selection feature in quite a few places, such as the ones marked below, and adding the selection to all the places it might be useful seems like a pain in the a** and something that's very easily forgotten.  

Would it maybe make sense to make all the `Text` components selectable by default?  

 
image

src/panels/Scene/SceneGraphNode/SceneGraphNodeMetaInfo.tsx
```diff
@@ -47,49 +49,61 @@ export function SceneGraphNodeMetaInfo({ uri }: Props) {
...
+        </Text>
+      ],
       [
         'Tags:',
         <Group gap={'xs'}>
           {propertyOwner?.tags.map((tag) => (
             <Pill key={tag}>
               <Flex gap={2}>
-                {tag}
+                <Text className={classes.selectable}>{tag}</Text>
```
WeirdRubberDuck, line 75:
```suggestion
                <span className={classes.selectable}>{tag}</span>
```
Making this a text component borked up the size, and there is no need for this to be a text

src/panels/Scene/SceneGraphNode/SceneGraphNodeMetaInfo.tsx
```diff
@@ -47,49 +49,61 @@ export function SceneGraphNodeMetaInfo({ uri }: Props) {
       [
         'Identifier:',
```
WeirdRubberDuck, line 50: Now that the TD items include text, these should probably be text too. But please make the text small to respect screenspace 

```suggestion
        <Text size={'sm'}>Identifier:</Text>,
```

And add the `size={'sm'}` part to all the other texts in here as well

WeirdRubberDuck (commented):
Selectable parts looks great to me

But the sizing of some texts changes when you made them into text components. PLease fix - see comment :)

src/components/InfoBox/InfoBox.tsx
```diff
@@ -3,6 +3,7 @@ import { useTranslation } from 'react-i18next';
 import { ActionIcon, Popover } from '@mantine/core';
 
 import { InformationIcon } from '@/icons/icons';
+import classes from '@/theme/global.module.css';
```
ylvaselling, line 6: I think we often call this "styles". No biggie tho

ylvaselling (approved):

WeirdRubberDuck (approved):

PR #164, opened 2025-06-11
Add support for saving base64 encoded images into temporary files

alexanderbock:
Adds support for adding base64-encoded images using the Screenspace Renderable window.  Needs the current `master` branch of OpenSpace for a new Lua function to save base64 images

1. Add a "Display name"
2. Paste a base64-encoded image into the "Path or URL to slide"
3. Add slide

An example for a base64 encoded image is in https://github.com/OpenSpace/OpenSpace/issues/3255  but they are regularly provided by Google when doing image searches and start with `data:image/jpeg;base64,` or `data:image/png;base64,`

Currently only jpeg and png encoded images are supported, but I have not seen any other images out in the wild and we don't support loading webp images in the engine (yet).
The base64-encoded image is saved into the temporary folder and the path to that image file is passed into the ScreenSpaceImageLocal code path

A todo where I'd need help with is signalling in a good way that the image was not encoded in a supported image format.

Closes OpenSpace/OpenSpace#3255

### DIFF
--- a/src/types/openspace-api-js.d.ts
+++ b/src/types/openspace-api-js.d.ts
@@ -534,6 +534,12 @@ export interface openspace {
    */
   resolveShortcut: (path: path) => Promise<path>
   /**
+   * This function takes a base64 encoded data string, decodes it and saves the resulting data to the provided filepath.
+
+\\param filePath The location where the data will be saved. Any file that already exists in that location will be overwritten \\param base64Data The base64 encoded data that should be saved to the provided file
+   */
+  saveBase64File: (filepath: path, base64Data: string) => Promise<void>
+  /**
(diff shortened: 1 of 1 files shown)


src/panels/ScreenSpaceRenderablePanel/ScreenSpaceRenderablePanel.tsx
```diff
@@ -37,13 +37,35 @@ export function ScreenSpaceRenderablePanel() {
...
-    const isHttpSlide = slideURL.indexOf('http') === 0;
+    let urlOrPath = slideURL;
+    if (slideURL.startsWith("data:image/")) {
+      let url = slideURL;
+      // Someone tried to paste a base64 encoded image. It starts with the text:
+      // data:image/{png/jpeg};base,
+      // followed by the rest of the image data in base64 encoding
+      url = url.substring("data:image/".length);
+
+      let filetype = url.substring(0, url.indexOf(";"));
```
engbergandreas, line 48:
```suggestion
      const filetype = url.substring(0, url.indexOf(";"));
```

engbergandreas (commented):
adding a suggestion for the todo in a sec

PR #170, opened 2025-06-16
Update search placeholder for exoplanets

WeirdRubberDuck:
To include size of dataset, like we do for skybrowser images

After: 
image

Before:
image


public/locales/en/panel-exoplanets.json
```diff
@@ -1,5 +1,5 @@
 {
-  "exoplanet-search-placeholder": "Star name...",
+  "exoplanet-search-placeholder": "Search {{numberStars}} host stars...",
```
engbergandreas, line 2: make this pluralized?

engbergandreas (approved):

PR #171, opened 2025-06-16
Change basename of repo to "gui" and update the routes accordingly

ylvaselling:
To be used with [this PR](https://github.com/OpenSpace/OpenSpace-WebGuiBackend/pull/23) and this.

 * Changes the base name of the repo to be `gui` as that is how it is set in the asset
 * Redirects all undefined routes to `gui`
 * `gui` is now where the gui is served
 * Changed the routes a bit

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "private": true,
   "version": "0.0.0",
   "type": "module",
+  "homepage": "/gui",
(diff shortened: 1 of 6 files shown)


engbergandreas (commented):
We should also add a link to the routes page from the help menu

PR #175, opened 2025-06-24
Disable navigation and time inputs based on engine mode

WeirdRubberDuck:
In the UI rewrite, we lost the functionality that disables some parts of the UI depending don't the engine mode This fixes some cases for session recording playback and camera paths, so that instead of getting error when trying to interact, it is not possible to e.g. set the time in the first place. 

Changes include: 
* Disable navigation and time changes on session recording playback
  * Added overlays to the Navigation and Time panel, and disabled the `NodeNavigationButton`s while a session recording playback is active. See video below. 
* Disable time change inputs if in camera path.

❓ **Question**: In the previous UI we also highlighted the navigation and time buttons in the bottom bar, both during recording and playback. Is this still something we want to do? 

https://github.com/user-attachments/assets/f10c8da4-2235-4690-b07c-02efd3fbd65a


src/components/RecordingPlaybackOverlay/RecordingPlaybackOverlay.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { useSubscribeToEngineMode } from '@/hooks/topicSubscriptions';
+import { VideocamIcon } from '@/icons/icons';
+import styles from '@/theme/global.module.css';
+import { EngineMode, IconSize } from '@/types/enums';
+
+interface Props {
+  text?: string;
+}
+
+export function RecordingPlaybackOverlay({ text }: Props) {
```
WeirdRubberDuck, line 12: Is this component too sepcific?  I.e. do we want a more general "disable overlay" component rather than one specific to the session recording playback?

ylvaselling: If we need a more general component down the line we can always change it:)

ylvaselling (approved):
👍

engbergandreas:
I think its fine without highlighting the other navigation and time buttons

PR #176, opened 2025-06-24
Fix broken quick-adjust slider

WeirdRubberDuck:
Closes OpenSpace/OpenSpace#3678

### DIFF
--- a/src/panels/TimePanel/SimulationIncrement.tsx
+++ b/src/panels/TimePanel/SimulationIncrement.tsx
@@ -15,6 +15,7 @@ import { Decimals, StepSizes, TimePart } from './types';
 export function SimulationIncrement() {
   const luaApi = useOpenSpaceApi();
   const [stepSize, setStepSize] = useState<TimePart>(TimePart.Seconds);
+  const [beforeAdjust, setBeforeAdjust] = useState<number | null>(null);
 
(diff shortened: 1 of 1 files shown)


ylvaselling:
https://github.com/user-attachments/assets/4f5298cc-70ec-47db-b647-e19c5bc70e09

WeirdRubberDuck:
> > Still some issues; i assume this is still the quick adjust? Or maybe it is not a part of this pr https://github.com/user-attachments/assets/4f5298cc-70ec-47db-b647-e19c5bc70e09
> 
> Interesting, and good catch. Seems like if it is clicked to fast there are still issues, indeed..

This is a bit of a diffiuclt issue to solve, though, since it's an issue with how we're setting the time rather than how the component works. It happens if we manage to trigger a new change in the quick-adjust before the time has started interpolating to the previous value

I made a change that makes it less likely to trigger this issue, but that also leads to it not being possible to set the quick-adjust by just clicking on the slider - one must drag.  Le time know what you think about this @ylvaselling, I'm not sure what a good solution is

alexanderbock:
The user-polling was unanimous in having it snap back to the previous version

WeirdRubberDuck:
Rolled back to just use the first commit, since the updates did not really solve the problem. 

Will create an issue for the bug you noticed @ylvaselling

PR #177, opened 2025-06-25
Add tab to add browser/webpage in screenspace renderables panel

WeirdRubberDuck:
Closes OpenSpace/OpenSpace#3387

Image tab (same as before, but titles made clearer that it is an image)
image

Website tab
image


alexanderbock:
🥳

PR #178, opened 2025-06-26
Adds confirmation modal when properties require one

engbergandreas:
Adds option to have a modal popup when changing a property.

Requires OpenSpace https://github.com/OpenSpace/OpenSpace/pull/3730

### DIFF
--- a/public/locales/en/components.json
+++ b/public/locales/en/components.json
@@ -132,7 +132,7 @@
     },
     "confirmation-modal": {
       "title": "Confirm Change",
-      "description": "You are about to change the value of '{{propertyName}}'. This property requires confirmation. Do you want to proceed?",
+      "description": "You are about to change the value of '{{propertyName}}'. Do you want to proceed?",
       "confirm": "Confirm",
       "cancel": "Cancel"
     }


public/locales/en/components.json
```diff
@@ -129,6 +129,12 @@
     "property-label": {
       "read-only-tooltip": "This property is read-only, meaning that it's not intended to be changed.",
       "read-only-label": "Read-only"
+    },
+    "confirmation-modal": {
+      "title": "Confirm Change",
+      "description": "You are about to change the value of '{{propertyName}}'. This property is marked as sensitive and requires confirmation. Do you want to proceed?",
```
WeirdRubberDuck, line 135: Is "sensitive" the right word here?  Can't come up with a better one, though. "protected"? But maybe that's also confusing...

It's difficult to come up with the description here since we can't generalize based on what type of property we have applied this setting to... Also, depending on the setting, this modal may always show, even for properties that are not marked  as sensititve/protected/dangerous to change.  So, how about just removing this part? 

```suggestion
      "description": "You are about to change the value of '{{propertyName}}'. Do you want to proceed?",
```
or 
```suggestion
      "description": "You are about to change the value of '{{propertyName}}', which requires confirmation. Do you want to proceed?",
```

WeirdRubberDuck (approved):
Looks good to me :) Very clean solution.

Didn't try with the different levels for the show modal settings, but I trust that you checked that it works as expected

public/locales/en/components.json
alexanderbock: I think `This property requires confirmation.` is probably unambiguous enough?

engbergandreas: Yep wasn't too sure on this text, going with `This property requires confirmation.`

ylvaselling:
Maybe the confirmation needs to be set on end for sliders or something? I get that that would be pretty annoying implementation wise though...

Otherwise i think it looks good

https://github.com/user-attachments/assets/8b76c20f-1092-44a2-a622-d9a96aa3a818

PR #182, opened 2025-07-02
Show all keybinds mapped to a key in keyboard view 

WeirdRubberDuck:
closes OpenSpace/OpenSpace#3679

Also fixes some styling-related issues when multiple mapped keys were shown. 

Here's what it looks like now. New/updated parts highlighted in pink
1. Scrollbar on keyboard instead of entire window
2. Show how many actions mapped to a certain key
3. Scrollbar on just the bottom part, where the list of mapped actions are (note that I could not get it to work for just the actions list, so the title scrolls too...)
4. Show keybind for each action card (necessary now that we show e.g. the SHIFT+X key on selecting X)

image

### DIFF
--- a/src/panels/KeybindsPanel/FullKeyboard/FullKeyboard.tsx
+++ b/src/panels/KeybindsPanel/FullKeyboard/FullKeyboard.tsx
@@ -52,8 +52,8 @@ export function FullKeyboard({
   }
 
   function getActionsForKey(key: string): KeybindInfoType[] {
-    // Find all action identifiers matching the given key and current modifiers.
-    // If no modifiers are active, it will match all keybinds with the given key.
+    // Find all action identifiers matching the given key and current modifiers
+    // If no modifiers are active, it will match all keybinds with the given key
     const currentKeybinds = keybinds.filter(
       (keybind) =>
         keybind.key &&
--- a/src/panels/KeybindsPanel/KeybindInfo.tsx
+++ b/src/panels/KeybindsPanel/KeybindInfo.tsx
@@ -28,7 +28,7 @@ export function KeybindInfo({ action }: Props) {
             [
               <Text size={'sm'}>{t('info')}:</Text>,
               // Using a custom, large, width for the longest text to make the
-              // table look better, and multiple stavked tables align better
+              // table look better, and multiple stacked tables align better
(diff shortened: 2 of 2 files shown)


alexanderbock:
How hard would it be to have the keyboard view shrink proportionally instead of having a scrollbar?

src/panels/KeybindsPanel/FullKeyboard/FullKeyboard.tsx
```diff
@@ -51,19 +51,30 @@ export function FullKeyboard({
     }
   }
 
-  function getActionsForKey(key: string): Action[] {
-    // Find all action identifiers matching the given key and current modifiers
+  function getActionsForKey(key: string): KeybindInfoType[] {
+    // Find all action identifiers matching the given key and current modifiers.
+    // If no modifiers are active, it will match all keybinds with the given key.
```
alexanderbock, line 56:
```suggestion
    // Find all action identifiers matching the given key and current modifiers
    // If no modifiers are active, it will match all keybinds with the given key
```

src/panels/KeybindsPanel/KeybindInfo.tsx
```diff
@@ -2,37 +2,55 @@ import { useTranslation } from 'react-i18next';
...
+      </Group>
       <Table
         data={{
           body: [
-            [`${t('info')}:`, action.documentation],
             [
-              `${t('is-local.title')}:`,
+              <Text size={'sm'}>{t('info')}:</Text>,
+              // Using a custom, large, width for the longest text to make the
+              // table look better, and multiple stavked tables align better
```
alexanderbock, line 31:
```suggestion
              // table look better, and multiple stacked tables align better
```

alexanderbock (approved):

WeirdRubberDuck:
> How hard would it be to have the keyboard view shrink proportionally instead of having a scrollbar?

I didn't look into it in great detail, but it seemed "hard" to me since we are currently setting all the sizes to exact pixels in the CSS file. My impression was that it would be time-consuming and hard to update the CSS to get it to work in a way that looks good when resizing, and the scrollbox seemed like a very easy compromise that at least works better than it did before

alexanderbock:
We'll leave it then :+1:

PR #184, opened 2025-07-03
Feature/add geolocation maps

engbergandreas:
Shows the camera location on a map for most Solar System bodies. Requires https://github.com/OpenSpace/OpenSpace/pull/3741 closes https://github.com/OpenSpace/OpenSpace/issues/257

Can add focus nodes and or fly to locations by clicking on the map. Works for all planets in the Solar System and most major moons. Updated Night Sky panel to use the same map (without the view direction cone, do we want it shown as well?). 
@micahnyc I wasn't sure if the Night Sky -> location still requires the alert when on another planet than Earth, or if that was only for the map you made?

The map is also available in fullscreen at http://localhost:4680/gui/#map so that it can be added as a screenspace renderable. Posting a video in Slack to show the new features.

This PR also allows you to add custom coordinates on any planet, and I removed the selection on the first tab since we only have Earth at the moment anyway. 

image

Obs, there might be some maps that show the location slightly off vs what you see in OpenSpace. If the difference is small, I believe it's due to the projection warping. But if it's something like 30-40+ degrees then let me know


alexanderbock:
Would be nice to clear the mouse location when switching between planets

public/locales/en/components.json
```diff
@@ -72,6 +72,17 @@
       }
     }
   },
+  "map": {
+    "no-map": {
+      "title": "Unavailable Map",
+      "description": "No map found for scene graph node '{{name}}'"
```
alexanderbock, line 78:
```suggestion
      "description": "No map found for '{{name}}'"
```

?
A bit shorter

public/locales/en/components.json
```diff
@@ -72,6 +72,17 @@
...
+    },
+    "aria-labels": {
+      "map": "Map of {{mapName}}",
+      "view-direction": "Cone showing the view direction of the OpenSpace camera",
```
alexanderbock, line 82:
```suggestion
      "view-direction": "Cone showing the view direction of the camera",
```

public/locales/en/panel-geolocation.json
```diff
@@ -16,22 +22,47 @@
...
-      "remove-node-aria-label": "Remove node"
+    "mouse-location": "Mouse Location",
+    "altitude-label": "$t(custom-coordinates.altitude-label)",
+    "node-name-label": "$t(custom-coordinates.node-name-label)",
+    "node-name-placeholder": "$t(custom-coordinates.node-name-placeholder)"
+  },
+  "notifications": {
+    "duplicate-scene-graph-node": {
+      "title": "Warning",
+      "description": "Cannot add focus node, a node with identifier '{{identifier}}' already exists."
```
alexanderbock, line 65:
```suggestion
      "description": "Cannot add focus node. A node with identifier '{{identifier}}' already exists."
```

src/components/Map/data.ts
```diff
@@ -0,0 +1,25 @@
+type Source = string;
+export const MapData: Record<string, Source> = {
```
alexanderbock, line 2: Instead of having the images in the repository, is there a way to use a sync folder for these instead?  Downloading the maps once with a version number and hosting that sync folder instead?

This gets quite specifically into adding data files into repositories

src/panels/GeoLocationPanel/Search/util.ts
```diff
@@ -0,0 +1,59 @@
+import { computeDistanceBetween, LatLng } from 'spherical-geometry-js';
+
+import { ArcGISJSON, Extent, MatchedLocation } from './types';
+
+export const ET_URL = 'http://localhost:3000/1/search/';
```
alexanderbock, line 5:
```suggestion
export const ET_URL = 'https://geocode.openspaceproject.com/1/search/';
```

alexanderbock (commented):
Updating the URL

src/panels/GeoLocationPanel/Search/GeoSearch.tsx
```diff
@@ -0,0 +1,107 @@
...
+export function GeoSearch({
+  onClick,
+  onHover,
+  search
+}: {
+  onClick: (lat: number, long: number, altitude: number, name: string) => void;
+  onHover: (lat: number, long: number) => void;
+  search: string;
+  mah?: number;
+}) {
```
WeirdRubberDuck, line 29:
```suggestion
interface Props {
  onClick: (lat: number, long: number, altitude: number, name: string) => void;
  onHover: (lat: number, long: number) => void;
  search: string;
}

export function GeoSearch({ onClick, onHover, search }: Props) {
```
Note that I removed `mah` from the props since it was seemingly not used

src/panels/GeoLocationPanel/Search/GeoSearch.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    const altitudeET =
+                      place?.diameter &&
+                      calculateAltitudeExtraTerrestial(
+                        centerLatitude,
+                        centerLongitude,
+                        place.diameter
+                      );
+                    const altitudeEarth =
+                      place?.extent && calculateAltitudeEarth(place.extent);
+                    const altitude = altitudeET ?? altitudeEarth ?? 0;
```
WeirdRubberDuck, line 88: This code is quite hard to read and understand. Could it be simplified somehow?

src/panels/GeoLocationPanel/Search/SearchOverlay.tsx
```diff
@@ -0,0 +1,75 @@
...
+  setCustomName
+}: {
+  search: string;
+  h: number;
+  visible: boolean;
+  close: () => void;
+  setMouseMarker: ({ x, y }: { x: number; y: number }) => void;
+  setCoordinates: (coordinates: { lat: number; long: number; alt: number }) => void;
+  setCustomName: (name: string) => void;
+}) {
```
WeirdRubberDuck, line 22:
```suggestion
interface Props {
  search: string;
  h: number;
  visible: boolean;
  close: () => void;
  setMouseMarker: ({ x, y }: { x: number; y: number }) => void;
  setCoordinates: (coordinates: { lat: number; long: number; alt: number }) => void;
  setCustomName: (name: string) => void;
}

export function SearchOverlay({
  search,
  visible,
  close,
  h,
  setMouseMarker,
  setCoordinates,
  setCustomName
}: Props) {
```

Or, considering how many times it's now done this way, have you guys made some code style decision related to this that has not been communicated?

src/panels/GeoLocationPanel/Search/SearchOverlay.tsx
```diff
@@ -0,0 +1,75 @@
...
+                  bg={'dark.8'}
+                  style={{ borderRadius: '8px', overflow: 'auto' }}
+                  ref={ref}
+                  mah={h}
+                >
+                  <GeoSearch
+                    search={search}
+                    onHover={(lat, long) => {
+                      coordsToMapCoords(lat, long);
+                    }}
```
WeirdRubberDuck, line 56:
```suggestion
                    onHover={(lat, long) => coordsToMapCoords(lat, long)}
```

src/panels/GeoLocationPanel/Search/util.ts
WeirdRubberDuck: Realised from other files that is it Extra Terrestrial :P  But still, please write that out in this variable name

WeirdRubberDuck (changes requested):
More comments, sorry. But Alex posted the URL so I could try out the search feature

1. Could we brighten or use an alternative map image for earth?  it is quite difficult to see where the shaded night area is located
image

1. What's the point of the filterlist in the search results overlay?  To me it seems like an extra redundant step that disconnects my search results from the search query 🤔 If I want to filter the results, I would just add more words to my search, so it also seems a bit redundant? I personally find the current behavior quite confusing and overly complicated

1. The search field should be clearable, e.g. have a clear button the same way the filter lists do

1. The search field behaves weirdly on blur. Maybe it should clear the results on blur? 
    - For example, search for "Rome", press enter
    - Write something else, like "jgldfkj"
    - Click anywhere else in the UI to blur the input. Notice how it still says "jgldfkj"
    - Click the search input again to focus it. Notice how the search results are still Rome. This is really confusing if I have been interacting with other parts of the UI for more than a few seconds

PR #186, opened 2025-07-07
Feature/advanced panels

ylvaselling:
* Moves ui visibility logic from hook to redux
* Unifying of names in local redux store
* Optional setting to mark panel as advanced (scriptlog)
* Makes `useMenuItems` hook return a union of the types for the data so we don't need to access `menuItemsDB` so often
* Filtering of menu items in the `useMenuItems` hook


WeirdRubberDuck (approved):
Look good!  

Wondering if we want some type of visual indicator in the Windows menu to show if a panel is advanced?

PR #188, opened 2025-07-24
Add tooltips to truncated text

WeirdRubberDuck:
Adds a new component that automatically adds a tooltip to any truncated texts. 
Closes OpenSpace/OpenSpace#3750

The tooltip is displayed only when the text is truncated

Works for both one-line truncation and texts with line clamp:
image image

This is particularly for added geo locations, that have very long names: 
image


alexanderbock (approved):
:+1:

PR #208, opened 2026-03-30
Remove fade value and use animation for checkbox

ylvaselling:
Switches the fading checkbox from this:

https://github.com/user-attachments/assets/3e4cea28-f1b9-45a3-8b7a-8ffad8a4cb98

To this:

https://github.com/user-attachments/assets/99d697a5-7654-478f-b0c4-51348db3c2f1

This has the benefit of:
* Feels more responsive
* More performant as it doesn't need the fade value

### DIFF
--- a/src/components/PropertyOwner/FadeIcon.tsx
+++ b/src/components/PropertyOwner/FadeIcon.tsx
@@ -1,25 +1,28 @@
+import styles from './FadeIcon.module.css';
+
 interface Props {
-  value: number; // 0–1
   size?: number;
   color?: string;
 }
 
-export function FadeIcon({
-  value,
-  size = 14,
-  color = 'var(--mantine-primary-color-5)'
-}: Props) {
-  const clamped = Math.max(0, Math.min(1, value));
-
+export function FadeIcon({ size = 24, color = 'var(--mantine-primary-color-5)' }: Props) {
   return (
-    <svg width={size} height={size} viewBox={`0 0 ${size} ${size}`}>
-      <rect
-        width={size}
-        height={size}
-        fill={color}
-        opacity={clamped}
-        style={{ transition: 'opacity 150ms ease' }}
-      />
-    </svg>
+    <div
+      style={{
+        width: size,
+        height: size,
+        flexShrink: 0,
+        color,
+        backgroundImage: `repeating-linear-gradient(
+          -45deg,
+          ${color},
+          ${color} 2.5px,
+          transparent 2.5px,
+          transparent 7px
+        )`,
+        animation: `${styles.slide} 1000ms linear infinite`,
+        transition: 'opacity 150ms ease'
+      }}
+    />
   );
 }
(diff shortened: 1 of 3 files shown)


alexanderbock (approved):

WeirdRubberDuck (commented):
Nice!  The animation shows an ugly "jack" when the icon size is increased, however. It would be great if we could fix this now rather than later, to make the component more usable. 

It might be solvable by just increasing the background size to make sure the edge is never visible

Here's how it looks for an `md` sized icon: 

https://github.com/user-attachments/assets/8e2b843d-87b1-42ae-a475-cdb0bf438967


Otherwise, it looks great

ylvaselling:
> Nice! The animation shows an ugly "jack" when the icon size is increased, however. It would be great if we could fix this now rather than later, to make the component more usable.
> 
> It might be solvable by just increasing the background size to make sure the edge is never visible
> 
> Here's how it looks for an `md` sized icon:
> incon.mp4
> 
> Otherwise, it looks great

👍 Fixed it, want to take a look before I merge?

src/components/PropertyOwner/VisiblityCheckbox.tsx
```diff
@@ -55,11 +51,9 @@ export function PropertyOwnerVisibilityCheckbox({
     }
   }
 
-  if (visibility === 'Fading' && fade !== undefined) {
+  if (visibility === 'Fading') {
     return (
-      <ActionIcon size={IconSize.sm} onClick={() => updateValue(!checked, isImmediate)}>
-        <FadeIcon value={fade} size={IconSize.sm} />
-      </ActionIcon>
+      <FadeIcon size={IconSize.xl} onClick={() => updateValue(!checked, isImmediate)} />
```
WeirdRubberDuck:
```suggestion
      <FadeIcon size={IconSize.sm} onClick={() => updateValue(!checked, isImmediate)} />
```
oops :)

WeirdRubberDuck (approved):

src/components/PropertyOwner/FadeIcon.tsx
```diff
@@ -1,25 +1,43 @@
...
-  value,
-  size = 14,
-  color = 'var(--mantine-primary-color-5)'
+  size = 20,
+  color = 'var(--mantine-primary-color-5)',
+  onClick
 }: Props) {
-  const clamped = Math.max(0, Math.min(1, value));
+  const stripeWidth = size * 0.25;
+  const repeat = size * 0.5;
```
WeirdRubberDuck:
```suggestion
  const stripeWidth = size * 0.15;
  const repeat = size * 0.3;
```
Up to you, but this looks more like the stripes did before the change that supports multiple scales. Now they are quite a bit wider

src/components/PropertyOwner/VisiblityCheckbox.tsx
ylvaselling: ops indeed :D

PR #209, opened 2026-03-30
Add batching to the skybrowser topic

ylvaselling:
Adds the newly introduced batching to the SkyBrowser topic updates

### DIFF
--- a/src/redux/skybrowser/skybrowserMiddleware.ts
+++ b/src/redux/skybrowser/skybrowserMiddleware.ts
@@ -5,6 +5,7 @@ import { api } from '@/api/api';
 import { onCloseConnection } from '@/redux/connection/connectionSlice';
 import { AppStartListening } from '@/redux/listenerMiddleware';
 import { ConnectionStatus } from '@/types/enums';
+import { Batcher } from '@/util/batcher';
 
 import {
   resetSkyBrowser,
@@ -26,9 +27,15 @@ export const setupSubscription = createAsyncThunk(
       event: 'start_subscription'
     });
 
+    function updateData(data: Partial<SkyBrowserUpdate>) {
+      thunkAPI.dispatch(updateSkyBrowser(data));
+    }
+
+    const batcher = new Batcher<SkyBrowserUpdate>(updateData);
+
     (async () => {
-      for await (const data of skybrowserTopic.iterator()) {
-        thunkAPI.dispatch(updateSkyBrowser(data as SkyBrowserUpdate));
+      for await (const data of skybrowserTopic.iterator() as AsyncGenerator<SkyBrowserUpdate>) {
+        batcher.add(data);
       }
     })();
     thunkAPI.dispatch(subscriptionIsSetup());
--- a/src/redux/skybrowser/skybrowserMiddleware.ts
+++ b/src/redux/skybrowser/skybrowserMiddleware.ts
@@ -34,7 +34,7 @@ export const setupSubscription = createAsyncThunk(
     const batcher = new Batcher<SkyBrowserUpdate>(updateData);
 
     (async () => {
-      for await (const data of skybrowserTopic.iterator() as AsyncGenerator<SkyBrowserUpdate>) {
+      for await (const data of skybrowserTopic.iterator() as AsyncIterable<SkyBrowserUpdate>) {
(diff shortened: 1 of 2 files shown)


src/redux/skybrowser/skybrowserMiddleware.ts
```diff
@@ -26,9 +27,15 @@ export const setupSubscription = createAsyncThunk(
...
+    function updateData(data: Partial<SkyBrowserUpdate>) {
+      thunkAPI.dispatch(updateSkyBrowser(data));
+    }
+
+    const batcher = new Batcher<SkyBrowserUpdate>(updateData);
+
     (async () => {
-      for await (const data of skybrowserTopic.iterator()) {
-        thunkAPI.dispatch(updateSkyBrowser(data as SkyBrowserUpdate));
+      for await (const data of skybrowserTopic.iterator() as AsyncGenerator<SkyBrowserUpdate>) {
```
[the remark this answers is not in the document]
engbergandreas: Yeah should be AsyncIterable instead of AsyncGenerator

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -54,15 +58,16 @@ export const skyBrowserSlice = createSlice({
...
-
-        // Derived state for easier access
-        state.browserIds = Object.keys(action.payload.browsers) ?? [];
-        state.browserColors = state.browserIds.map(
-          (id) => action.payload.browsers[id].color
-        );
-        state.browserNames = state.browserIds.map(
-          (id) => action.payload.browsers[id].name
-        );
+        if (action.payload.browsers !== undefined) {
```
engbergandreas: With the above `if (action.payload.browsers)` check, this one is redundant, no?

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -54,15 +58,16 @@ export const skyBrowserSlice = createSlice({
...
+          // Derived state for easier access
+          state.browserIds = Object.keys(action.payload.browsers) ?? [];
```
engbergandreas: Object.keys should already return an empty array if browsers is empty
```suggestion
          state.browserIds = Object.keys(action.payload.browsers);
```

src/redux/skybrowser/skybrowserSlice.ts
```diff
@@ -54,15 +58,16 @@ export const skyBrowserSlice = createSlice({
...
+          state.browserColors = state.browserIds.map(
+            (id) => action.payload.browsers?.[id].color ?? [255, 255, 255]
```
engbergandreas: Actually surprised that TypeScript can't infer that the browsers is not undefined here. Am I missing something? 
Doing the following is creating squiggly lines :O 
```
            (id) => action.payload.browsers[id].color ?? [255, 255, 255]
```

engbergandreas: Okay the solution would be to create a variable and then TypeScript can infer correctly..
i.e., 
```ts
const browsers = action.payload.browsers;
...
if (browsers && state.selectedBrowserId in browsers) {
          state.browserIds = Object.keys(browsers);
          state.browserColors = state.browserIds.map(
            (id) => browsers[id].color ?? [255, 255, 255]
          );
}
```

engbergandreas: Also when doing this, is the `SkyBrowserBrowser` interface correctly typed here, since the check color ?? [255, 255, 255] indicates that color might be undefined now. Same goes for name. Or are these always well defined? If so we could remove the `??` check with the above.

engbergandreas (commented):
Lol realised now that I'm basically reiterating what Copilot already found 🤷‍♂️

src/redux/skybrowser/skybrowserMiddleware.ts
```diff
@@ -40,7 +47,7 @@ function tearDownSubscription() {
     return;
   }
   skybrowserTopic.talk({
-    event: 'stop_supscription'
+    event: 'stop_subscription'
```
WeirdRubberDuck: Good catch

WeirdRubberDuck (approved):

PR #210, opened 2026-03-30
Add readme for arrow function returns

ylvaselling:
Lmk if you agree

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,27 @@ We apply a custom theme to the Mantine components. If you are building a separat
    - useEffect
    - useMemo, useCallback
    - functions
+  
+## A note on arrow function returns
+Use the implicit return only when something actually is returned; for example:
+
+```
+// ✅ 
+const double = n => n * 2;
+const getLabel = user => user.name;
+const doubled = nums.map(n => n * 2);
+```
+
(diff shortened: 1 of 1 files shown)


alexanderbock (approved):

WeirdRubberDuck (approved):

WeirdRubberDuck:
Just realized: Do we want to make an exception for things like `onClick` functions on components? In those cases, it is already quite clear that there is not value being retuned, I guess, and adding the extra lines makes the rendered component harder to overview (at least in my opinion)

For example: 
```ts
  if (visibility === 'Fading') {
    return (
      <ActionIcon onClick={() => updateValue(!checked, isImmediate)}>
        <FadeIcon  />
      </ActionIcon>
    );
  }
```
compared to:
```ts
  if (visibility === 'Fading') {
    return (
      <ActionIcon
        onClick={() => {
          updateValue(!checked, isImmediate);
        }}
      >
        <FadeIcon />
      </ActionIcon>
    );
  }
```
In this case, I'm not a fan of the extra clutter that the `{}` and `;` add, but that's only my personal opinion

@ylvaselling  @engbergandreas  @alexanderbock

PR #211, opened 2026-04-07
Add new button to create a new folder when saving screenshots

### LINKED ISSUE #3680: More user-friendly name to reset the screenshot date folder
Right now, when screenshot-recording multiple session recordings, there is a workaround to toggle the `RenderEngine.ScreenshotUseDate` property to force a new folder to be created.  Users (understandably) don't know about it, but it is a useful feature, so we should make it available from the SessionRecording panel.

Envisioning something akin to a checkbox "Save in new folder", but there are probably better options

### LINKED ISSUE #3777: Frames from multiple session recording end up the same folder
Steps to reproduce:
1. Output frames form a session recording
2. Immediately after, output the frames for another session recording
3. Notice how the screenshots end up in the same folder in OpenSpace/user/screenshots, and it's impossible to distinguish where one recording ends and the next starts... :(

Am I crazy, or didn't it used to be that a new folder was created? When did this break?

alexanderbock:
Closes OpenSpace/OpenSpace#3777
Closes OpenSpace/OpenSpace#3680

image

### DIFF
--- a/src/panels/SessionRecordingPanel/PlaySession.tsx
+++ b/src/panels/SessionRecordingPanel/PlaySession.tsx
@@ -32,6 +32,7 @@ export function PlaySession() {
   const [loopPlayback, setLoopPlayback] = useState(false);
   const [shouldOutputFrames, setShouldOutputFrames] = useState(false);
   const [outputFramerate, setOutputFramerate] = useState(60);
+  const [shouldOutputUseNewFolder, setShouldOutputUseNewFolder] = useState(false);
   const { latestFile, hideGuiOnPlayback, hideDashboardsOnPlayback } = useAppSelector(
     (state) => state.sessionRecording.settings
   );
@@ -106,7 +107,7 @@ export function PlaySession() {
     }
   }
 
-  function onShouldUpdateFramesChange(shouldOutputFrames: boolean): void {
+  function onShouldOutputFramesChange(shouldOutputFrames: boolean): void {
     if (shouldOutputFrames) {
       setShouldOutputFrames(true);
       setLoopPlayback(false);
@@ -131,7 +132,7 @@ export function PlaySession() {
           <BoolInput
             label={t('output-frames.label')}
             value={shouldOutputFrames}
-            onChange={onShouldUpdateFramesChange}
+            onChange={onShouldOutputFramesChange}
             disabled={!isIdle}
             info={t('output-frames.tooltip')}
           />
@@ -144,6 +145,17 @@ export function PlaySession() {
             disabled={!shouldOutputFrames || !isIdle}
           />
         </Group>
+        {shouldOutputFrames && (
+          <BoolInput
+            label={t('output-new-folder.label')}
+            value={shouldOutputUseNewFolder}
+            onChange={(shouldOutputUseNewFolder) =>
+              setShouldOutputUseNewFolder(shouldOutputUseNewFolder)
+            }
+            info={t('output-new-folder.tooltip')}
+            disabled={isPlayingBack}
+          />
+        )}
         <BoolInput
           label={t('hide-gui-on-playback.label')}
           value={hideGuiOnPlayback}
@@ -202,6 +214,7 @@ export function PlaySession() {
             loopPlayback={loopPlayback}
             shouldOutputFrames={shouldOutputFrames}
             outputFramerate={outputFramerate}
+            shouldOutputUseNewFolder={shouldOutputUseNewFolder}
           />
         </Group>
         <Group gap={'xs'}>
--- a/src/panels/SessionRecordingPanel/PlaySession.tsx
+++ b/src/panels/SessionRecordingPanel/PlaySession.tsx
@@ -153,7 +153,7 @@ export function PlaySession() {
               setShouldOutputUseNewFolder(shouldOutputUseNewFolder)
             }
             info={t('output-new-folder.tooltip')}
-            disabled={isPlayingBack}
+            disabled={!isIdle}
           />
         )}
         <BoolInput
--- a/src/panels/SessionRecordingPanel/PlaySession.tsx
+++ b/src/panels/SessionRecordingPanel/PlaySession.tsx
@@ -149,9 +149,7 @@ export function PlaySession() {
           <BoolInput
             label={t('output-new-folder.label')}
             value={shouldOutputUseNewFolder}
-            onChange={(shouldOutputUseNewFolder) =>
-              setShouldOutputUseNewFolder(shouldOutputUseNewFolder)
-            }
+            onChange={setShouldOutputUseNewFolder}
             info={t('output-new-folder.tooltip')}
             disabled={!isIdle}
           />
--- a/src/panels/SessionRecordingPanel/PlaySession.tsx
+++ b/src/panels/SessionRecordingPanel/PlaySession.tsx
@@ -128,32 +128,30 @@ export function PlaySession() {
           onChange={onLoopPlaybackChange}
           disabled={!isIdle}
         />
-        <Group>
-          <BoolInput
-            label={t('output-frames.label')}
-            value={shouldOutputFrames}
-            onChange={onShouldOutputFramesChange}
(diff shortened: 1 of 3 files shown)


src/panels/SessionRecordingPanel/Playback/PlaybackPlayButton.tsx
```diff
@@ -37,6 +39,15 @@ export function PlaybackPlayButton({
     dispatch(showGUI(false));
 
     if (shouldOutputFrames) {
+      if (shouldOutputUseNewFolder) {
+        // Toggling the property will cause a new folder to be created and used
+        const prev = (await luaApi?.propertyValue(
+          'RenderEngine.ScreenshotUseDate'
+        )) as boolean;
```
ylvaselling: prev can be undefined, so we should probably handle that

ylvaselling: since the property is set as well, use the `useProperty` hook for "RenderEngine.ScreenshotUseDate" and the top of this file and use the setFunction to set the property

src/panels/SessionRecordingPanel/PlaySession.tsx
```diff
@@ -144,6 +145,17 @@ export function PlaySession() {
             disabled={!shouldOutputFrames || !isIdle}
           />
         </Group>
+        {shouldOutputFrames && (
+          <BoolInput
+            label={t('output-new-folder.label')}
+            value={shouldOutputUseNewFolder}
+            onChange={(shouldOutputUseNewFolder) =>
+              setShouldOutputUseNewFolder(shouldOutputUseNewFolder)
+            }
```
ylvaselling: can be
```suggestion
            onChange={setShouldOutputUseNewFolder}
```

ylvaselling (changes requested):
This does not close #3777 because the folder is called the date of today, which means the screenshots go in the same folder? I tried toggling the checkbox as well but multiple session recordings end up in the same folder, with the name of today's date

alexanderbock:
> This does not close #3777 because the folder is called the date of today, which means the screenshots go in the same folder? I tried toggling the checkbox as well but multiple session recordings end up in the same folder, with the name of today's date

Mhh.. not sure I understand?  The dates should have the hours and minutes in the folder as well?  I think we might want to add seconds to it?
image

src/panels/SessionRecordingPanel/PlaySession.tsx
```diff
@@ -144,6 +145,15 @@ export function PlaySession() {
...
         </Group>
+        {shouldOutputFrames && (
+          <BoolInput
+            label={t('output-new-folder.label')}
+            value={shouldOutputUseNewFolder}
+            onChange={setShouldOutputUseNewFolder}
+            info={t('output-new-folder.tooltip')}
+            disabled={!isIdle}
+          />
+        )}
```
WeirdRubberDuck: How about always showing this checkbox and just having it disabled if shouldOutputFrames is not true? Similarly to the framerate above

It's generally better to disable rather than have UI elements "pop up" with state changes, especially for accessibility reasons

```suggestion
        <BoolInput
          label={t('output-new-folder.label')}
          value={shouldOutputUseNewFolder}
          onChange={setShouldOutputUseNewFolder}
          info={t('output-new-folder.tooltip')}
          disabled={!shouldOutputFrames || !isIdle}
        />
```

WeirdRubberDuck (changes requested):
~~I'm experiencing the same issue that Ylva is explaining - there is a new folder created for the current date, with the timestamp when I created it. However, when I make a new recording, no new folder is created.~~   NVM pulling the last changes seems to have fixed it. Thought I had it all, sorry

Another wis h is that now that we have another button related to the outputting of frames, that we group these setting together a bit more. How baout something like this? 

Image

And if you wish, here's how the code for that looks like. 
```ts
        <BoolInput
          label={t('output-frames.label')}
          value={shouldOutputFrames}
          onChange={onShouldOutputFramesChange}
          disabled={!isIdle}
          info={t('output-frames.tooltip')}
        />
        <Stack ml={'xs'} mb={'xs'}>
          <NumericInput
            value={outputFramerate}
            placeholder={t('framerate.placeholder')}
            aria-label={t('framerate.aria-label')}
            label={'Framerate'} // TODO should be translated
            onEnter={(value) => setOutputFramerate(value)}
            disabled={!shouldOutputFrames || !isIdle}
          />
          <BoolInput
            label={t('output-new-folder.label')}
            value={shouldOutputUseNewFolder}
            onChange={setShouldOutputUseNewFolder}
            info={t('output-new-folder.tooltip')}
            disabled={!shouldOutputFrames || !isIdle}
          />
        </Stack> 
```

src/panels/SessionRecordingPanel/Playback/PlaybackPlayButton.tsx
```diff
@@ -37,6 +44,12 @@ export function PlaybackPlayButton({
     dispatch(showGUI(false));
 
     if (shouldOutputFrames) {
+      if (shouldOutputUseNewFolder) {
+        // Toggling the property will cause a new folder to be created and used
+        setUseDate(false);
+        setUseDate(useDate ?? true);
```
WeirdRubberDuck: This seems a little bit like a hack 🤔 Should it not be enough that this property is set to either true or false?

alexanderbock: Fixed with a new master where that functionality is made into a trigger property

src/panels/SessionRecordingPanel/Playback/PlaybackPlayButton.tsx
```diff
@@ -37,6 +44,10 @@ export function PlaybackPlayButton({
     dispatch(showGUI(false));
 
     if (shouldOutputFrames) {
+      if (shouldOutputUseNewFolder) {
+        useNewScreenshotFolder(null);
```
WeirdRubberDuck: This is generating an error as functions starting with `use` are reserved for hooks. Rename the function to fix it: 
```suggestion
        setUseNewScreenshotFolder(null);
```

src/panels/SessionRecordingPanel/Playback/PlaybackPlayButton.tsx
```diff
@@ -12,19 +13,25 @@ interface Props extends ButtonProps {
...
   loopPlayback,
   outputFramerate,
+  shouldOutputUseNewFolder,
   ...props
 }: Props) {
   const { t } = useTranslation('panel-sessionrecording', { keyPrefix: 'button-labels' });
 
   const luaApi = useOpenSpaceApi();
   const dispatch = useAppDispatch();
+  const [_, useNewScreenshotFolder] = useProperty(
```
WeirdRubberDuck:
```suggestion
  const [, setUseNewScreenshotFolder] = useProperty(
```
Unused variables lead to a linting error

WeirdRubberDuck (approved):
Works and looks good now

Maybe a bit confusing that when unchecking the checkbox, it keeps using the latest folder (with seconds in the name), which was set to be a "new" folder just before. But it's also just the way it works I guess

src/panels/SessionRecordingPanel/Playback/PlaybackPlayButton.tsx
```diff
@@ -12,19 +13,25 @@ interface Props extends ButtonProps {
...
   ...props
 }: Props) {
   const { t } = useTranslation('panel-sessionrecording', { keyPrefix: 'button-labels' });
 
   const luaApi = useOpenSpaceApi();
   const dispatch = useAppDispatch();
+  const [, setUseNewScreenshotFolder] = useProperty(
+    'TriggerProperty',
+    'RenderEngine.UseNewScreenshotFolder'
+  );
```
ylvaselling, line 33: There is a hook order list in the Readme :) This should go above the dispatch

ylvaselling (approved):

PR #216, opened 2026-05-08
Update WebGui using new types and API functions

engbergandreas:
Description inc ™️


src/api/LuaApiContext.tsx
```diff
@@ -1,3 +1,4 @@
 import { createContext } from 'react';
+import { OpenSpaceLibrary } from 'openspace-api-js/generated';
```
WeirdRubberDuck: As we discussed, this path (and others) should preferably be:

```suggestion
import { OpenSpaceLibrary } from 'openspace-api-js/types';
```

As an API user, I don't care whether this is "generated", but I would need to know that this is a type

src/types/types.ts
```diff
@@ -14,55 +14,26 @@ export type LanguageInfo = {
...
   [key: Uri]: AnyProperty | undefined;
 }
 
 export interface PropertyOwners {
-  [key: Uri]: PropertyOwner;
+  [key: Uri]: PropertyOwnerRedux | undefined;
 }
 
-export interface PropertyOwner {
+export interface PropertyOwnerRedux {
```
WeirdRubberDuck, line 45: I find it a bit weird that we use a custom "Redux" version for the property owner type, but I see the point. Add a comment above on why this is needed?  I.e. that this is a version that only includes the URIs for the proeprties and subowners rather than their full data.  

It would be great if we didn't have to call this one "PropertyOwnerRedux" though, since it's the only one we actually use explicitly in the code, it seems. Is this possible?

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -19,7 +20,7 @@ export function VectorProperty({ uri, isInt = false, readOnly }: Props) {
     return (
       <ColorView
         value={value}
-        setPropertyValue={setPropertyValue}
+        setPropertyValue={(v) => setPropertyValue(v as VectorValues)}
```
WeirdRubberDuck: The `as VectorValues` does not seem to be needed here?

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -1,5 +1,6 @@
 import { PropertyProps } from '@/components/Property/types';
 import { useProperty } from '@/hooks/properties';
+import { VectorValues } from '@/types/Property/property';
```
WeirdRubberDuck: If the type is not needed, can this (and maybe the entire type) be removed?

src/components/Property/Types/MatrixProperty.tsx
```diff
@@ -19,7 +20,7 @@ export function MatrixProperty({ uri, readOnly }: PropertyProps) {
     if (!value) {
       return;
     }
-    const v = [...value];
+    const v = [...value] as MatrixValues;
```
WeirdRubberDuck: Same here, `value` seems to already be the correct type. Is the `MatrixValues` type really needed?

src/types/types.ts
engbergandreas: There is a real risk that you would import the wrong `PropertyOwner` since the API exports its own interface under the same name, hence why I decided to rename. But I'm down for alternative names or if we want to change the name of the exported type from the API. But maybe this also gives us the extra nudge to improve the PropertyOwner so that the data is exported in a different format

src/components/Property/Types/VectorProperty/VectorProperty.tsx
engbergandreas: So this will look something like this when we manage to fix the `property.ts`, `propertyGroups.t` and `propertyTypes.ts`. I was not able to merge the generated types with our generic types without introducing problems with TypeScript that blocked building. So we're still running the "old" property type system which is why its "not needed" currently..

src/panels/AssetsFolderPanel/AssetEntry/AssetsEntry.tsx
```diff
@@ -18,7 +18,7 @@ interface Props {
   asset: Asset;
 }
 
-const stateMap: Record<AssetLoadingEvent['State'], AssetLoadState> = {
+const stateMap: Record<AssetLoadingData['state'], AssetLoadState> = {
```
WeirdRubberDuck: With the renaming, I found it a bit hard to understand that this is the data for an event. 

Should it maybe be `AssetLoadingEventData`?  I would prefer this for clarity, but it would be a lot of work to change all of them, though... 😬

src/redux/logging/loggingMiddleware.ts
```diff
@@ -25,43 +26,24 @@ export const handleNotificationLogging = createAction(
...
   async (level: LogLevel, thunkApi) => {
     topic = api.startTopic('errorLog', {
       event: 'start_subscription',
       settings: {
-        logLevel: level
+        logLevel: level,
+        logLevelStamping: true,
+        timeStamping: false,
+        categoryStamping: true,
+        dateStamping: false
```
WeirdRubberDuck, line 41: What's this? :D

engbergandreas: You can now decide what additional data to get back in the callback. So if we want we could get the date and time as well but I didn't see a need for it currently. But could be useful when we decide to add a log panel

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -19,7 +19,7 @@ export function VectorProperty({ uri, isInt = false, readOnly }: Props) {
     return (
       <ColorView
         value={value}
-        setPropertyValue={setPropertyValue}
+        setPropertyValue={(v) => setPropertyValue(v)}
```
WeirdRubberDuck, line 22:
```suggestion
        setPropertyValue={setPropertyValue}
```

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -31,7 +31,7 @@ export function VectorProperty({ uri, isInt = false, readOnly }: Props) {
     return (
       <MinMaxRangeView
         value={value}
-        setPropertyValue={setPropertyValue}
+        setPropertyValue={(v) => setPropertyValue(v)}
```
WeirdRubberDuck, line 34:
```suggestion
        setPropertyValue={setPropertyValue}
```

src/components/Property/Types/VectorProperty/VectorProperty.tsx
```diff
@@ -41,7 +41,7 @@ export function VectorProperty({ uri, isInt = false, readOnly }: Props) {
   return (
     <VectorDefaultView
       disabled={readOnly}
-      setPropertyValue={setPropertyValue}
+      setPropertyValue={(v) => setPropertyValue(v)}
```
WeirdRubberDuck, line 44:
```suggestion
      setPropertyValue={setPropertyValue}
```

src/types/types.ts
```diff
@@ -14,55 +14,26 @@ export type LanguageInfo = {
...
-  color?: [number, number, number, number]; // rgba color, [0, 1]
-}
+export type KeybindModifiers = (keyof Keybind['modifiers'])[];
 
-export type KeybindModifiers = ('super' | 'alt' | 'shift' | 'control')[];
-
-export interface Keybind {
-  action: string;
-  key: string;
+export type KeybindRedux = Pick<Keybind, 'action' | 'key'> & {
```
WeirdRubberDuck, line 23: Add a comment above this type and the PropertyOwnerRedux below

src/redux/logging/loggingMiddleware.ts
WeirdRubberDuck: Cool!  Maybe consider renaming the keys to something more bool-like, such as `useLogLevelStamping`, `useTimeStamping`? 

[Code style](https://docs.openspaceproject.com/releases-v0.21/contribute/development/coding-style.html#the-prefixes-is-has-should-or-can-should-be-used-for-boolean-variables-and-methods)

WeirdRubberDuck (approved):

PR #219, opened 2026-05-20
Fix panel crash when removing a globe layer

### LINKED ISSUE #3790: User interface crashes when deleting globebrowsing layer
To reproduce:
1. Open Earth / Color Layers
2. Open any layer and select "Remove" property to remove the layer

Result:
Image

engbergandreas:
The issue seems to be some kind of render race condition, where when a layer is removed, we get one additional re-render with the now removed layer, which on the engine side is already removed. As a result, we try to get a `propertyOwner` that no longer exists and thus throws an error. 

closes https://github.com/OpenSpace/OpenSpace/issues/3790

requires https://github.com/OpenSpace/OpenSpace-WebGui/pull/216 to be merged before it can be tested. Changes are based on that branch though


alexanderbock (approved):

WeirdRubberDuck (approved):
👍 Tried it too

Great that this fixes the crash. Although it could still be considered unexpected/undesired behavior that an error is logged when the layer is removed. Should we make a new issue for this?

PR #222, opened 2026-06-01
Add translation for menu items title

### LINKED ISSUE #4003: WebGui - Can't translate Menu Item title
When creating a new menu item, I wanted to be able to translate the title but our current structure doens't seem to support that. This would be a nice feature to add.

engbergandreas:
closes https://github.com/OpenSpace/OpenSpace/issues/4003

Use translation titles for File and Help menu. 
Add info icon to File -> About instead of empty icon slot

### DIFF
--- a/src/windowmanagement/data/MenuItems.tsx
+++ b/src/windowmanagement/data/MenuItems.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import i18next from 'i18next';
 
 import {
   BrowserIcon,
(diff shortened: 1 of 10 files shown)


src/windowmanagement/data/MenuItems.tsx
```diff
@@ -69,7 +70,7 @@ export interface MenuItem {
 
 export const menuItemsData: Record<string, MenuItem> = {
   scene: {
-    title: 'Scene',
+    title: () => i18next.t('menu:panels.scene'),
```
alexanderbock, line 73: Is there a special reason why this is not using the

```ts
  const { t } = useTranslation('panel-actions', { keyPrefix: 'actions-breadcrumbs' });
```

from other files?

engbergandreas: Yes the `useTranslation` is a hook and those can only be used inside React components, since this is a DB (with some JSX) we need to use the function version instead

alexanderbock: Ah.. gotcha! :+1:

WeirdRubberDuck (approved):
Looks good to me 👍

PR #227, opened 2026-06-10
Do not display the Flyto/JumpTo/Frame buttons in the kebab menu for SceneGraphNodes that are not focussable

alexanderbock:
The focus button in the menu for non-focussable objects was hidden, but it was still possible to break things by opening the Kebab menu and select the navigation buttons there. This PR doesn't add these buttons if the scene graph node is marked as non-focussable.

Before:
image

After:
image

### DIFF
--- a/src/panels/Scene/SceneGraphNode/SceneGraphNodeMoreMenu.tsx
+++ b/src/panels/Scene/SceneGraphNode/SceneGraphNodeMoreMenu.tsx
@@ -63,8 +63,9 @@ export function SceneGraphNodeMoreMenu({ uri }: Props) {
         >
           {t('pop-out')}
         </Button>
-        {isFocusable && <Divider m={'xs'} />}
         {isFocusable && (
+        <>
+          <Divider m={'xs'} />}
           <Stack gap={'xs'}>
             <Group gap={'xs'}>
               <NodeNavigationButton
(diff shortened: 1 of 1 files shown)


src/panels/Scene/SceneGraphNode/SceneGraphNodeMoreMenu.tsx
```diff
@@ -61,28 +63,30 @@ export function SceneGraphNodeMoreMenu({ uri }: Props) {
...
+            </Group>
+            <Group gap={'xs'}>
+              <NodeNavigationButton
+                type={NavigationType.Frame}
+                identifier={propertyOwner.identifier}
+                showLabel
+              />
+            </Group>
+          </Stack>
+        )}
```
engbergandreas:
```suggestion
        {isFocusable && (
        <>
          <Divider m={'xs'} />}
          <Stack gap={'xs'}>
            <Group gap={'xs'}>
              <NodeNavigationButton
                type={NavigationType.Fly}
                identifier={propertyOwner.identifier}
                showLabel
              />
              <NodeNavigationButton
                type={NavigationType.Jump}
                identifier={propertyOwner.identifier}
                showLabel
              />
            </Group>
            <Group gap={'xs'}>
              <NodeNavigationButton
                type={NavigationType.Frame}
                identifier={propertyOwner.identifier}
                showLabel
              />
            </Group>
          </Stack>
          </>
        )}
```

engbergandreas: messed up the indentation, but basically just wrap the thing in an empty fragment, and we can have both the divider and stack in one `isFocusable &&` check

engbergandreas (approved):

